PR #51, opened 2026-04-14
Replace bare assert with explicit RuntimeError

Jah-yee:
Good day,

## Summary
Replaces bare `assert self._client is not None` statements with explicit `RuntimeError` to provide better error handling.

## Problem
- bare `assert` is stripped by `python -O` — if the interpreter runs with optimizations, this guard silently disappears
- Unhelpful error message — when the assert fires, the traceback shows `AssertionError` with no context

## Fix
Replaced assert statements with explicit RuntimeError that:
- Cannot be stripped by python -O
- Provides clear error message: "httpx is required for LLM compression but is not installed. Install it with: pip install httpx"

## Changes
- `src/memtomem_stm/proxy/compression.py`: Replaced 4 assert statements in _call_api, _openai, _anthropic, _ollama methods

Thank you for your work on this project. I hope this small fix is helpful. Please let me know if there's anything to adjust.

Warmly,
RoomWithOutRoof


memtomem (commented):
Thanks for the contribution, @RoomWithOutRoof! The intent is sound — bare `assert` is indeed stripped by `python -O`, so replacing it with an explicit check is the right direction.

A few observations before merging:

### 1. `httpx` is a core dependency — the error message is misleading

`httpx` is listed in `pyproject.toml` `dependencies` (not optional), so the "httpx is not installed" scenario only applies to a broken installation. The more realistic path to `self._client is None` is **after `close()` is called** (line 1120 sets it to `None`). In that case, telling the user to `pip install httpx` is confusing.

A more accurate message would be something like:

```python
raise RuntimeError("LLM compressor client is not available (not initialized or already closed)")
```

### 2. The checks in provider methods are redundant

`_call_api` is the sole caller of `_openai` / `_anthropic` / `_ollama`. If `_call_api` already guards `self._client is None`, the three provider methods can never reach their own guard. Having the check in `_call_api` alone is sufficient.

If you prefer defense-in-depth for the provider methods too, a simple `assert` there is actually fine — those are internal invariants (the public path already raised), and `assert` is the idiomatic tool for "this should never happen" internal contracts.

### 3. Minor: 4× identical blocks

The same 4-line `if/raise` block is copy-pasted four times. If you keep checks in all four methods, consider extracting a small helper:

```python
def _ensure_client(self) -> httpx.AsyncClient:
    if self._client is None:
        raise RuntimeError(...)
    return self._client
```

Then each call site becomes `client = self._ensure_client()`.

---

**tl;dr** — The fix in `_call_api` is welcome. I'd suggest (a) fixing the error message and (b) dropping the redundant checks in the provider methods (or using a helper if you want to keep them).

memtomem (commented):
Thanks for this contribution, @Jah-yee! The motivation is spot-on — `assert` can be stripped by `python -O`, and replacing it with an explicit `RuntimeError` is the right pattern.

## Review notes

**Error message is accurate.** The only reachable path to `self._client is None` at `_call_api` is when `httpx` is not installed (`compression.py:1144`). The `_closed` guard at line 1185 already prevents post-`close()` calls from reaching `_call_api`, and the shutdown gate (`_idle.wait()`) ensures no in-flight call sees `_client = None` during teardown. So the "httpx is not installed" message is correct.

**Minor observation — the 3 provider-method checks are redundant but harmless.** `_call_api` already checks `self._client is None` before dispatching to `_openai`/`_anthropic`/`_ollama`. The duplicated checks in each provider method are purely defensive. Keeping them is fine (defense in depth), but if you'd like a smaller diff you could drop those 3 and keep only the one in `_call_api`. Either way works — your call.

## Before merge

The CLA bot needs your signature before we can merge. Please follow the instructions in the bot's comment above — it's a one-time "I agree" reply.

Once the CLA is signed, we'll get this merged. Thanks again for the clean, focused PR!

memtomem:
Hi @Jah-yee — friendly ping. Review's resolved on our side; the only thing still blocking merge is the CLA check.

The bot flagged a specific issue: the commit author email isn't currently linked to any GitHub account, so it can't verify the signer. Two ways to unblock:

1. **Link the commit email to your GitHub account** — *Settings → Emails → Add email address*, using the email in the PR's commit author line. Then comment `recheck` on this PR and sign with:
   > I have read the CLA Document and I hereby sign the CLA
2. **Or amend the commit to use a GitHub-linked email**:
   ```
   git config user.email <your-github-linked-email>
   git commit --amend --reset-author --no-edit
   git push -f
   ```

Either path works. Once the CLA check goes green we'll get this merged. Thanks for the patience!

Jah-yee:
I have read the CLA Document and I hereby sign the CLA

Issue #64, opened 2026-04-14
Guard PendingStore.get() against corrupted JSON in database


memtomem:
## Problem

`SQLitePendingStore.get()` calls `json.loads(row[0])` without error handling:

**Location:** `src/memtomem_stm/proxy/pending_store.py:135-136`

```python
return PendingSelection(
    chunks=json.loads(row[0]),  # JSONDecodeError if data is corrupted
    ...
)
```

If the SQLite database contains corrupted JSON (from partial writes, manual edits, or schema migration issues), this raises an unhandled `JSONDecodeError` that propagates to the caller.

## Suggested fix

Wrap in try/except, log the corruption, and return `None` (treat as cache miss):

```python
try:
    chunks = json.loads(row[0])
except json.JSONDecodeError:
    logger.warning("Corrupted JSON in pending_selections for key=%s", key)
    return None
```

memtomem:
Already fixed in 06ffd57 (squashed release). `SQLitePendingStore.get()` now wraps `json.loads()` in try/except and returns `None` on `JSONDecodeError` with a warning log. Tests: `tests/test_pending_store.py`. Closing.

PR #79, opened 2026-04-14
fix: defensive numeric parsing for untrusted external input (#66)

### LINKED ISSUE #66: Defensive numeric parsing: float("nan")/"inf" accepted and unguarded float() on external input
## Problem

Two places parse numeric values from external sources (LLM output, remote MCP core) without adequate defense:

### A1 — `src/memtomem_stm/proxy/extraction.py:52`
```python
confidence=float(item.get("confidence", 0.5)),
```
Python's `float()` accepts `"nan"`, `"inf"`, `"-inf"` as valid IEEE 754 values silently. The surrounding `try/except (json.JSONDecodeError, ValueError, TypeError)` does not catch these — NaN propagates downstream where confidence-based sort/filter becomes undefined.

**Trigger**: LLM returns `{"confidence": "nan"}` or `{"confidence": "inf"}`.

### A2 — `src/memtomem_stm/surfacing/mcp_client.py:137`
```python
score=float(item.get("score", 0.0)),
```
No `try/except` guard. A malformed score from the remote core parser (e.g. `"n/a"`, `"1.2.3"`) raises `ValueError` and breaks the entire response parse — not just one item.

**Trigger**: Remote MCP core returns a non-numeric score string.

## Proposed fix

Introduce a shared `_safe_float(value, default, *, reject_nonfinite=True)` helper that:
- Wraps `float()` in `try/except (TypeError, ValueError)`
- Rejects non-finite values via `math.isfinite()` when `reject_nonfinite=True`

Apply at both call sites. Add unit tests for the NaN/Inf/malformed-string paths.

memtomem:
## Summary

Closes #66.

- Introduces `memtomem_stm.utils.numeric.safe_float(value, default, *, reject_nonfinite=True)` — coerces to float, catches `TypeError`/`ValueError`, rejects `NaN`/`Inf` by default (which plain `float()` accepts silently).
- Applies it at the two call sites identified in the issue:
  - `proxy/extraction.py:52` — confidence from LLM output (was: `float("nan")` slipped through as NaN and broke downstream sort/filter).
  - `surfacing/mcp_client.py:137` — score from remote core (was: malformed numeric string raised `ValueError` and broke the entire response parse).
- Unit tests in `tests/test_safe_float.py` cover finite passthrough, malformed-string fallback, explicit `NaN`/`Inf` rejection, and the opt-in `reject_nonfinite=False` escape hatch.

## Commits

1. `fix:` — the defensive parsing change (the actual #66 fix).
2. `chore:` — a one-line pre-existing `ruff format` drift in `progressive.py` left over from PR #64/#65. Unrelated to this issue but was blocking CI on all branches, so it is fixed here to unblock the PR. Reviewers can skip this commit.

## Test plan

- [x] `uv run pytest -m "not ollama"` — 1081 passed
- [x] `uv run mypy src/memtomem_stm/utils/numeric.py src/memtomem_stm/proxy/extraction.py src/memtomem_stm/surfacing/mcp_client.py` — clean
- [ ] CI green on PR

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -50,6 +50,8 @@ export MEMTOMEM_STM_PROXY__RELEVANCE_SCORER__SCORER=bm25           # "bm25" or "
 export MEMTOMEM_STM_PROXY__RELEVANCE_SCORER__EMBEDDING_PROVIDER=ollama
 export MEMTOMEM_STM_PROXY__RELEVANCE_SCORER__EMBEDDING_MODEL=nomic-embed-text
 export MEMTOMEM_STM_PROXY__RELEVANCE_SCORER__EMBEDDING_BASE_URL=http://localhost:11434
+# When embedding_provider is "openai", set OPENAI_API_KEY:
+# export OPENAI_API_KEY=sk-...
 
 # Extraction (Stage 4b — auto fact extraction)
 export MEMTOMEM_STM_PROXY__EXTRACTION__ENABLED=false
@@ -194,7 +196,6 @@ Full example with all options:
     "scorer": "bm25",
     "embedding_provider": "ollama",
     "embedding_model": "nomic-embed-text",
-    "embedding_base_url": "http://localhost:11434",
     "embedding_timeout": 10.0
   },
   "extraction": {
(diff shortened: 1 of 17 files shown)


tsdata (approved):
LGTM — defensive float parsing looks clean. Helper placement in utils/numeric.py matches the existing utils/ layout. Unit tests cover the NaN/Inf rejection and the opt-in escape hatch. chore commit unblocks repo-wide ruff format.

PR #80, opened 2026-04-14
fix: guard LLM provider responses against empty/malformed payloads (#67)

### LINKED ISSUE #67: LLM response parsing: unchecked choices[0]/content[0] access crashes on empty responses
## Problem

LLM response parsing assumes non-empty response arrays at multiple call sites.

### A3 — `src/memtomem_stm/proxy/compression.py:1150` (OpenAI)
```python
return resp.json()["choices"][0]["message"]["content"]
```
OpenAI can return `{"choices": []}` under quota exhaustion, content filtering, or certain error recovery paths. This raises `IndexError`, which the caller cannot distinguish from a parse error.

### B3 — `src/memtomem_stm/proxy/compression.py` Anthropic response path
The Anthropic handler in the same file follows the same pattern for `content[0]["text"]` and is vulnerable to empty `content` arrays on truncated/empty completions.

## Impact

- `IndexError` with no context raised deep in the compression stack
- Caller's fallback chain cannot react correctly — the whole compression attempt fails instead of gracefully falling back
- Operators see opaque tracebacks in logs

## Proposed fix

1. Check array length before indexing in both OpenAI and Anthropic paths
2. Raise a specific `LLMCompressionError("empty response")` (or return empty string to trigger the existing fallback path — whichever matches the intended semantics)
3. Add tests with mocked HTTP responses for:
   - Empty `choices`/`content` arrays
   - Missing `message`/`text` field within the first element

## Related

Same defensive-parse-pattern family as #66 (numeric parsing of external input).

memtomem:
## Summary

Closes #67.

The three `LLMCompressor` provider methods (`_openai`, `_anthropic`, `_ollama`) indexed into response arrays and nested fields without validation. OpenAI returns `{"choices": []}` on quota / content filtering; Anthropic can return empty `content` on truncated completions; Ollama can return a bare error envelope. All three cases raised opaque `IndexError`/`KeyError` rather than a descriptive error.

The outer `compress()` already catches `Exception` and falls back to truncate, so the **fallback behavior is unchanged**. This PR only improves the log message so operators can distinguish "empty provider response" from "network/parse error" in incident triage.

### Scope note

Issue #67 called out OpenAI (A3) and Anthropic (B3). Ollama uses the same unguarded-access pattern in the same file and is fixed here as compounding scope — same PR, same test file, same review.

## Test plan

- [x] `uv run pytest -m "not ollama"` — 1091 passed
- [x] `uv run ruff check src tests/test_compression.py` — clean
- [x] `uv run ruff format --check src` — clean
- [x] `uv run mypy src/memtomem_stm/proxy/compression.py` — clean
- New tests in `tests/test_compression.py::TestLLMCompressorEmptyResponseGuard`:
  - Each provider: empty-array path, missing-content path, happy path
  - End-to-end: empty OpenAI `choices` triggers `llm_error` fallback without unhandled exception
- [ ] CI green on PR

### DIFF
--- a/src/memtomem_stm/proxy/compression.py
+++ b/src/memtomem_stm/proxy/compression.py
@@ -1147,7 +1147,15 @@ class LLMCompressor:
             },
         )
         resp.raise_for_status()
-        return resp.json()["choices"][0]["message"]["content"]
+        data = resp.json()
+        choices = data.get("choices") or []
+        if not choices:
+            raise ValueError("OpenAI response has empty 'choices' (likely quota or content filter)")
+        message = choices[0].get("message") or {}
+        content = message.get("content")
+        if not isinstance(content, str):
+            raise ValueError("OpenAI response missing 'choices[0].message.content'")
+        return content
 
     async def _anthropic(self, text: str, system_prompt: str) -> str:
         if self._client is None:
@@ -1171,7 +1179,16 @@ class LLMCompressor:
             },
         )
         resp.raise_for_status()
-        return resp.json()["content"][0]["text"]
+        data = resp.json()
+        content = data.get("content") or []
+        if not content:
+            raise ValueError(
+                "Anthropic response has empty 'content' (likely empty completion or filter)"
+            )
+        text_block = content[0].get("text")
+        if not isinstance(text_block, str):
+            raise ValueError("Anthropic response missing 'content[0].text'")
+        return text_block
 
     async def _ollama(self, text: str, system_prompt: str) -> str:
         if self._client is None:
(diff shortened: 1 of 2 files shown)


tsdata (approved):
LGTM — defensive guards on all three provider paths (OpenAI / Anthropic / Ollama) with clear ValueError messages. Fallback behavior via the outer compress() exception handler is preserved, so this is a pure log-quality improvement. Tests cover each provider's empty-array path, missing-content path, happy path, and an end-to-end fallback assertion.

PR #81, opened 2026-04-14
fix: preserve input order when embedding provider omits "index" field (#68)

### LINKED ISSUE #68: Embedding response sort fails with KeyError when provider omits "index" field
## Problem

`src/memtomem_stm/proxy/relevance.py:189`:
```python
data.sort(key=lambda x: x["index"])
```

This assumes every embedding object in the provider response has an `"index"` field. OpenAI's official API does include it, but several common OpenAI-compatible endpoints **do not**:

- Ollama (when exposed via OpenAI-compatibility layer)
- LiteLLM proxy (varies by underlying provider)
- LM Studio
- Other self-hosted OpenAI-compatible servers

When the field is missing, `KeyError: "index"` is raised and **the entire embedding request fails**, not just one item.

## Severity

Elevated from original triage — Ollama-compat providers are a primary supported config, and this is a single-point-of-failure for any embedding request in such environments. Fix is a two-line change.

## Proposed fix

Use `enumerate()`-based fallback index when `"index"` is missing. Preserve input order in that case:

```python
# If any item lacks "index", trust input order instead of sorting.
if all("index" in x for x in data):
    data.sort(key=lambda x: x["index"])
return [d["embedding"] for d in data]
```

Add test coverage with:
- Standard OpenAI response (with `index`)
- Ollama-style response (no `index`)
- Mixed response (partial `index` — should fall back to input order)

memtomem:
## Summary

Closes #68.

`EmbeddingScorer._embed_openai` sorted the response by the `"index"` field on every embedding object. Official OpenAI populates this field, but many OpenAI-compatible servers (Ollama compat layer, LiteLLM, LM Studio, self-hosted) omit it — sorting raised `KeyError` and broke the entire embedding request.

**Fix** (2 lines): sort by `"index"` only when every item populates it; otherwise trust input order.

```python
if data and all("index" in d for d in data):
    data.sort(key=lambda x: x["index"])
return [d["embedding"] for d in data]
```

This covers:
- Official OpenAI: all items have `index` → sort happens as before
- Ollama-compat / LiteLLM / LM Studio: no items have `index` → trust input order
- Mixed responses (defensive): fall back to input order rather than `KeyError`

## Test plan

- [x] `uv run pytest -m "not ollama"` — 1095 passed (4 new in `TestEmbeddingOpenAIResponseParsing`)
- [x] `uv run ruff check src tests/test_query_aware.py` — clean
- [x] `uv run ruff format --check src` — clean
- [x] `uv run mypy src/memtomem_stm/proxy/relevance.py` — clean
- New tests cover: sorted-by-index, no-index (compat), partial-index (mixed), empty-data
- [ ] CI green on PR

### DIFF
--- a/src/memtomem_stm/proxy/relevance.py
+++ b/src/memtomem_stm/proxy/relevance.py
@@ -185,8 +185,11 @@ class EmbeddingScorer:
         )
         resp.raise_for_status()
         data = resp.json()["data"]
-        # Sort by index to ensure order matches input
-        data.sort(key=lambda x: x["index"])
+        # Sort by "index" when the provider populates it (official OpenAI).
+        # OpenAI-compatible servers — Ollama's compat layer, LiteLLM, LM Studio —
+        # often omit the field, in which case we trust the input order.
+        if data and all("index" in d for d in data):
+            data.sort(key=lambda x: x["index"])
         return [d["embedding"] for d in data]
 
 
--- a/tests/test_query_aware.py
+++ b/tests/test_query_aware.py
@@ -164,6 +164,85 @@ class TestRelevanceScorerProtocol:
         assert third_len >= first_len
 
 
+# ── EmbeddingScorer OpenAI response parsing (#68) ──────────────────────
+
+
+class TestEmbeddingOpenAIResponseParsing:
+    """Provider compatibility for ``EmbeddingScorer._embed_openai``.
+
+    Official OpenAI populates ``index`` on every embedding object. Many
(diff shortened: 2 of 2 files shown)


tsdata (approved):
LGTM — the `all("index" in d for d in data)` guard cleanly covers official OpenAI, compat providers (Ollama/LiteLLM/LM Studio), and mixed responses without adding provider-specific branches. Four tests lock in each shape. Simple and correct.

PR #92, opened 2026-04-14
test: CLA e2e smoke (pandas-studio) — DO NOT MERGE

pandas-studio:
E2E verification for #89 / PR #90.

This PR exists to confirm that `contributor-assistant/github-action`, now authenticated with `PERSONAL_ACCESS_TOKEN`, can commit a new signature to `main` automatically instead of failing with branch-protection errors.

Plan:
1. Expect the bot to comment "please sign the CLA" on this PR.
2. I'll post the sign comment next.
3. Watch `signatures/v1/cla.json` for an automatic commit adding `pandas-studio`.
4. Close this PR and delete the branch/marker file once verified — do not merge.

Issue #91, opened 2026-04-14
CI: migrate CLA bot auth from admin PAT to GitHub App bypass actor


memtomem:
Follow-up to #89. Once #89's admin-PAT approach is stable, migrate the CLAAssistant workflow off the PAT and onto a **GitHub App registered as a ruleset bypass actor** for `main`.

## Why migrate

- **No token rotation.** PATs expire; App installations don't. Removes an ongoing maintenance item and an outage mode (silent PAT expiry → CLA check silently broken again).
- **Finer-grained.** Bypass is scoped to a specific ruleset, not "anything this PAT can do anywhere."
- **GitHub-recommended** for "bot needs to push to a protected branch" patterns.

## Dependencies

Blocked on #89 completing the admin-PAT path. Don't migrate until:

- PAT-based setup has passed the e2e test on at least one real external-contributor PR.
- The workflow's behavior under `PERSONAL_ACCESS_TOKEN` is well understood (so the App swap is a pure auth change, not a debugging session combined with a migration).

## Rough plan

1. Create a GitHub App owned by `memtomem` with permissions: `Contents: write`, `Pull requests: write`, `Metadata: read`. Install it on `memtomem/memtomem-stm` only.
2. Add the App to the `main` branch ruleset as a bypass actor.
3. In CI, mint an App installation token per run and pass it as `PERSONAL_ACCESS_TOKEN` to `contributor-assistant/github-action` (e.g. via `actions/create-github-app-token` or equivalent).
4. Remove the admin PAT secret once the App path is verified green on a real signing flow.
5. Update #89's resolution notes and the `project_cla_bot_bypass_gap.md` project memory with the final architecture.

## Open questions

- Does the App need to be public, or does a private App installed only on this repo suffice? (Private is preferred.)
- Are there cases where a ruleset bypass actor is subtly different from the `enforce_admins: false` admin bypass we're using now? Worth verifying with a dry-run push from the App before removing the PAT.

pandas-studio:
Adding a data point from #89's e2e test that strengthens the case for migrating here:

A fine-grained PAT scoped to this repo with `Contents: write` + `Pull requests: write` **cannot bypass branch protection** — only classic PATs (which inherit the user's admin role) can. So the interim fix in #89 ends up using a `repo`-scope classic PAT, which is broader than ideal.

[...]

PR #90, opened 2026-04-14
ci(cla): wire PERSONAL_ACCESS_TOKEN so CLAAssistant can commit signatures

memtomem:
Closes #89 once the secret is in place and the e2e test passes.

## Summary
- Add `PERSONAL_ACCESS_TOKEN` env var to the CLAAssistant step.
- Inline comment documents why the built-in `GITHUB_TOKEN` is not enough.

## Why
`contributor-assistant/github-action` commits the updated `signatures/v1/cla.json` to `main` directly. The built-in `GITHUB_TOKEN` is not an admin actor, so branch protection on `main` (`required_status_checks: [lint, test, notebooks]` + 1 approving review) rejects the push. `enforce_admins` is `false`, so an admin PAT can bypass protection and let the commit through. See #89 for full context including the failing run from PR #87.

## Prerequisites before merging this PR

This PR only references `secrets.PERSONAL_ACCESS_TOKEN` — the secret itself still needs to be created. **Do not merge until all three are done:**

1. Generate a fine-grained PAT from the `memtomem` admin account:
   - Resource: `memtomem/memtomem-stm` only
   - Repository permissions: `Contents: write`, `Pull requests: write`
   - Expiry: longest allowed; note expiry on a calendar
2. Add the secret:
   ```bash
   gh secret set PERSONAL_ACCESS_TOKEN --repo memtomem/memtomem-stm
   ```
3. Verify: `gh secret list --repo memtomem/memtomem-stm` shows `PERSONAL_ACCESS_TOKEN`.

If the secret is absent when the bot runs, `${{ secrets.PERSONAL_ACCESS_TOKEN }}` expands to an empty string and the action falls back to `GITHUB_TOKEN` — i.e. the same broken behavior as today. Merging early is not unsafe, just ineffective.

## Test plan

- [ ] `PERSONAL_ACCESS_TOKEN` secret created and visible in `gh secret list`.
- [ ] Open a throwaway PR from a second account (e.g. `pandas-studio`) that has not yet signed the CLA.
- [ ] Post the sign comment on that PR.
- [ ] Confirm `signatures/v1/cla.json` gets a new entry committed to `main` **automatically** (no manual PR needed) and the `CLAAssistant` check turns green.
- [ ] Close the throwaway PR without merging; leave the signature entry in place.

## Not included here

- Scheduled workflow to alert on PAT expiry — can be added as a follow-up if a calendar reminder feels fragile.
- GitHub App migration — tracked as a separate follow-up issue referenced from #89.

Issue #89, opened 2026-04-14
CI: CLA bot cannot auto-commit signatures — branch protection blocks direct push


memtomem:
## Context

`contributor-assistant/github-action` (CLAAssistant workflow) is supposed to append a new entry to `signatures/v1/cla.json` when an external contributor signs via comment. After a signature it should auto-commit to `main` and re-run, turning the `CLAAssistant` check green with no human intervention.

Since at least 2026-04-14 this has stopped working for external contributors. The bot still edits its PR comment to "All contributors have signed ✅" (false positive), but the workflow exits 1, the signature file is never updated, and the PR stays blocked.

### Evidence (PR #87 / ayushi-wq)

Workflow run [24412017536](https://github.com/memtomem/memtomem-stm/actions/runs/24412017536) failed with:

```
##[error]Could not update the JSON file: Could not update file:
Changes must be made through a pull request.
3 of 3 required status checks are expected.
```

Current `main` protection (`gh api repos/memtomem/memtomem-stm/branches/main/protection`):

- `required_status_checks.contexts`: `["lint", "test", "notebooks"]` (`strict: true`)
- `required_approving_review_count`: 1
- `enforce_admins`: **false** (important — admins can bypass)

The bot uses the built-in `GITHUB_TOKEN`, which is **not** an admin actor, so it gets rejected by branch protection.

### Why this surfaced recently

- ✅ 2026-04-11 — tsdata, PR #19: auto-signed, committed to main.
- ✅ 2026-04-14 06:28Z — nithishkumar-dev-10, PR #39: auto-signed, committed to main.
- ❌ 2026-04-14 16:57Z — ayushi-wq, PR #87: rejected with the error above.

Protection was tightened somewhere in that 10-hour window (most likely a required-status-checks addition). **Action item:** verify in `Settings → Security log` (Org audit log isn't available — `memtomem` is a User account, so `gh api` cannot surface this; the user security log is UI-only). Confirming it was an intentional tightening matters, because the fix must bypass it by design, not undo it.

### Interim workaround applied for #87

Manual PR #88 (commit `0fbe23c`) added ayushi-wq's signature, merged by admin, then `recheck` on #87 re-ran CLAAssistant → pass. This works but requires admin action per external contributor, which is exactly the toil we want to remove.

## Proposed fix (a): admin PAT as `PERSONAL_ACCESS_TOKEN` secret

[...]

PR #93, opened 2026-04-15
ci(cla): pin remote-* inputs so PERSONAL_ACCESS_TOKEN actually gets used

memtomem:
Follow-on to #90. Discovered during the e2e test in #89: PR #90's `PERSONAL_ACCESS_TOKEN` wiring was correct but the action was ignoring it.

## Root cause

`contributor-assistant/github-action` only uses `PERSONAL_ACCESS_TOKEN` when `remote-organization-name` or `remote-repository-name` is set. Source ([persistence.ts](https://github.com/contributor-assistant/github-action/blob/v2.6.1/src/persistence/persistence.ts)):

\`\`\`ts
const octokitInstance =
  isRemoteRepoOrOrgConfigured() ? getPATOctokit() : getDefaultOctokitClient()

function isRemoteRepoOrOrgConfigured(): boolean {
  return !!(input.getRemoteRepoName() || input.getRemoteOrgName())
}
\`\`\`

With both inputs empty (the default for "store signatures in this same repo"), the action falls back to `GITHUB_TOKEN` regardless of whether `PERSONAL_ACCESS_TOKEN` is set. Branch protection then rejects the push, exactly as before #90.

## Fix

Set `remote-organization-name` and `remote-repository-name` to this repo's own coordinates. The action treats it as "remote", flips to PAT octokit, but ends up calling the same GitHub API endpoint — just authenticated as the PAT owner instead of `github-actions[bot]`. PAT bypasses branch protection because the PAT owner has admin role.

## Test plan

PR #92 stays open as the e2e harness:

- [ ] Merge this PR.
- [ ] Comment `recheck` on PR #92.
- [ ] Verify the bot commits a `pandas-studio` entry to `signatures/v1/cla.json` on `main` automatically (no manual PR).
- [ ] Verify CLAAssistant check on PR #92 turns green.
- [ ] Close PR #92, delete its branch and the `.cla-e2e-marker.md` file.

If this works, #89 is resolved. If it still fails, escalate to GitHub App migration (#91).

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -30,6 +30,13 @@ jobs:
           PERSONAL_ACCESS_TOKEN: ${{ secrets.PERSONAL_ACCESS_TOKEN }}
         with:
           path-to-signatures: 'signatures/v1/cla.json'
+          # Pin the action's `isRemoteRepoOrOrgConfigured` branch to true even
+          # though signatures live in this same repo. Without these inputs the
+          # action ignores PERSONAL_ACCESS_TOKEN and uses GITHUB_TOKEN, which
+          # cannot bypass branch protection on `main`. See #89 for the source-
+          # code analysis (src/persistence/persistence.ts).
+          remote-organization-name: memtomem
+          remote-repository-name: memtomem-stm
           path-to-document: 'https://github.com/memtomem/memtomem-stm/blob/main/CLA.md'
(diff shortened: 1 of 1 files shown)


tsdata (approved):
LGTM — source-code analysis is convincing, and minimal-surface-area fix is the right call before escalating to the App route.

PR #132, opened 2026-04-15
fix(proxy): validate reconnect delay ordering in UpstreamServerConfig

ayushi-wq:
Fix:
- Added @model_validator(mode="after") check
- Prevents invalid configuration at startup

This ensures correct retry backoff behavior and avoids silent misconfiguration.


memtomem (commented):
## Overview

Nice catch on the silent misconfiguration — the validator logic itself is correct, and `manager.py:887`'s `min(delay * 2, max_reconnect_delay_seconds)` backoff would otherwise just truncate a reversed config without complaining. There's one import issue blocking CI though:

## Blockers

- **Broken imports — CI fails lint + test.** The diff drops the entire `from typing import Any, Literal` line, but both names are still used in this file:
  - `Any` — 7 call sites (`collect_proxy_env_overrides` return, `_deep_merge`, `ProxyConfig.load_from_file`, etc.)
  - `Literal` — `pending_store: Literal[\"memory\", \"sqlite\"]` at line 140
  - `Self` — the new validator's return annotation is `-> Self:`, but `Self` was never added to imports

  Runtime impact is not just lint — pydantic raises `PydanticUserError: UpstreamServerConfig is not fully defined; you should define 'Literal'` on first model construction, which takes out the whole proxy. CI confirms: \`lint fail\`, \`test fail\`.

  **Fix:** restore the import and add \`Self\`:
  \`\`\`python
  from typing import Any, Literal, Self
  \`\`\`

- **No regression test for the validator.** Please add a case to \`tests/test_proxy_config*.py\` asserting that \`UpstreamServerConfig(prefix=\"x\", reconnect_delay_seconds=10, max_reconnect_delay_seconds=5)\` raises \`ValidationError\`, and a positive case where equal values are accepted.

## Nits

- Error message is clear; the \`f\"... ({value}) must be <= ...\"\` form is consistent with pydantic's own style.
- \`>\` (not \`>=\`) correctly permits \`reconnect == max_reconnect\`, which is a degenerate-but-valid \"no exponential growth\" config. Good call.
- Same kind of ordering check could apply to \`connect_timeout_seconds\` vs. the delays, but that feels out of scope for this PR.

## Verdict

Logic is right, but the import deletion makes the PR currently unmergeable. Once \`Any, Literal, Self\` are re-imported and a test is added, this is a small, worthwhile fix.

memtomem (changes requested):
## Overview

Nice catch on the silent misconfiguration — the validator logic itself is correct (`>` rather than `>=` lets `reconnect == max_reconnect` through, which is a valid "no exponential growth" config). There are a couple of blockers before this can merge though:

## Blockers

- **Broken imports — CI fails lint + test.** The diff drops the entire `from typing import Any, Literal` line, but both names are still used in this file:
  - `Any` — 7 call sites (`collect_proxy_env_overrides` return, `_deep_merge`, `ProxyConfig.load_from_file`, etc.)
  - `Literal` — `pending_store: Literal["memory", "sqlite"]` at line 140
  - `Self` — the new validator is annotated `-> Self:`, but `Self` was never added to imports

  Runtime impact isn't just lint — pydantic raises `PydanticUserError: UpstreamServerConfig is not fully defined; you should define 'Literal'` on first model construction, which takes out the whole proxy. CI confirms: `lint` fail, `test` fail.

  **Fix:** restore the import and add `Self`:

  ```python
  from typing import Any, Literal, Self
  ```

- **No regression test for the validator.** Please add a case to `tests/test_proxy_config*.py` asserting that `UpstreamServerConfig(prefix="x", reconnect_delay_seconds=10, max_reconnect_delay_seconds=5)` raises `ValidationError`, plus a positive case where equal values are accepted.

## Nits

- Error message is clear; the `f"... ({value}) must be <= ..."` form is consistent with pydantic's own style.
- `>` (not `>=`) correctly permits `reconnect == max_reconnect`, which is a degenerate-but-valid "no exponential growth" config. Good call.
- Whether the same kind of check belongs on `connect_timeout_seconds` vs. the delays is probably out of scope for this PR — separate concern.

## Verdict

Logic is right, but the import deletion makes the PR currently unmergeable. Once `Any, Literal, Self` are re-imported and a test is added, this is a small, worthwhile fix.

ayushi-wq:
Hi! Thanks for the review.

You mentioned adding tests in `tests/test_proxy_config*.py`, but I'm not able to find a file matching that pattern in the repository.

Could you please clarify which test file would be the best place to add these validation tests?

memtomem:
Good question — my review used a glob that doesn't match any existing file, sorry about the confusion.

The right place is **`tests/test_config_constraints.py`** — it already covers pydantic `ValidationError` assertions for proxy config validators (`LLMCompressorConfig`, `SelectiveConfig`, etc.). Add `UpstreamServerConfig` to the imports and your test cases would fit naturally under the existing `TestProxyNumericConstraints` class:

```python
def test_reconnect_delay_must_not_exceed_max(self) -> None:
    with pytest.raises(ValidationError):
        UpstreamServerConfig(prefix="x", reconnect_delay_seconds=10, max_reconnect_delay_seconds=5)

def test_reconnect_delay_equal_to_max_is_valid(self) -> None:
    cfg = UpstreamServerConfig(prefix="x", reconnect_delay_seconds=5, max_reconnect_delay_seconds=5)
    assert cfg.reconnect_delay_seconds == 5
```

Also a reminder on the import fix from the earlier review — `from typing import Any, Literal, Self` needs to be restored in `config.py` before CI will pass.

memtomem:
Thanks for the quick turnaround. Imports are restored (`Any, Literal, Self`), the validator is in place, and the two new tests cover both the rejection and the equal-to-max positive case cleanly. CI is fully green — merging.

Small non-blocking note for future PRs: running `uv run ruff format tests` locally would normalize the slight indent/trailing-whitespace on the new test block. Our CI currently only format-checks `src/`, so these slip through, but `format` locally over `tests/` catches them.

PR #190, opened 2026-04-19
test(contract): cover empty-structured JSON from parent PR #231

memtomem:
## Summary

- Pin regression coverage for parent memtomem PR #231 (commit `7d184f1`, 2026-04-18), which changed `mem_search(output_format="structured")` to return `{"results": []}` JSON (optionally with a `"hints"` array) on empty results instead of the plain-text `"No results found."`
- `StructuredResultParser` already handles both shapes correctly — this commit just locks that in with two tests so a future drift is caught fast.

## Why the pin is loose

Parent is classified `Development Status :: 3 - Alpha` (`packages/memtomem/pyproject.toml:12`). We pin only `results == []`; the `hints` field is **opportunistic** — present or absent both parse as empty, and a future rename or removal upstream must not block STM.

## What's NOT in this PR

- Fake-server (`_fake_memtomem_server.py`) extension to emit empty-structured JSON end-to-end. That integration-path coverage lands in a later bucket alongside the hints-observability follow-up.
- No production code change. `StructuredResultParser.parse()` already works for both shapes.

## Test plan

- [x] `uv run pytest tests/test_core_format_contract.py -v` → 27 passed (25 existing + 2 new)
- [x] `uv run ruff check src tests` → clean

### DIFF
--- a/tests/test_core_format_contract.py
+++ b/tests/test_core_format_contract.py
@@ -309,6 +309,39 @@ class TestStructuredFormatSnapshots:
         assert results[1].chunk.id == "def456"
 
 
+# Parent PR #231 (7d184f1, 2026-04-18): mem_search(output_format="structured")
+# now returns JSON on empty results instead of the plain "No results found."
+# text. We pin only ``results == []`` — parent is alpha (Development Status
+# :: 3 - Alpha) and the ``hints`` field is opportunistic, so a future rename
+# or removal upstream must not block STM.
+
+STRUCTURED_NO_RESULTS_PLAIN = '{"results": []}'
+
+STRUCTURED_NO_RESULTS_WITH_HINTS = (
+    '{"results": [], "hints": ['
+    '"No results match your filters (3 results found before filtering). '
+    'Try broader filters or remove source_filter/tag_filter."'
+    ']}'
+)
+
+
+class TestStructuredEmptyResults:
+    """StructuredResultParser tolerates parent PR #231's empty-result JSON
+    in both the bare and hints-augmented shapes."""
+
+    def test_structured_parser_returns_empty_without_hints(self):
+        from memtomem_stm.surfacing.mcp_client import StructuredResultParser
+
+        parser = StructuredResultParser()
+        assert parser.parse(STRUCTURED_NO_RESULTS_PLAIN) == []
+
+    def test_structured_parser_returns_empty_with_hints(self):
+        from memtomem_stm.surfacing.mcp_client import StructuredResultParser
+
+        parser = StructuredResultParser()
+        assert parser.parse(STRUCTURED_NO_RESULTS_WITH_HINTS) == []
+
+
 class TestOutputFormatForwarding:
     """Verify output_format is sent to MCP when using structured parser."""
 


memtomem (commented):
LGTM — plan-consistent and ready to merge.

**Working well**

- Loose pin is precise: both tests assert only `parse(...) == []`, with the `hints` value left unchecked. Alpha-tolerance rule implemented exactly as agreed — parent can rename or drop `hints` without blocking STM.
- Provenance in the module comment (PR #231 / commit `7d184f1` / `pyproject.toml:12` alpha marker) means anyone digging into this test two years from now can reconstruct the "why" from the test file alone.
- "What's NOT in this PR" callout scopes this correctly and points at where the fake-server integration path coverage will land.
- Two separate tests instead of `@pytest.mark.parametrize` — one line longer but gives a named traceback if exactly one shape regresses. Good trade.

**Minor (non-blocking)**

- Import style: `StructuredResultParser` is imported inside each test method; the rest of the file does a top-level import (`McpClientSearchAdapter` at line 15). If other `TestStructured*` classes use inline imports, keep as-is for local consistency; otherwise move to module top.
- Module-level comment (lines 312-317) and the class docstring overlap slightly. Not worth churning.

**One question (does not block merge)**

Where did the B0 precheck items end up recorded? Specifically:

1. `core`-mode routing for `mem_do(action="increment_access"|"scratch_get")` in parent
2. STM's declared minimum supported parent version
3. Downstream surface `@mcp.tool()` that B3 will forward `hints` through
4. Env var naming verified against parent / STM `Settings` (`MEMTOMEM_INDEXING__AUTO_DISCOVER`, `MEMTOMEM_STM_SURFACING__LTM_MCP_COMMAND`, etc.)

None needed for this PR — but B2 smoke depends on all four. If they're already in a memo / issue / scratch note, ignore. Otherwise, worth capturing before B2 kicks off so the isolation envs don't silently no-op.

memtomem:
Good catch — the B0 notes are in a local plan file, so re-surfacing them here so they're visible before B2 kicks off.

### B0 resolution (2026-04-19)

**(1) `mem_do(action=...)` routing in parent `core` mode — OK, no flag needed.**
`mem_do` is in `_CORE_TOOLS` (`packages/memtomem/src/memtomem/server/__init__.py:122`). The action registry imports in the same file pull in `mem_scratch_get` (`tools/scratch.py:55`) and `mem_increment_access` (`tools/search.py:256`); both use `@register` and are name-stripped via `tool_registry.py:79` → actions `"scratch_get"` and `"increment_access"` become dispatchable through `mem_do` even in `core` mode. STM's existing `ltm_mcp_command` + empty `ltm_mcp_args` default works as-is. (Tool mode env is `MEMTOMEM_TOOL_MODE`, unprefixed — only relevant if we ever want to opt into `standard`/`full`.)

**(2) STM's minimum supported parent version — intentionally unpinned.**
No `memtomem` dep in STM `pyproject.toml`; only `CHANGELOG.md:65` historical reference to `v0.1.7`. Given parent's alpha status we're keeping it unpinned at the dep level. A separate docs PR can add a "tested against 0.1.12" note to the README if we want to set expectations without shipping a hard constraint. **B2 version matrix: skipped** (conditional on a concrete minimum being declared, which we're not doing).

**(3) Downstream surface — text prepend, not a structured `@mcp.tool()` response.**
`SurfacingEngine.surface()` (`src/memtomem_stm/surfacing/engine.py:116`) is invoked from `src/memtomem_stm/proxy/manager.py:599,653` and the result is prepended to the proxied tool's response text when `injection_mode == "prepend"` (line 643). STM does not emit a structured JSON body that could carry a `hints` field through to downstream. **Implication for B3:** "forward" means (a) SURFACE INFO log + (b) exposing hints via an operator-visible path (e.g. a counter/last-hints snapshot in `stm_proxy_stats`). Changing the prepend body to inject "no memories surfaced (reason: ...)" when LTM returned zero hits is a policy change and is **deliberately out of B3 scope** — if we want it, it should be a separate proposal.

**(4) Env var naming for B2 isolation — validated against source.**
- Parent `Mem2MemConfig` (`packages/memtomem/src/memtomem/config.py:434-436`): `env_prefix="MEMTOMEM_"`, `env_nested_delimiter="__"`.
- STM `STMConfig` (`src/memtomem_stm/config.py:37-39`): `env_prefix="MEMTOMEM_STM_"`, `env_nested_delimiter="__"`.
- → The four env vars in the B2 plan (`MEMTOMEM_INDEXING__AUTO_DISCOVER`, `MEMTOMEM_STORAGE__SQLITE_PATH`, `MEMTOMEM_INDEXING__MEMORY_DIRS`, `MEMTOMEM_STM_SURFACING__LTM_MCP_COMMAND`) are correct as written.
- `memory_dirs: list[str]` accepts `"[]"` (JSON empty-list decode is pydantic-settings' default for list fields).
- Will add a positive assertion in B2 step 4 (parent startup log / `mem_stats` response check) to confirm `auto_discover=False` actually took effect — otherwise a silent typo fallback would pass B2 green while isolation was inert.

B2 will proceed after this PR merges. B3 design (Option A `tuple[list, list[str]]` return as soft default vs. B stateful stash) finalizes post-caller-audit.

memtomem:
### B2 smoke — real parent v0.1.12 (green)

Ran against the real parent binary as planned. Follow-up to B0/B1 in this PR.

**Environment**
- Date: 2026-04-19
- STM HEAD: `1ac40ac` (this PR's squash merge)
- Parent HEAD: `8b6d343` (tag `test-v0.1.12a1`, in `~/work/agent-harness/memtomem`)
- Parent PR #231 (`7d184f1`) confirmed as ancestor of `8b6d343` — empty-structured JSON path is live.
- Mode: **core** (parent default, `--mode` unset, STM `ltm_mcp_args=[]`).
- Isolation env (all four vars set, per B0 (4)):
  - `HOME=$(mktemp -d)/stm-smoke-home`
  - `MEMTOMEM_INDEXING__AUTO_DISCOVER=0`
  - `MEMTOMEM_INDEXING__MEMORY_DIRS=[]`
  - `MEMTOMEM_STORAGE__SQLITE_PATH=$HOME/stm-smoke.db`
  - `MEMTOMEM_STM_SURFACING__LTM_MCP_COMMAND=<abs parent binary>`

**Positive assertion — isolation took effect (not a silent fallback)**
- After the run, `$SMOKE_HOME/stm-smoke.db` exists (204 800 B, fresh). `~/.memtomem/memtomem.db` timestamp unchanged from before the smoke.
- `mem_stats` on the isolated parent reports `Total chunks: 0, Total sources: 0`. With `AUTO_DISCOVER=1` (default) on a dev machine this would be non-zero (dogfood dirs), so zero is evidence both `AUTO_DISCOVER=0` and `MEMORY_DIRS=[]` took effect — typo would have silently rolled back to the default.
- `~/.memtomem/stm_feedback.db` also untouched (STM's own state never touched dogfood path because `HOME` was overridden for the whole run).

**Empty-search contract — real parent matches B1 fixture**
Called `mem_search(query="...", limit=3, output_format="structured")` on the empty isolated DB. Response:

```json
{"results": [], "hints": ["semantic search unavailable: no such table: chunks_vec"]}
```

- `results == []` ✅ — matches the B1 loose-pin in `tests/test_core_format_contract.py::TestStructuredEmptyResults`.
- `hints` present and carrying a real parent-side note ✅ — `STRUCTURED_NO_RESULTS_WITH_HINTS` path is exercised.
- No `WARNING: invalid JSON` emitted by parent.
- For reference, default (text) format still returns `"No results found. (Note: semantic search unavailable: no such table: chunks_vec)"` — `StructuredResultParser` is only invoked when STM opts into `output_format="structured"`, matching Phase 2.

**mem_do dispatch in core mode — routing confirmed (B0 (1))**
- `mem_do(action="scratch_get", params={})` → `"Working memory is empty."`, `isError=False`. Action is reachable through `mem_do` in core mode (no standard-mode flag needed).
- `mem_do(action="increment_access", params={"id": "…"})` → `"Error: invalid parameter for action 'increment_access' — mem_increment_access() got an unexpected keyword argument 'id'. Use action='help' …"`. The action **was dispatched** — the error is a kwarg-name mismatch (`id` vs whatever parent now expects), not an "unknown action" routing failure. STM's actual feedback path passes the parent-expected kwarg, so this isn't an STM regression; it's just what happens when I probe with a synthetic kwarg from the outside.

**Tool inventory (core mode)**
9 tools: `mem_add, mem_do, mem_index, mem_list, mem_read, mem_recall, mem_search, mem_stats, mem_status`. All core MCP tools present; STM's surfacing client only needs `mem_search` + `mem_do`.

**pytest**
`uv run pytest -m "not ollama"` with all four isolation env vars set → **1445 passed in 20.64s**, 0 failed. No `WARNING: invalid JSON` in the log. Most tests use `tests/_fake_memtomem_server.py` (explicit `SurfacingConfig(ltm_mcp_command=sys.executable, ltm_mcp_args=[_FAKE_SERVER])`), so the env var override is defensive for any test that falls back to the default command. The MCP-level checks above are the ones that actually exercise the real parent.

**Gaps / followups**
- None actionable. Green smoke → no `feedback_` / `project_` memory churn per the B2 deliverable rule. B3 design can proceed.
- Skipped per B0: minimum-version matrix (no hard parent dep declared; B0 (2)).

**Not performed / out of scope**
- Changing STM code (B2 is read-only; the plan says stop and re-plan if code changes).
- Any mutating calls (`mem_add`, `mem_index`) against the real parent — the contract we're re-validating is read/empty-shape only.

PR #210, opened 2026-04-20
feat(compression): timeout-bound LLM compression stage (#207)

### LINKED ISSUE #207: feat(compression): timeout-bound LLM compression stages (P1a follow-up to #206)
## Background

Follow-up to #206, which added `call_timeout_seconds` + `overall_deadline_seconds` for upstream `call_tool` waits. That PR deliberately left the **pipeline stages** (CLEAN → COMPRESS → SURFACE → INDEX) out of scope. CLEAN and INDEX are fast; SURFACE already has a 3s `asyncio.wait_for` at `surfacing/engine.py:158-161`. **COMPRESS is the gap.**

## The gap

`LLMCompressor.compress()` and `_apply_hybrid()` (dispatched at `manager.py:506-680`) `await` the LLM client directly with no timeout. A slow or hung LLM endpoint freezes the proxy call chain AFTER the upstream has already responded — so #206's `call_timeout_seconds` doesn't cover it. The hang shifts from the upstream into the pipeline.

## Doesn't #206's deadline cover this?

No — `overall_deadline_seconds` scopes the **upstream retry loop**. The compression stage runs post-upstream, outside the `wait_for` wrapper. This is a different wait that needs its own timeout.

## Why separate from #206

- Different scope: #206 is about **upstream transport liveness**; this is about **pipeline stage liveness**.
- Different config surface: compression timeouts belong in `LLMCompressorConfig` / `HybridConfig`, not `UpstreamServerConfig`.
- Different failure handling: a compression timeout should degrade gracefully (passthrough + log), not propagate as an MCP error — whereas #206 propagates `TimeoutError` to the caller.

## Proposed change

- Add `llm_timeout_seconds` (suggested default `60.0`) to `LLMCompressorConfig` and `HybridConfig`.
- Wrap `LLMCompressor.compress()` and `_apply_hybrid()` in `asyncio.wait_for`.
- On `TimeoutError`: log a WARNING, record a metric (new `ErrorCategory.COMPRESSION_TIMEOUT`), fall through to passthrough content.

## Files

- `src/memtomem_stm/proxy/compression.py` — `LLMCompressor.compress()`
- `src/memtomem_stm/proxy/manager.py:506-680` — dispatch + `_apply_hybrid()`
- `src/memtomem_stm/proxy/config.py` — `LLMCompressorConfig`, `HybridConfig`
- `src/memtomem_stm/proxy/metrics.py` — `ErrorCategory` enum

## Tests

- Fake LLM client that `await asyncio.sleep(10)` → timeout fires, passthrough content returned, metric recorded.
- Normal path: compression completes within timeout, output matches existing expectations.
- Regression: existing `test_compression.py`, `test_auto_compression.py`, and `test_compression_ratio_guard.py` must still pass.

memtomem:
## Summary

- Add `llm_timeout_seconds` (default 60.0) to `LLMCompressorConfig` and wrap `LLMCompressor._call_api` in `asyncio.wait_for`.
- On timeout: circuit-breaker failure + WARNING log + `last_fallback="timeout"` + fallback to `TruncateCompressor`. Existing plumbing surfaces this as `compression_strategy = "llm_summary→timeout_fallback"` in `proxy_metrics.db`, matching the `privacy` / `circuit_breaker` / `llm_error` / `closed` pattern.
- Add `ErrorCategory.COMPRESSION_TIMEOUT` to the enum for the metrics taxonomy.

Closes #207.

## Why #206 doesn't cover this

#206 bounds the upstream `session.call_tool` wait via `call_timeout_seconds` / `overall_deadline_seconds`. It deliberately left pipeline stages out of scope. CLEAN / INDEX are fast; SURFACE already has a 3s `asyncio.wait_for` (`surfacing/engine.py:158-161`). `LLMCompressor.compress()` was the remaining gap: a slow or hung LLM endpoint would freeze the proxy call chain AFTER the upstream responded — outside #206's guard.

## Behavior change

- Default 60s ceiling on `LLMCompressor.compress()`. Existing deployments that don't configure `llm_timeout_seconds` inherit this.
- Persistent LLM slowness now trips the circuit breaker in ≤3 calls (was: indefinite hangs), so subsequent calls short-circuit to `circuit_breaker` fallback instead of paying the timeout latency every time.
- `compression_strategy` field in `proxy_metrics.db` gains a new value: `"llm_summary→timeout_fallback"`. Dashboards filtering on `"llm_summary→llm_error_fallback"` now miss timeouts.

## Scope note

Issue #207 proposed `llm_timeout_seconds` on `HybridConfig` too, but `_apply_hybrid` does not await an LLM — it dispatches to `HybridCompressor` which uses `SelectiveCompressor` / `TruncateCompressor` (both sync). Only `LLM_SUMMARY` via `LLMCompressor.compress()` needs the guard. `HybridConfig` field skipped to avoid dead config.

## Why truncate fallback, not passthrough

Issue #207 suggested "passthrough content" on timeout. That would return the uncompressed text verbatim — risking downstream size-limit violations when the whole point of the LLM stage was to shrink oversized content. Truncate matches the existing behavior for `privacy` / `circuit_breaker` / `llm_error` / `closed` — consistent graceful degradation.

## Test plan

- [x] `TestLLMCompressorTimeout::test_timeout_fires_and_falls_back_to_truncate` — patched `_call_api` sleeps 5s, timeout 0.05s fires, `last_fallback == "timeout"`, truncate marker in result.
- [x] `TestLLMCompressorTimeout::test_timeout_counts_as_circuit_breaker_failure` — 3 timeouts open the breaker.
- [x] `TestLLMCompressorTimeout::test_fast_call_does_not_trigger_timeout` — normal path unaffected (no fallback).
- [x] `TestLLMCompressorTimeout::test_config_rejects_non_positive_timeout` — `llm_timeout_seconds=0.0` / `-1.0` raise `ValidationError`.
- [x] Regression: `test_compression.py`, `test_auto_compression.py`, `test_compression_ratio_guard.py`, `test_config_constraints.py` all green (111 tests).
- [x] Full non-ollama CI suite: 1538 passed.
- [x] `ruff check src && ruff format --check src` clean.

Refs #206.

### DIFF
--- a/src/memtomem_stm/proxy/compression.py
+++ b/src/memtomem_stm/proxy/compression.py
@@ -1193,9 +1193,21 @@ class LLMCompressor:
         self._in_flight += 1
         self._idle.clear()
         try:
-            result = await self._call_api(text, max_chars=max_chars)
+            result = await asyncio.wait_for(
+                self._call_api(text, max_chars=max_chars),
+                timeout=self._cfg.llm_timeout_seconds,
+            )
             self._cb.success()
             return result
+        except asyncio.TimeoutError:
+            self._cb.failure()
+            logger.warning(
+                "LLM compression timed out after %.1fs (strategy=llm/%s), falling back to truncate",
+                self._cfg.llm_timeout_seconds,
+                self._cfg.provider.value,
+            )
+            self.last_fallback = "timeout"
+            return TruncateCompressor().compress(text, max_chars=max_chars)
         except Exception as exc:
             self._cb.failure()
             logger.warning(
--- a/src/memtomem_stm/proxy/config.py
+++ b/src/memtomem_stm/proxy/config.py
@@ -100,6 +100,12 @@ class LLMCompressorConfig(BaseModel):
         "Keep the summary under {max_chars} characters."
     )
     max_tokens: int = Field(default=500, gt=0)
+    # Timeout for a single LLM compression call. A slow or hung LLM endpoint
+    # would otherwise freeze the pipeline AFTER the upstream has already
+    # responded — outside the upstream ``call_timeout_seconds`` (#206).
+    # On timeout the compressor falls back to TruncateCompressor (matching
+    # other LLM failure modes: privacy, circuit_breaker, llm_error).
+    llm_timeout_seconds: float = Field(default=60.0, gt=0.0)
 
     @model_validator(mode="after")
     def _require_api_key_for_hosted_providers(self) -> LLMCompressorConfig:
--- a/src/memtomem_stm/proxy/metrics.py
+++ b/src/memtomem_stm/proxy/metrics.py
@@ -68,11 +68,14 @@ class ErrorCategory(StrEnum):
     """Classification of proxy call errors for metrics tracking."""
 
     TRANSPORT = "transport"  # OSError, ConnectionError, EOFError
-    TIMEOUT = "timeout"  # asyncio.TimeoutError
+    TIMEOUT = "timeout"  # asyncio.TimeoutError on upstream call_tool
     PROTOCOL = "protocol"  # JSON-RPC errors (-32600..-32603)
     UPSTREAM_ERROR = "upstream_error"  # result.isError=True from upstream
     PROGRAMMING = "programming"  # TypeError, AttributeError, etc.
-    INTERNAL_ERROR = "internal_error"  # raised inside the COMPRESS/SURFACE/INDEX pipeline
+    INTERNAL_ERROR = "internal_error"  # raised inside COMPRESS/SURFACE/INDEX pipeline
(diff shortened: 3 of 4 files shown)


memtomem:
Addressed review: dropped dead `ErrorCategory.COMPRESSION_TIMEOUT` in `1f548f6`.

Chose drop over wire-into-`record_error` because compression timeout is a graceful fallback — the call succeeds and returns a truncated result. Wiring it through `record_error` would inflate `_total_errors` / `_errors_by_category` for what is, by design, a successful degraded response. The signal already lives in `compression_strategy = "llm_summary→timeout_fallback"` (same pattern as `llm_error_fallback` / `privacy_fallback` / etc.); dashboards filter via `WHERE compression_strategy LIKE '%timeout_fallback%'`. Enum re-addition is cheap if a dedicated counter becomes warranted later.

Issue #204, opened 2026-04-20
progressive: complete trace_id propagation missing from #173


memtomem:
## Background

PR #173 introduced a keyword-only `trace_id` on `ProxyManager.call_tool` and documented that *"the value flows verbatim through the span metadata and into `proxy_metrics`/`surfacing_events` rows"*. The progressive delivery path was not included in that enumeration — in practice `_apply_progressive` receives `server`/`tool` but drops them, never sees `trace_id`, and the random UUID `key` that identifies a progressive entry never escapes the function. As a result `stm_proxy_read_more` is the one path in the call pipeline where Langfuse cannot correlate a revisit to its originating call.

Surfaced while scoping `progressive_reads` metrics (planned follow-up): at that point the absence of call-origin context became a hard blocker for stratified analysis (follow-up rate by tool, by strategy, by response size), so the context-threading fix needs to land before the metrics PR.

## Concrete gap

Current span metadata across the call pipeline (`src/memtomem_stm/proxy/manager.py`):

| span | line | metadata | `trace_id`? | `key`? |
|---|---|---|---|---|
| `proxy_call` (outer) | 875 | `{server, tool, trace_id}` | ✓ | ✗ |
| `proxy_call_clean` | 1186 | `{server, tool}` | ✗ | ✗ |
| `proxy_call_compress` | 1258 | `{server, tool, strategy, max_chars}` | ✗ | ✗ |
| `proxy_call_surface` | 1406 / 1420 | `{server, tool, [path]}` | ✗ | ✗ |
| `proxy_call_index` | 1481 | `{server, tool}` | ✗ | ✗ |

OpenTelemetry nested spans inherit context *within* a single `call_tool` invocation, so inner spans do not need to re-declare `trace_id`. But `stm_proxy_read_more` runs in a **later MCP turn** — a separate invocation with its own root — so parent-child propagation does not carry. The progressive `key` (UUID generated at `manager.py:789`) is the one fact that could stitch the two turns together, and it is never persisted to any span or to `ProgressiveResponse`.

Data layer side, `ProgressiveResponse` (`progressive.py:29-40`) stores `content`, `total_chars`, `total_lines`, `content_type`, `structure_hint`, `created_at`, `ttl_seconds`, `access_count`. No `server`, no `tool`, no `trace_id`. `_apply_progressive` takes `server`/`tool` as parameters but uses them only for first-chunk formatting, then lets them go out of scope.

## Proposed scope (this PR)

[...]

PR #213, opened 2026-04-20
feat(proxy): progressive_reads telemetry + stm_progressive_stats

memtomem:
## Summary

Second and final part of issue #204 — lands the telemetry layer on top of the `trace_id` propagation merged in #205.

- New `progressive_reads` table in `~/.memtomem/stm_feedback.db` (shared file, disjoint namespace — same pattern as `compression_feedback`). One row per initial progressive response plus one row per `stm_proxy_read_more` follow-up.
- New `stm_progressive_stats` MCP tool (parity with `stm_compression_stats` / `stm_surfacing_stats`) reporting total reads, distinct responses, follow-up rate, avg chars served, avg total chars, avg coverage, and a per-tool breakdown.
- Tracker injection on `ProxyManager` mirrors the existing `cache=` / `surfacing_engine=` pattern. Writes are guarded at two layers (non-null check + tracker-side try/except) so a telemetry outage cannot affect the response served to the caller.

Refs #204.

## Behavior change

- `ProxyManager` now performs one SQLite insert per progressive initial-chunk and one per `stm_proxy_read_more` follow-up. Writes go to `~/.memtomem/stm_feedback.db` (table: `progressive_reads`), shared with surfacing and compression feedback via WAL + disjoint tables.
- Default configuration is `proxy.progressive_reads.enabled=true`; setting it to `false` disables initialization and leaves the tool returning `"not enabled"`.
- No new DB file, no new retention policy — unbounded growth for now, monitored via `stm_progressive_stats`.

## Why this matters

Before this PR the progressive path emitted only in-memory counters. The three-row "nudge strength vs call rate" comparison across compression strategies (the motivating analysis in #204) was unmeasurable for the progressive row because no persistent telemetry existed. After merge, every `trace_id` is joinable back to its read events for stratified analysis.

## Test plan

- [x] Unit tests for store aggregation: empty DB, single initial row, full follow-up rate, mixed follow-up rate, tool filter, coverage cap-at-1.0, close idempotency, post-close resilience (14 new cases in `tests/test_progressive_reads.py`).
- [x] Integration tests exercise the `ProxyManager` hot path end-to-end: `_apply_progressive` + `read_more` produce correctly-aggregated rows, no-tracker / closed-tracker do not break the response (4 new cases in `tests/test_progressive.py`).
- [x] MCP handler tests: `not enabled` path, formatted output with per-tool breakdown, tool-filter omits `by_tool` block (3 new cases in `tests/test_server_tools.py`).
- [x] `TestAdvertiseObservabilityFlagEndToEnd` extended to cover the new tool (11 model-facing + observability tools when flag is default-on).
- [x] Full CI filter (`uv run pytest -m "not ollama"`) passes: 1567 / 1567 on the final revision. One flaky pre-existing test (`TestAutoIndexStartupWarning::test_no_warning_when_compression_none`, MCP stdio subprocess race on `echo`) was observed on pre-change main as well, reproducible under the full suite and clean in isolation.
- [x] `uv run ruff check src && uv run ruff format --check src`: clean.
- [x] `uv run mypy src` (advisory): only 2 pre-existing unrelated errors in `proxy/manager.py` (lines 416, 794), neither touched by this PR.

Manual smoke:
```bash
uv run mms call <progressive-triggering-tool>
uv run mms call stm_proxy_read_more key=<k> offset=<n>
sqlite3 ~/.memtomem/stm_feedback.db \
    "SELECT key, offset, chars, served_to, total_chars FROM progressive_reads;"
uv run mms call stm_progressive_stats
```

### DIFF
--- a/src/memtomem_stm/proxy/config.py
+++ b/src/memtomem_stm/proxy/config.py
@@ -336,6 +336,23 @@ class CompressionFeedbackConfig(BaseModel):
     db_path: Path = Path("~/.memtomem/stm_feedback.db")
 
 
+class ProgressiveReadsConfig(BaseModel):
+    """Configuration for progressive-delivery read telemetry.
+
+    Records one row per initial progressive response plus one row per
+    ``stm_proxy_read_more`` follow-up into ``progressive_reads``.
+    Aggregates surface via ``stm_progressive_stats`` and enable
+    stratified analysis of follow-up rate by tool / compression
+    strategy / response size. Shares the user-wide
+    ``~/.memtomem/stm_feedback.db`` file with surfacing and
+    compression feedback (disjoint tables; WAL mode makes concurrent
+    access safe).
+    """
+
+    enabled: bool = True
+    db_path: Path = Path("~/.memtomem/stm_feedback.db")
+
+
 # Static context window sizes (tokens) for known model families.
 # Used by ProxyConfig.effective_max_result_chars() to scale compression budget.
 # Prefix-matched: "claude-sonnet-4-20250514" matches "claude-sonnet-4".
@@ -459,6 +476,7 @@ class ProxyConfig(BaseModel):
     compression_feedback: CompressionFeedbackConfig = Field(
         default_factory=CompressionFeedbackConfig
     )
+    progressive_reads: ProgressiveReadsConfig = Field(default_factory=ProgressiveReadsConfig)
 
     def effective_max_result_chars(self) -> int:
         """Compute max_result_chars scaled by consumer model's context window.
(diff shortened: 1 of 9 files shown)


memtomem:
Thanks for the review — addressed issue #1 in 58c29c6.

**`(no more content)` sentinel guard**: `read_more` now gates the `record_follow_up` call on `PROGRESSIVE_FOOTER_TOKEN` being present in the output, so a sentinel response (offset ≥ len(content)) is a silent pass-through. The prior logic would have recorded a `chars=17, served_to=offset+17` row that inflated `follow_up_rate` and could push `avg_chars_served` past `total_chars`. Regression test: `TestProgressiveReadsTelemetry::test_read_more_past_end_does_not_record_row`.

**`offset` SQLite reserved-word** (issue #2): deferred to a follow-up per the review. No current query is ambiguous; will rename to `chunk_offset` (or quote) the next time the schema changes to avoid a drive-by schema migration here.

Full CI filter clean after the fix (1568 passed).

Issue #164, opened 2026-04-17
Pre-existing mypy errors in proxy/manager.py (pinned)


memtomem:
## Pre-existing mypy errors in `src/memtomem_stm/proxy/manager.py`

Surfaced while working on #160 (progressive footer token). Out of scope for that PR.

- [ ] `manager.py:411` — `Argument "base_url" to "create_scorer" has incompatible type "str | None"; expected "str"` `[arg-type]`
- [ ] `manager.py:773` — `Incompatible types in assignment (expression has type "InMemoryPendingStore", variable has type "SQLitePendingStore")` `[assignment]`

Repro:

```
uv run mypy src
```

Likely fixes:

- **411**: narrow `base_url` before the call (e.g. `assert` / `if base_url is None: ...`), or widen `create_scorer` signature to accept `Optional[str]`.
- **773**: widen the variable annotation to the common `PendingStore` protocol/base, or branch the assignment so each branch binds a consistent type.

Note: these are independent — one fix does not resolve the other. They can be split into two PRs if desired.

mypy is advisory in CI (only `ruff check` + `ruff format --check src` are required), so these have been sitting quietly.

memtomem:
Both checklist items shipped: #374 (part 1, `create_scorer` base_url arg-type via `assert`) and #375 (part 2, PendingStore variable annotation widened to the Protocol) both merged on main (`b79d1bf`, `8e06b9e`).

`uv run mypy src/memtomem_stm/proxy/manager.py` is now clean.

PR #235, opened 2026-04-22
fix(config): validate langfuse package is installed when enabled

### LINKED ISSUE #233: config: silent disable when langfuse.enabled=true but the 'langfuse' extra is not installed
## Background

`LangfuseConfig` (`src/memtomem_stm/config.py:16-33`) already has a symmetry-enforcing validator — `_require_keys_when_enabled` raises `ValueError` at config-load when `enabled=true` but `public_key` / `secret_key` are empty. This catches the common misconfiguration ("turned on tracing but didn't put in credentials").

At startup, `server.py:173-180` then tries to initialize the SDK:

```python
try:
    from memtomem_stm.observability.tracing import init_langfuse
    langfuse_client = init_langfuse(config.langfuse)
except ImportError:
    pass
except Exception:
    logger.warning("Langfuse init failed, continuing without tracing", exc_info=True)
```

The `except ImportError: pass` intentionally swallows the case where the `langfuse` package is not installed, so users who never opt in to tracing don't see a warning.

## Concrete gap

The silent branch is narrower than expected but real: if a user has

```json
{"langfuse": {"enabled": true, "public_key": "pk_...", "secret_key": "sk_..."}}
```

but the `langfuse` extra is not installed in the current interpreter, the config-load validator passes (all required fields are set), and then the runtime `except ImportError: pass` skips initialization **without logging anything**. The server starts, proxy works, but no traces are ever produced. The user has no surface to discover this — `mms status` and `mms health` don't expose tracing state either.

Reproducer:

```bash
uv tool install --reinstall memtomem-stm        # no extras
cat > /tmp/stm.json <<'J'
{"langfuse": {"enabled": true, "public_key": "pk_x", "secret_key": "sk_x"}}
J
mms status --config /tmp/stm.json              # reports enabled=true, no mention that tracing is dead
```

This is the \"config survived a re-install without extras\" scenario — e.g. user reinstalls their tool globally for an unrelated fix and doesn't realize the `[langfuse]` extra got dropped.

## Why not just rely on the existing startup warning

There is no startup warning in this path. The `logger.warning(\"Langfuse init failed, continuing without tracing\", exc_info=True)` branch fires only when the SDK is installed but initialization itself fails (network, auth, invalid host). The `ImportError` branch above it is a silent `pass`.

## Why not `mms health` surfacing or a runtime RuntimeError

- `mms health` surfacing — valid follow-up but separate scope. It's a CLI-visibility layer, not a fail-fast. Doesn't stop a daemon-running server from silently skipping traces until the next `mms health` run.
- Per-call `RuntimeError` (httpx-style, `src/memtomem_stm/proxy/relevance.py:175-178`) — worse UX: server starts OK, then every traced call raises at the point of use.

The right layer is config validation, next to the existing `_require_keys_when_enabled`: environment check belongs adjacent to the schema check. Both are single-responsibility validators, both fail fast at load-time, both point the user at a specific fix.

## Proposed fix

Add a second `@model_validator(mode=\"after\")` on `LangfuseConfig`:

```python
@model_validator(mode=\"after\")
def _require_langfuse_package_when_enabled(self) -> \"LangfuseConfig\":
    if self.enabled:
        from importlib.util import find_spec
        if find_spec(\"langfuse\") is None:
            raise ValueError(
                \"LangfuseConfig.enabled=true but the 'langfuse' package is \"
                \"not installed. Install the langfuse extra \"
                \"(e.g. `uv tool install --reinstall 'memtomem-stm[langfuse]'` \"
                \"or `pip install 'memtomem-stm[langfuse]'`).\"
            )
    return self
```

Unit tests parallel the existing `_require_keys_when_enabled` tests: missing → `ValidationError` with the install hint, present → passes, `enabled=false` → skipped (does not probe `find_spec`).

## Explicitly out of scope for this issue

- `mms health` adding an \"Optional features\" section — separate follow-up if user demand surfaces
- LangChain extra handling — unrelated; the extra is not orphaned, it's for tutorial notebook `04_*_langchain.ipynb`
- Per-call tracer error pathways — already covered by the `except Exception` branch in `server.py:179`

armorbreak001:
When `langfuse.enabled=true` but the `langfuse` extra is not installed, config validation passes silently and tracing is never initialized — with no warning or error surface for the user to discover the problem.

Add a `_require_langfuse_package_when_enabled` validator (alongside the existing `_require_keys_when_enabled`) that uses `importlib.util.find_spec` to check whether the package is importable at config-load time, and raises a clear `ValueError` with install instructions if not.

Fixes #233

### DIFF
--- a/src/memtomem_stm/config.py
+++ b/src/memtomem_stm/config.py
@@ -32,6 +32,19 @@ class LangfuseConfig(BaseModel):
             )
         return self
 
+    @model_validator(mode="after")
+    def _require_langfuse_package_when_enabled(self) -> "LangfuseConfig":
+        if self.enabled:
+            from importlib.util import find_spec
+            if find_spec("langfuse") is None:
+                raise ValueError(
+                    "LangfuseConfig.enabled=true but the 'langfuse' package is "
+                    "not installed. Install the langfuse extra "
+                    "(e.g. `uv tool install --reinstall 'memtomem-stm[langfuse]'` "
+                    "or `pip install 'memtomem-stm[langfuse]'`)."
+                )
+        return self
+
 
 class STMConfig(BaseSettings):
     model_config = SettingsConfigDict(


memtomem:
Thanks for jumping on this so quickly, @armorbreak001 — and sorry for the timing. We had #234 implementing the same validator in flight; it merged about 4 minutes after you opened this PR and shipped as v0.1.16 today. Your approach matches the accepted fix almost line-for-line, so there's nothing to salvage into a second PR on this axis.

Closing as superseded by #234. Please don't let this discourage you — if you're up for another, #164 (pre-existing mypy errors in `proxy/manager.py`) is open and unclaimed. It's smaller in scope than it looks; the file is already pinned out of the mypy run, so the goal is just to peel the ignores off one by one with real type fixes.

One tip for next time: the CLA bot needs the commit author email to be linked to your GitHub account (Settings → Emails). The CLA check failed here because your commit email isn't associated with a GitHub user — either add it to your GitHub email settings, or `git commit --amend --reset-author` with a GitHub-linked email before pushing. That'll save a round-trip next PR.

Thanks again!

PR #238, opened 2026-04-23
docs: sync CONTRIBUTING pytest filter with CI and flag macOS-only Desktop discovery

pandas-studio:
## Summary

Quick-review of the public docs surfaced two small drifts, fixed together in one
PR with regression tests so they can't silently reappear.

- **`CONTRIBUTING.md` pytest command did not match CI.** CONTRIBUTING quoted
  `pytest -m "not ollama"`; the CI `test` job (`.github/workflows/ci.yml:41`)
  runs `pytest -m "not ollama and not bench_qa_meta and not bench_qa_llm_judge"`.
  A new contributor running the CONTRIBUTING command locally hits the
  `bench_qa_meta` self-tests (intentional failures, per the `pyproject.toml`
  marker description) and the `bench_qa_llm_judge` scenarios (which require
  `OPENAI_API_KEY` / `ANTHROPIC_API_KEY`), both of which look like real
  regressions. Matched CI verbatim and dropped the misleading
  "full suite (requires running Ollama)" follow-up line.

- **`docs/cli.md` implied `mms add --import` Claude Desktop scan was OS-aware.**
  The paragraph on `--from-clients` called the Claude Desktop path
  "(OS-appropriate)". That's true of the paste hints `mms init` prints
  (fixed in v0.1.13 via #219), but **not** of the discovery helper:
  `_desktop_config_path()` in `src/memtomem_stm/cli/proxy.py` is pinned to the
  macOS path (`~/Library/Application Support/Claude/...`), so Linux/Windows
  callers of `mms add --import` silently see zero Claude Desktop candidates.
  Replaced the parenthetical with an explicit "Claude Desktop discovery is
  macOS-only — on Linux/Windows register those servers with `mms add` instead"
  note.

## Why these two together

Both are stale wording, both single-line edits, and both lacked any pin to the
source of truth. Rather than land the two edits naked, `tests/test_docs_sync.py`
adds a TDD-style guard for each:

- `test_contributing_pytest_command_matches_ci` extracts the CI filter with a
  regex and asserts CONTRIBUTING quotes it — so future CI filter changes will
  trip CI until CONTRIBUTING follows.
- `test_cli_docs_flag_desktop_discovery_is_macos_only` checks the
  `_desktop_config_path` helper is still macOS-only (no `APPDATA` / `.config/Claude`
  embedded), and if it is, requires `docs/cli.md` to keep a `macOS[- ]only` caveat.
  If the helper ever grows OS-aware variants, the guard fails loud and points
  at both sides to adjust in lockstep.

No runtime / behavior change. v0.1.16 user-facing numbers
(11 MCP tools = 7 `_obs_tool` + 4 `@mcp.tool()`, 10 compression strategies /
5 CLI-exposed) were re-verified against `src/memtomem_stm/server.py` and the
`--compression` Click choice; both match the docs.

## Test plan

- [x] `uv run pytest tests/test_docs_sync.py -v` — both new tests fail before
      the edits, pass after.
- [x] `uv run ruff check tests/test_docs_sync.py` — clean.
- [x] `uv run ruff format --check tests/test_docs_sync.py` — already formatted.
- [ ] Full CI (`.github/workflows/ci.yml`) — delegated to the PR.


pandas-studio:
Review feedback addressed:

- **#1 (regex fragility + multi-job guard)** → `c78961f`. `assert` swapped for explicit `pytest.fail` with a message that points at both this test file and `CONTRIBUTING.md`. Added a second `pytest.fail` when `len(test_job_filters) > 1` so a future CI job also adopting `not ollama` surfaces the ambiguity loudly instead of silently picking `[0]`.

- **#3 (macOS-only gate scope)** → `2d9bee1`. Replaced whole-file `re.search` with a paragraph-scoped check: split `cli.md` on blank lines (markdown paragraph boundaries), keep paragraphs mentioning `--from-clients` / `--import`, assert at least one carries the `macOS[- ]only` caveat. Paragraph scope rather than ±N-line window per your note, so prose restructuring inside the same paragraph won't false-fail. Mutation-verified locally: stripping the caveat from the current `--import` paragraph and planting `macOS-only` elsewhere in the file correctly fails the new check (the old whole-file regex passed).

- **#4 (CI green)** → pre-review run was 6/6 green (`lint`, `typecheck`, `test`, `notebooks`, `bench_qa`, `CLAAssistant`); new commits trigger a fresh run.

- **#2 (cli.md Note-block restructuring)** → skipping per your note that it's pure prose preference. Current wording is factually correct; a Note-block follow-up can land separately if the paragraph starts feeling long in practice.

memtomem (approved):
Done.

Issue #226, opened 2026-04-21
mms add --import: offer to prune from source client after import


memtomem:
## Problem

After `mms add --import` succeeds, the upstream MCP server stays registered in its source client (Claude Code user/local/project scope, Claude Desktop). The same tools are then advertised on two paths — direct from the source client and via STM's proxied namespace — and the direct path bypasses compression, caching, and LTM surfacing entirely.

#202 covered the diagnosis; #203 surfaced a **post-import warning** pointing the user at the manual cleanup command. This issue proposes the next step: offering to do the prune itself.

## Why now (signal from the field)

Real user hit this today:

```
~/.claude.json mcpServers:      ['shadcn','playwright','memtomem','docs-langchain','langfuse-docs','memtomem-stm']
~/.memtomem/stm_proxy.json:     ['docs-langchain','langfuse-docs','Context7']
```

`/mcp` in Claude Code accordingly shows both `docs-langchain` (direct) **and** `langchain__*` tools (proxied via STM) — the exact dual-surface that the #203 warning was meant to preempt. The warning did its job (documentation), but the user still ended up on the dual-path because the manual `claude mcp remove …` step is a separate action they have to remember after a config-write flow that otherwise felt done.

This flips the wait-for-signal state tracked internally after #203 — real dual-registration has been observed, so the prune affordance that was deferred is now warranted.

## Didn't #203 already fix this?

#203 is a **warning** — it prints a hint after import but takes no action. That was the right call for the first incident-free pass. The gap now visible is the last-mile automation: doing the prune, not just explaining it.

## Why a separate concern from #203

- **Scope**: #203 was a docs/UX change inside the existing import flow. Prune is a **config-write** into a different client's file (or a shell-out to `claude mcp remove`) — a new failure surface that deserves its own review and test coverage.
- **Invariant**: the STM-discovery-is-read-only invariant (no writes to source-client config) is load-bearing. Breaking it requires an explicit user action, which is what this issue designs.

## Design options

Two shapes, both acceptable — the tradeoff is a single extra prompt vs. opt-in via flag:

### A) Interactive prompt (default)

After a successful import, prompt:

> Imported `docs-langchain` via STM. Remove the direct registration from Claude Code (user scope)? \[y/N\]

- Default **No** — preserves the read-only invariant; only writes back on explicit consent.
- Skipped on non-TTY (pipe, CI, `--no-tui`).
[...]

PR #241, opened 2026-04-23
feat(cli): add `mms prune` + `init --prune-originals` to collapse dual-reg

memtomem:
## Summary

Close the dual-registration gap left by `mms init` (and by `mms add --import` without `--prune`): onboarding with `mms init --mcp claude` imports source-client upstreams into STM and registers STM with the client, but the original direct registrations stay in place — so tool calls bypass STM's compression, caching, and LTM surfacing whenever the direct path is taken.

Two entry points, both built on the existing `_handle_source_prune` / `_prune_imported_candidates` pipeline:

1. **`mms prune [NAMES...] [--all] [--yes] [--dry-run]`** — standalone post-hoc pruner. Refuses to run without explicit scope (`--all` or `NAMES`); unknown names abort before any writes; non-TTY callers must pass `--yes`; partial failures exit non-zero with per-row manual-command fallback.
2. **`mms init --prune-originals`** — opt-in flag for scripted onboarding; on TTY a single y/N prompt (default No). Non-TTY without the flag preserves the #203 hint-only default (regression-guarded).

New helper `_find_dual_registered` intersects STM upstream names with `_discover_candidates` output so the prune call site feeds directly into the shared source-removal machinery. STM's own `stm_proxy.json` is never touched; only source-client files change.

## Invariants preserved

- **bootstrap-only `mms init`** (#200): init still writes the STM config once. The new flag only adds a post-save step; it doesn't re-enter the wizard.
- **read-only discovery by default** (#203): scan + hint remains the non-opt-in path. `--prune-originals` / `mms prune` are the explicit opt-ins.
- **writer parity with `mms add --import --prune`** (#226): same `_prune_imported_candidates` pipeline, same per-source scope mapping, same non-fatal partial-failure handling — just surfaced via a new exit-code contract for the standalone command.

## Test plan
- [x] 14 new tests for `mms prune` covering: usage errors (no scope / `--all` + NAMES), missing config, no-op paths, NAMES filter with unknown-name rejection, `--all`, `--dry-run`, `--yes`, non-TTY error, TTY accept/decline, `duplicate_in`, partial failure exit non-zero.
- [x] 4 new tests for `mms init --prune-originals`: scripted non-TTY path, TTY accept, TTY decline, and a regression guard that the existing non-flag non-TTY default keeps the hint-only behavior.
- [x] `TestInitImportFlow` autouse fixture extended to also stub `_handle_source_prune` — keeps the pick-UI coverage focused; prune coverage lives in the new classes.
- [x] Full suite: `uv run pytest -m "not ollama"` → 1647 passed.
- [x] Lint: `uv run ruff check src` + `uv run ruff format --check src` clean.
- [x] Internal-link audit on `docs/cli.md` changes: new `#prune` anchor resolves.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -35,6 +35,7 @@ Commands:
   health    Check upstream server connectivity.
   init      Guided first-time setup for memtomem-stm.
   list      List configured upstream servers.
+  prune     Remove dual-registered upstreams from source MCP clients.
   register  Register memtomem-stm with an MCP client.
   remove    Remove an upstream MCP server from the proxy configuration.
   status    Show proxy gateway configuration and server list.
@@ -61,6 +62,11 @@ Options:
                             'json' = write .mcp.json, 'skip' = no
                             registration. Omit the flag for the interactive
                             prompt.
+  --prune-originals         After a successful import + registration, remove
+                            the direct registrations from source MCP clients
+                            so tools route through STM only. TTY callers get
+                            a single y/N prompt (default No); non-TTY
+                            scripted callers must pass the flag explicitly.
 ```
 
 Interactive wizard for the first-time setup. Prompts for a single upstream server (name, prefix, transport, command/URL), optionally probes connectivity, writes the config, then offers a 3-way MCP-client registration prompt:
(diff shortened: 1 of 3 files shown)


memtomem:
Thanks for the review — pushed a follow-up commit (2dcc8d6) addressing the actionable items:

**#1 path-expansion** — `data = _load(path)` → `_load(resolved)`. `_load` actually expands internally, so the original path wasn't a functional bug, but the one-liner removes the double-expansion redundancy and is defensive against future `_load` refactors.

**#2 identity check** — Implemented. `_find_dual_registered` now requires a name **and** `_server_signature` match, mirroring the dedup `_add_from_clients` uses in the opposite direction so the import↔prune round-trip agrees on what "the same server" is. Divergent-identity same-name pairs are skipped. Missing-signature entries fall back to name-only (kept for users who onboarded via `init` and never edited). New test `test_divergent_identity_not_dual_registered` locks it in.

**#3 preview/action contract** — Added a comment at the preview-iteration site pointing at the shared `source + duplicate_in` contract with `_prune_imported_candidates`, referencing `test_duplicate_in_sources_all_pruned` as the pin.

**#4 cosmetic** — Dropped the hard-coded `:<20`; width is now computed from the set. Preview also surfaces the source entry's command/URL via `_format_candidate_detail` so users can eyeball what they're about to remove (complements the stricter identity check: even though the filter now skips divergent entries, showing the command keeps the user in the loop).

**#5 `--config` help string** — Leaving as-is. Checked project style: `mms init`, `mms add`, `mms health`, `mms list` all use bare `--config TEXT [default: ...]` with no help string. Only `mms register` adds one (because the "must already exist" nuance isn't captured by the default). Matching the majority.

**#6 `--dry-run` + NAMES** — Added `test_dry_run_with_names_filters_before_preview` to lock in the composition: preview header count reflects the filtered subset, not the full set.

Docs: prune section now documents the identity requirement. Full CLI suite 165/165 (13 → 14 base + 2 new).

PR #254, opened 2026-04-24
feat(metrics): persist error message text in proxy_metrics.db (#253)

### LINKED ISSUE #253: metrics: persist upstream/protocol error message text in proxy_metrics.db
## Problem

When a proxied tool call fails, `proxy_metrics.db` records `error_category` and `error_code` but **not the error message text**. The message *is* propagated to the MCP caller (`ToolError(original_text)` at `src/memtomem_stm/proxy/manager.py:1309`) so the agent sees it once at call time — but it's never persisted. Post-mortem the row only says "this call failed with category X", not *why*.

This is asymmetric with how internal pipeline stages already record error text:

| Source | Category | Message stored? |
|---|---|---|
| INDEX stage raise | `internal_error` | ✅ `index_error` |
| EXTRACT stage raise | `internal_error` | ✅ `extract_error` |
| SURFACE stage raise | `internal_error` | ✅ `surface_error` |
| Upstream `result.isError=True` | `upstream_error` | ❌ |
| JSON-RPC `-32600..-32603` | `protocol` | ❌ (only `error_code`) |
| Transport / timeout | `transport` / `timeout` | ❌ |
| Programming bug | `programming` | ❌ |

## Concrete gap (real evidence)

Today `mcp__memtomem-stm__langfuse__getLangfuseDocsPage` failed 4× in two clusters:

```
2026-04-25 08:09:46 | protocol       | -32602 | (text lost)
2026-04-25 08:35:09 | upstream_error |   —    | (text lost, 123 bytes)
2026-04-25 08:35:10 | upstream_error |   —    | (text lost, 107 bytes)
2026-04-25 08:35:11 | upstream_error |   —    | (text lost, 120 bytes)
```

`-32602` is JSON-RPC "Invalid params" — most likely a wrong argument name from the caller. The 3 subsequent `upstream_error` rows could be the same bug retried, or a separate Langfuse-side rejection (rate limit, missing slug, etc.). **Without re-running the call, the DB cannot distinguish these hypotheses.**

This is the second time in two weeks an after-the-fact debugging session has been blocked on missing error text (the first one led to operator running ad-hoc reproductions).

## Why now, preempting three reviewer questions

- **\"Doesn't the caller already see the message?\"** Yes, at call time. But the metrics DB is the only **post-mortem** surface — `mms status`, future operator commands, dashboards. The asymmetry with `index_error`/`extract_error`/`surface_error` is exactly what makes the gap visible.
- **\"Why not just inspect logs?\"** Protocol and upstream paths don't log the message body at INFO; only retry warnings include it (and get truncated). DEBUG isn't on in normal operation.
- **\"Could this leak sensitive data?\"** Only if the upstream tool echoes user input in its error string. Same risk class as the existing `*_error` text columns. Mitigations: (1) cap length at 500 chars; (2) column lives in `~/.memtomem/proxy_metrics.db` (operator-local), never returned over MCP.

## Proposed change

Self-contained, ~30-50 LOC + tests:

1. **Schema** (`src/memtomem_stm/proxy/metrics_store.py:_migrate`) — add `error_message TEXT DEFAULT NULL` to the idempotent migration list. Insert into `record()`.
2. **Dataclass** (`src/memtomem_stm/proxy/metrics.py`) — add `error_message: str | None = None` to `CallMetrics`.
3. **Capture sites** in `src/memtomem_stm/proxy/manager.py`:
   - L1150 `PROGRAMMING`, L1170 `PROTOCOL`, L1199 `TIMEOUT`/`TRANSPORT` — `error_message=f\"{type(exc).__name__}: {exc}\"[:500]`
   - L1301 `UPSTREAM_ERROR` — `error_message=original_text[:500]`
4. **Test** (`tests/proxy/test_metrics_store.py` or sibling) — one parametrized test that triggers each category and asserts `error_message` is non-NULL and truncated correctly.

## Out of scope

- No new query path, CLI command, or `mms status` field — separate follow-up once the column exists and we know what queries are useful.
- No surfacing change in MCP tool responses — `ToolError` text already reaches the caller.
- No structured field extraction; opaque text is enough.

memtomem:
Closes #253.

## Summary

- Adds `error_message TEXT DEFAULT NULL` column to `proxy_metrics.db`, populated at all five capture sites in `ProxyManager.call_tool` (PROGRAMMING / PROTOCOL / TRANSPORT / TIMEOUT / UPSTREAM_ERROR).
- Capped at 500 chars (`MAX_ERROR_MESSAGE_CHARS` in `proxy/metrics.py`) — long enough to retain typical JSON-RPC envelopes, short enough that a chatty upstream cannot blow up the DB.
- Migration is idempotent (matches the existing per-column `PRAGMA table_info` pattern), so existing `~/.memtomem/proxy_metrics.db` files migrate transparently on next start.

## Behavior change

Operator-visible only: `proxy_metrics.db` now stores the error text alongside `error_category` / `error_code`. No surfacing change in MCP responses (the same text already reaches the caller via `ToolError(original_text)` for `UPSTREAM_ERROR`, or via re-raised exception for the other categories).

Removes the asymmetry where `index_error` / `extract_error` / `surface_error` already stored TEXT but the upstream/protocol/transport/timeout categories did not.

## Test plan

- [x] `tests/test_error_metrics.py::TestErrorMessagePersistence` — one test per ErrorCategory triggers the failure path through `ProxyManager.call_tool` with a real `MetricsStore`, then SQL-asserts the persisted `error_message` matches the expected text.
- [x] Truncation test: a >`MAX_ERROR_MESSAGE_CHARS` upstream message is capped at the boundary.
- [x] Success path persists `error_message=NULL`.
- [x] Schema layer: fresh-DB column existence + migration of a pre-existing schema-less DB.
- [x] Full suite: `uv run pytest -m "not ollama"` → 1675 passed.
- [x] `uv run ruff check src tests` → clean. `ruff format` of touched files → clean.

## Out of scope

- No new query path, CLI command, or `mms status` field — separate follow-up once the column exists and we know what queries are useful.
- No surfacing change in MCP tool responses.
- No structured field extraction; opaque text is enough for the post-mortem use case.

### DIFF
--- a/src/memtomem_stm/proxy/manager.py
+++ b/src/memtomem_stm/proxy/manager.py
@@ -65,7 +65,12 @@ from memtomem_stm.proxy.progressive import (
 )
 from memtomem_stm.proxy.progressive_reads import ProgressiveReadsTracker
 from memtomem_stm.proxy._locks import LockTimeoutError, bounded_lock
-from memtomem_stm.proxy.metrics import CallMetrics, ErrorCategory, TokenTracker
+from memtomem_stm.proxy.metrics import (
+    MAX_ERROR_MESSAGE_CHARS,
+    CallMetrics,
+    ErrorCategory,
+    TokenTracker,
+)
 from memtomem_stm.observability.tracing import traced
 
 # JSON-RPC error codes that indicate bad input, not connection problems.
@@ -1115,6 +1120,9 @@ class ProxyManager:
                         compressed_chars=0,
                         is_error=True,
                         error_category=ErrorCategory.TIMEOUT,
+                        error_message=(f"{type(deadline_exc).__name__}: {deadline_exc}")[
+                            :MAX_ERROR_MESSAGE_CHARS
+                        ],
                         trace_id=trace_id,
                     )
                 )
@@ -1148,6 +1156,9 @@ class ProxyManager:
                             compressed_chars=0,
                             is_error=True,
                             error_category=ErrorCategory.PROGRAMMING,
+                            error_message=(f"{type(exc).__name__}: {exc}")[
+                                :MAX_ERROR_MESSAGE_CHARS
+                            ],
                             trace_id=trace_id,
                         )
                     )
@@ -1169,6 +1180,9 @@ class ProxyManager:
                             is_error=True,
                             error_category=ErrorCategory.PROTOCOL,
                             error_code=err_code,
+                            error_message=(f"{type(exc).__name__}: {exc}")[
+                                :MAX_ERROR_MESSAGE_CHARS
+                            ],
                             trace_id=trace_id,
                         )
                     )
@@ -1197,6 +1211,9 @@ class ProxyManager:
                             compressed_chars=0,
                             is_error=True,
                             error_category=cat,
+                            error_message=(f"{type(exc).__name__}: {exc}")[
+                                :MAX_ERROR_MESSAGE_CHARS
+                            ],
                             trace_id=trace_id,
                         )
                     )
@@ -1299,6 +1316,7 @@ class ProxyManager:
                     compressed_chars=len(original_text),
                     is_error=True,
                     error_category=ErrorCategory.UPSTREAM_ERROR,
+                    error_message=original_text[:MAX_ERROR_MESSAGE_CHARS],
                     trace_id=trace_id,
                 )
             )
--- a/src/memtomem_stm/proxy/manager.py
+++ b/src/memtomem_stm/proxy/manager.py
@@ -70,6 +70,7 @@ from memtomem_stm.proxy.metrics import (
     CallMetrics,
     ErrorCategory,
     TokenTracker,
+    format_error_message_from_exc,
 )
 from memtomem_stm.observability.tracing import traced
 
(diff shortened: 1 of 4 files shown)


memtomem:
Thanks for the review. All three points addressed in `1688909`:

**Required — LOCK_TIMEOUT coverage.** Added `error_message=format_error_message_from_exc(exc)` at the 6th capture site in `_call_tool` (the outer wrapper that handles pipeline-stage exceptions). The site sets `pipeline_category` to either `LOCK_TIMEOUT` or `INTERNAL_ERROR` — both branches now persist the text. INTERNAL_ERROR is technically defensible via the existing `index_error`/`extract_error`/`surface_error` columns, but populating `error_message` in the same call costs nothing and gives operators a single column to query for any pipeline-stage failure text. New test `test_lock_timeout_persists_message` triggers the path via `LockTimeoutError("stm_lock", 5.0)` and asserts the row carries the lock name. Companion test `test_internal_error_persists_message` covers the INTERNAL_ERROR branch via a compress-stage `RuntimeError`.

**Helper extraction.** Lifted `(f"{type(exc).__name__}: {exc}")[:MAX_ERROR_MESSAGE_CHARS]` to `format_error_message_from_exc(exc)` in `proxy/metrics.py` next to `MAX_ERROR_MESSAGE_CHARS`. All 6 exception sites now use the helper. UPSTREAM_ERROR keeps its inline `original_text[:MAX_ERROR_MESSAGE_CHARS]` because its source is the upstream tool's text, not a Python exception — different shape, single site, not worth a second helper.

**Privacy posture comment.** Added to the `error_message` dataclass field comment: notes the column is operator-local (never returned over MCP), the cap limits worst-case leakage size, and the content is *not* sanitized — same risk class as the existing `index_error`/`extract_error`/`surface_error` columns. Explicit "don't add a redactor here without reviewing all four sites together" so future readers don't break the symmetry.

Full suite green: `uv run pytest -m "not ollama"` → 1677 passed (+2 new tests). `ruff check` + `ruff format` clean.

PR #256, opened 2026-04-26
feat(surfacing): expose skip-reason / outcome / cache counters in stm_surfacing_stats (Phase 1)

memtomem:
## Summary

- New `SurfacingObservability` class — in-memory per-tool counters for the 9 skip reasons + 4 outcomes + cache hit/miss covered by the surfacing pipeline.
- `RelevanceGate` records its 5 sub-reasons (`gate_excluded_tool`, `gate_write_tool`, `gate_tool_disabled`, `gate_rate_limit`, `gate_cooldown`) directly. `SurfacingEngine` records the 4 engine-level skips (`disabled`, `response_too_short`, `circuit_open`, `no_query`), the 4 post-search no-result paths (`no_results_score`, `no_results_dedup`, `no_results_invalidated`, `no_results_empty_cache`), the 4 outcomes (`surfaced_cache_hit`, `surfaced_cache_miss`, `error_timeout`, `error_other`), and cache hit/miss on every cache lookup.
- `stm_surfacing_stats` output extended with three new sections (`Skip reasons`, `Outcomes`, `Cache`) appended after the existing DB-backed stats.
- Phase 1 of the surfacing observability + UX RFC in `memtomem-docs/memtomem-stm/planning/stm-surfacing-observability-ux-rfc.md`. Phases 2–5 are documented as wait-for-signal with explicit reopen triggers.

## Why now

The surfacing engine itself is production-grade — race-condition-safe, fail-safe, auto-tuned, per-tool override (#247), 7 gating conditions, 3 injection modes. The remaining gap is not in the engine; it is in what an operator can *see and act on*. Before this PR, every reject path logged at DEBUG and disappeared, leaving operators no quantitative path from "surfacing feels quiet" to a root cause. This unblocks evidence-based decisions on the rest of the RFC roadmap: truncation policy (Phase 3) needs to know how often `max_injection_chars` actually fires; implicit feedback (Phase 4) needs to compare against the current explicit-feedback rate; auto-tune saturation warning (Phase 2) needs the saturation event to be visible in the first place.

## Behavior change

- `stm_surfacing_stats` output is a **strict superset**. Every existing line, label, and order is preserved byte-for-byte. The new sections are suppressed entirely when no surfacing call has been recorded, so zero-traffic deployments and scripted callers parsing the legacy fields keep working unchanged.
- New keyword-only `observability=` parameter on `SurfacingEngine.__init__` and `RelevanceGate.__init__`. Defaults to `None`. Existing embedding callers and test fixtures that omit the argument see no behavior change.
- No SQLite migration. No env-var change. No new MCP tool. Counters are in-memory only and reset on process restart, mirroring the in-process aggregate shape of `stm_compression_stats` / `stm_progressive_stats`.

## Resolution invariants

- Each `surface()` invocation records exactly one skip OR one outcome (no double-counting). Verified by `test_no_double_counting_per_call`.
- The gate is responsible for recording its own 5 reject reasons. The engine records `disabled` and `no_query` upstream of the gate; recording either of those at the gate would double-count. Verified by `test_disabled_does_not_record_at_gate` and `test_query_none_does_not_record_at_gate`.
- Cache hit/miss is incremented on every cache lookup independent of what the lookup ultimately renders — a cache hit that returns an empty/invalidated entry still counts as a cache hit even though the outcome is a no-results skip. The two axes are independent so the cache hit ratio reflects raw cache behavior, not post-filter results.

## Test plan

- [x] `uv run ruff format --check src` — clean (new module reformatted in-place)
- [x] `uv run mypy src` — 2 pre-existing errors in `proxy/manager.py` (issue #164), no new errors introduced
- [x] `uv run pytest -m "not ollama and not bench_qa_meta and not bench_qa_llm_judge"` — 1694 passed, 10 deselected
- [x] New file `tests/test_surfacing_observability.py` — 10 tests covering counter contract + stats formatter (empty snapshot, all three sections, tool filter, descending sort, cache section omission, snapshot independence)
- [x] New class `TestRelevanceGateObservability` in `tests/test_relevance_gate.py` — 7 tests covering all 5 gate sub-reasons + the no-double-counting invariant
- [x] New class `TestSurfacingEngineObservability` in `tests/test_surfacing_engine.py` — 10 tests covering all 4 engine skips, the cache hit/miss path, both error outcomes, and the `observability=None` backwards-compat case

## RFC reference

`memtomem-docs/memtomem-stm/planning/stm-surfacing-observability-ux-rfc.md` (separate cross-repo PR — pending push confirmation).


memtomem:
Follow-up commit `276eba8` addresses the substantive review points from the parallel-session pass — all in the same branch since the PR isn't merged yet (one fewer rebase + reviewer cognitive load lower than splitting).

**Applied:**

- ✅ **NoOp observability pattern** (the biggest item — 18 sites). New `_NOOP_OBSERVABILITY` singleton in `observability.py` is a no-op recording sink so the call sites in `engine.py` (13) and `relevance.py` (5) stay unconditional. The public `observability` property still returns `SurfacingObservability | None` so `stm_surfacing_stats` keeps short-circuiting on absence — only the internal sink became unconditional. Crosses the codebase's 4-site helper-extract threshold cleanly.
- ✅ **`__total__` ordering pinned explicitly.** New `_ordered_tool_keys` helper inserts `__total__` at index 0 regardless of ASCII order. Regression test `test_total_pinned_first_regardless_of_ascii_order` covers a PascalCase tool name (`ReadFile`, 0x52) that would previously have sorted before `__total__` (0x5F).
- ✅ **Inline import moved to module top** in `server.py`, matching the placement of sibling `SurfacingEngine` / `FeedbackTracker` imports.

**Deferred (with rationale):**

- ⏸ **CHANGELOG bullet length.** Kept as-is — STM convention is detail-rich CHANGELOG bullets (see the v0.1.18 entries for reference). The RFC link is already in the bullet for readers who want the structured roadmap. Stylistic, no blocker.
- ⏸ **Unbounded tool-key cardinality (counter dict growth).** Captured as RFC Phase 2+ consideration (LRU eviction). Out of Phase 1 scope and not a current blocker (all proxied tools come from registered upstreams in real deployments — the abuse vector requires a malicious upstream).
- ⏸ **`_TOTAL_KEY` collision with a literal `__total__` tool name.** Addressed via docstring note in `observability.py` (vanishingly low probability + explicit ordering means even a `__`-prefixed tool name no longer buries the aggregate). Switching to a sentinel `object()` would lose JSON serializability of the snapshot, so the string key stays.
- ⏸ **`threading.Lock` over asyncio.** Kept for defense against future thread-pool offload. Cost is zero in single-threaded asyncio contexts (uncontended lock acquire/release in CPython is ~50ns).
- ⏸ **`was_empty = not cached` truthy pattern.** The comment one line above already documents intent; rewriting to `cached is None or len(cached) == 0` would also work but isn't materially clearer. Kept.
- ⏸ **`28.6%` hard-coded literal in test.** Intentional pin as the reviewer notes — if the rounding policy changes, the test should fail and force a deliberate update.

Tests: 1696 pass on the full `not ollama and not bench_qa_meta and not bench_qa_llm_judge` filter, mypy clean (only the 2 pre-existing errors in `proxy/manager.py` tracked by #164).

PR #262, opened 2026-04-27
fix(cli): unify entry-point dispatch so all three boot the MCP server

### LINKED ISSUE #260: mms / memtomem-stm-proxy bare invocation: dispatch parity with memtomem-stm as MCP stdio server
## Summary

The package ships three console entry points that look like aliases:

```
memtomem-stm v0.1.19
- memtomem-stm
- memtomem-stm-proxy
- mms
```

But their **bare-invocation behavior diverges**, which is surprising and breaks MCP-client registrations that pick the wrong name:

| Entry | Bare invocation behavior |
|-------|--------------------------|
| `memtomem-stm` | Boots as MCP stdio server (working processes observed via `ps -A`, registered in `~/.gemini/settings.json` and `~/.claude.json` with no args, all functional) |
| `memtomem-stm-proxy` | Click group help → `exit 2` |
| `mms` | Click group help → `exit 2` |

## Reproduction

```bash
# These three should be aliases, but they aren't symmetric:
memtomem-stm               # → MCP stdio server (waits on stdin for JSON-RPC)
memtomem-stm-proxy         # → Click help, exit 2
mms                        # → Click help, exit 2
```

Concrete user-visible failure: registering the proxy MCP server in Antigravity (or any client) with `command: "mms"` produces `: calling \"initialize\": EOF` because the spawned process prints help to stdout and exits before the MCP `initialize` round-trip completes.

## Why this matters

- The README/docs/`mms register` UX implies `mms` is the CLI surface; users naturally expect `mms` (or `memtomem-stm-proxy`, given its name) to be the MCP server entry too. Today only `memtomem-stm` works for that.
- This mirrors memtomem's `mm ≡ memtomem` invariant (CLAUDE.md states both must resolve to the same callable). STM should pick the same model.

## Proposed fix

Pick one of:

1. **Unify all three entry points to dispatch identically** — bare invocation with non-TTY stdin enters MCP stdio mode; bare with TTY prints help. This matches whatever `memtomem-stm` is doing today.
2. **Make `memtomem-stm-proxy` the canonical MCP server entry**, document `mms` / `memtomem-stm` as CLI-only, and update README + `mcp-clients` docs accordingly. (More invasive — breaks existing user registrations that use `memtomem-stm`.)

(1) seems cleaner and backwards-compatible.

## Repro environment

- macOS 15.4 (Darwin 25.4.0)
- `memtomem-stm 0.1.19` installed via `uv tool install`
- Surfaced while registering mms with Antigravity (Google's VS Code fork, both Gemini-agent path and VS Code-extension path)

pandas-studio:
## Summary

Registering `mms` (or `memtomem-stm-proxy`) as an MCP client's `command` failed with `: calling "initialize": EOF` because those entry points resolved to the Click group, whose bare invocation printed help and exited 2. Only `memtomem-stm` worked because it pointed straight at `server.main`. The three names are documented as interchangeable, so the divergence was a real surprise — surfaced during Antigravity onboarding.

This change makes the Click group `invoke_without_command=True` and dispatches bare calls on `sys.stdin.isatty()`:

- **TTY** (`mms` in a terminal) → prints help, exit 0
- **Non-TTY** (MCP-client stdio pipes, CliRunner, shell pipes) → lazy-imports `memtomem_stm.server.main` and runs it

All three `[project.scripts]` entry points now resolve to `memtomem_stm.cli.proxy:cli` — registering any of them as an MCP server's `command` works identically.

## Why the previous behavior was wrong

`README.md`/docs and the existing `CLAUDE.md` invariant call `mms` and `memtomem-stm-proxy` interchangeable. But:

| Entry | Before this PR |
|-------|----------------|
| `memtomem-stm` | → `server.main` (worked as MCP server, no CLI) |
| `memtomem-stm-proxy` | → Click group help, exit 2 (broken as MCP server) |
| `mms` | → Click group help, exit 2 (broken as MCP server) |

After:

| Entry | After this PR |
|-------|---------------|
| `memtomem-stm` | → `cli.proxy:cli` (TTY=help, non-TTY=server.main; full CLI surface available) |
| `memtomem-stm-proxy` | → `cli.proxy:cli` (same) |
| `mms` | → `cli.proxy:cli` (same) |

## Backward compatibility

- `memtomem-stm` registered in MCP configs (Claude Code, Gemini CLI, Antigravity) → still works (non-TTY → `server.main`).
- `memtomem-stm` invoked in a terminal with no args was previously a hang (waiting for stdio MCP messages). Now prints help — strictly better for that pathological case.
- `memtomem-stm <random_arg>` previously was ignored (server.main doesn't read sys.argv); now Click parses it as a subcommand. Anyone passing extra args to `memtomem-stm` was already in undefined-behavior territory; the new behavior is at worst a clearer error.
- `mms list`, `mms add`, etc. — unchanged (subcommand path skips the new dispatch via the `invoked_subcommand` guard).

## Tests

`TestBareInvocationDispatch` covers all three branches:

- `test_bare_with_tty_prints_help` — TTY → help text on stdout, exit 0
- `test_bare_with_pipe_dispatches_to_mcp_server` — non-TTY → `server.main` is called
- `test_subcommand_skips_dispatch` — `mms version` (subcommand) must NOT trip the dispatch even when stdin is non-TTY (the `invoked_subcommand` guard)

The `_stdin_is_tty()` indirection follows the existing `_should_use_tui` seam pattern (CliRunner replaces `sys.stdin` so monkeypatching the function attribute directly doesn't reach the module's bound reference — `feedback_clirunner_isatty_seam.md` for the parent repo's parallel finding).

## Test plan

- [x] `uv run pytest tests/cli/test_proxy_cli.py` (179/179 pass)
- [x] `uv run pytest -m "not ollama"` full suite (1714/1714 pass on this branch — same as main)
- [x] `uv run ruff check src tests` + `ruff format --check` on touched files
- [ ] Smoke: install from this branch, register as `mms` in an MCP client, observe the proxy boots without `EOF` error

Closes #260.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased]
 
+### Fixed
+
+- **All three console scripts (`memtomem-stm`, `memtomem-stm-proxy`, `mms`) now boot the MCP stdio server when invoked bare with a piped stdin** (#260) — previously only `memtomem-stm` worked as an MCP-server registration target because it pointed straight at `server.main`, while `memtomem-stm-proxy` and `mms` resolved to the Click group whose bare invocation printed help and exited 2. An MCP client config like `{"command": "mms"}` therefore failed with `: calling "initialize": EOF` even though the package documentation calls these names interchangeable. The `cli` Click group now uses `invoke_without_command=True` and dispatches based on `sys.stdin.isatty()`: a TTY still prints the help text (now exit 0, matching Click's default `get_help` flow), and a non-TTY (the MCP-client stdio case, CliRunner, pipes) lazy-imports `memtomem_stm.server.main` and runs it. All three `[project.scripts]` entry points in `pyproject.toml` now resolve to `memtomem_stm.cli.proxy:cli`, so registering any of them in an MCP client's config works identically — closing the parity gap that surprised first-time Antigravity adopters.
+
 ## [0.1.19] — 2026-04-26
 
 ### Added
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -23,8 +23,13 @@ The `ollama` marker auto-skips when Ollama isn't running; CI always uses
 
 - **No Python-level dependency on `memtomem` core.** STM talks to the LTM
   server only through the MCP protocol. Don't `import memtomem` from `src/`.
-- **`mms` ≡ `memtomem-stm-proxy`.** Both entry points in `pyproject.toml`
-  resolve to the same CLI — keep them in sync, don't diverge behavior.
+- **`mms` ≡ `memtomem-stm-proxy` ≡ `memtomem-stm`.** All three
+  `[project.scripts]` entry points in `pyproject.toml` resolve to
+  `memtomem_stm.cli.proxy:cli` (#260). The Click group's
+  `invoke_without_command=True` callback dispatches bare invocations on
+  `sys.stdin.isatty()`: TTY → help, non-TTY → `server.main` (the MCP stdio
+  server). Don't diverge behavior between the three names — registering
+  any of them as an MCP client's `command` must work identically.
 - **Pipeline order is CLEAN → COMPRESS → SURFACE → INDEX** — comments in
   `src/memtomem_stm/proxy/` are the source of truth for the per-stage
   contracts; full architecture write-up lives in the private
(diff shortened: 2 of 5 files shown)


tsdata (approved):
Two regression tests added in 922d312 cover the unknown-subcommand non-TTY path and the --version short-circuit under piped stdin — both invariants worth pinning given the invoke_without_command shift. CI green on both commits. lgtm.

PR #263, opened 2026-04-27
fix(proxy): catch tool name overflow against the 64-char MCP limit

### LINKED ISSUE #261: mms add: validate composed proxied tool name against 64-char MCP regex
## Summary

`mms add` accepts any prefix and any upstream URL without checking whether the **composed proxied tool name** the client will see exceeds the MCP / Anthropic spec limit of `^[a-zA-Z0-9_-]{1,64}$`. When upstream tool names are long, the registration silently \"succeeds\" but the client rejects the offending tools at runtime.

## Reproduction

Existing `~/.memtomem/stm_proxy.json` entry:

```json
\"docs-langchain\": {
  \"prefix\": \"docs_langchain\",
  \"transport\": \"streamable_http\",
  \"url\": \"https://docs.langchain.com/mcp\"
}
```

The upstream `query_docs_filesystem_docs_by_lang_chain` tool (40 chars) gets composed by the client as:

```
mcp_memtomem-stm_docs_langchain__query_docs_filesystem_docs_by_lang_chain
└┬─┘ └────┬────┘ └─────┬──────┘└┬─┘└──────────────┬─────────────────────┘
 4    +   12   + 1 +  14    + 2 +              40                = 73 chars
mcp    server   _  prefix    __  upstream tool name
```

→ Antigravity rejects with:

```
Error: tool name mcp_memtomem-stm_docs_langchain__query_docs_filesystem_docs_by_lang_chain
violates ^[a-zA-Z0-9_-]{1,64}$.
```

Other 6 proxied tools register fine; only the long one breaks. Workaround: shorten prefix to `\"lc\"` → composed name 61 chars.

## Why mms should catch this

- The user's first symptom is a single missing tool buried in a list of 9. Easy to miss; even harder to root-cause without knowing about the 64-char limit.
- mms knows the prefix at `add` time and learns the upstream tool inventory shortly after via initialization handshake (`tools/list`). It can compute `len(\"mcp_<server>_<prefix>__<upstream>\")` for every upstream tool and warn/refuse on overflow.
- The `<server>` segment depends on what name the user picks in their MCP client config — mms can either (a) require the user to pass `--client-server-name <name>` so it can check exactly, or (b) check against a conservative bound (e.g. assume 16-char server name and warn if total ≥ 64).

## Proposed fix

1. After `tools/list` handshake, compute composed tool name for each upstream tool using `\"mcp_X_<prefix>__\" + tool_name` (where X is a placeholder representing the longest reasonable client-side server name, e.g. 16 chars).
2. If any composed name exceeds 64 chars, **emit a warning** during `mms add` and `mms list` showing which upstream tools will be silently dropped by clients.
3. Optionally suggest a shorter prefix that fits.

Stretch goal: `mms doctor` or `mms health --names` flag for explicit on-demand check.

## Repro environment

- `memtomem-stm 0.1.19`
- 3 upstreams configured (context7, langfuse-docs, docs-langchain) — only docs-langchain trips this; context7 and langfuse_docs prefixes are short enough.
- Surfaced via Antigravity registration (\`mcp__memtomem-stm__\` namespace).

## Related

May share a fix surface with #260 (entry point parity).

pandas-studio:
## Summary

Clients (Claude Code, Antigravity, Anthropic SDK) compose proxied tool names as `mcp__<client-server>__<prefix>__<tool>` and silently drop any tool whose composed name exceeds the spec's `^[a-zA-Z0-9_-]{1,64}$` regex. Until now a single overflowing tool from an otherwise-healthy upstream just disappeared from the client's catalogue with no signal — surfaced in production after registering LangChain's 40-char `query_docs_filesystem_docs_by_lang_chain` under a 14-char prefix and watching that one tool go missing.

Three layers of enforcement, escalating from advisory to authoritative:

| Layer | Where | Behavior | Why this layer |
|-------|-------|----------|----------------|
| 1. **`mms add` sanity** | CLI registration | Reject prefix > 42 chars (1-char tool wouldn't fit), warn > 21 chars | Early signal, no upstream contact needed |
| 2. **Boot-time skip** | `_connect_server` | Per-tool overflow check after `tools/list`; log WARNING + skip the offending tool; register siblings | Authoritative — has actual upstream tool inventory |
| 3. **`mms health --names`** | On-demand CLI | Re-runs the boot check via dry-run probe; report-only | Diagnose "one tool from server X is missing" without spinning STM up |

The arithmetic lives in `memtomem_stm.proxy.tool_name_budget` and assumes Claude Code's stricter `mcp__…__…__…` template (3 × `__`) so any false-positives we report are conservative — clients with a leaner format (Antigravity's single-underscore between mcp/server/prefix) would never reject a tool we cleared. `MMS_CLIENT_SERVER_NAME` overrides the default 12-char `memtomem-stm` for users who registered STM under a custom (typically shorter) name.

## README change — recommend `mms` over `memtomem-stm`

The Quickstart now uses `mms` (3 chars) instead of `memtomem-stm` (12 chars) as the registered server name. `mms` saves 9 chars of overhead, which is exactly enough to keep an upstream's longer tool names under the limit (e.g. the LangChain repro: `mms` server + `lc` prefix + 40-char tool = 61 chars; `memtomem-stm` server + `lc` prefix + same tool = 70 chars, over). Both names continue to work — #260's entry-point unification means all three (`mms`, `memtomem-stm`, `memtomem-stm-proxy`) act as the MCP server entry — and the rationale is documented inline so existing users can keep their setup.

## Tests

- `tests/test_tool_name_budget.py` (14) — pin the arithmetic against Claude Code's empirical 75-char render of the user-facing #261 repro, the inclusive 64-char boundary, and the env-var override path
- `TestAddValidation` (4 new) — hard-reject, soft-warn, silent-OK below threshold, env-loosens-when-server-short
- `TestConnectServerOverflowSkip` (2 new) — overflow tool is logged + skipped while siblings register cleanly; short prefix lets the same long tool through
- `TestHealth` (3 new) — `--names` text output, no-op message when clean, `--json --names` schema stability

Local: `uv run pytest -m "not ollama"` 1738 passed, `ruff check src tests` clean, `ruff format --check` clean on touched files.

## Known flake (not introduced by this PR)

`tests/test_proxy_manager.py::TestAutoIndexStartupWarning::test_no_warning_when_compression_none` is intermittently failing (~50% locally) on this branch. Root cause is pre-existing: the test wraps `await mgr.start()` in `try/except Exception`, but the MCP SDK's stdio reader can raise `asyncio.CancelledError` (a `BaseException`) when an `echo` upstream exits before the JSON-RPC handshake completes. The exception then leaks past the catch.

This PR's additional test mass (~22 new tests) shifted async timing enough to surface the fragility more often than on `main`. The fix is orthogonal to #261 — broaden the catch to `BaseException` (or stop spawning real subprocesses for the warn-assertion tests) — and deferred to keep this PR focused on the overflow handling.

## Test plan

- [x] `uv run pytest tests/test_tool_name_budget.py tests/cli/test_proxy_cli.py tests/test_proxy_manager_lifecycle.py -k 'TestAddValidation or TestHealth or TestConnectServer or TestComposed or TestOverflow or TestPrefix or TestClient or TestSpec'` — all 23 new tests pass
- [x] Touched-files format + ruff clean
- [ ] Smoke (post-merge): register a real upstream with a 40-char tool name, observe boot-time skip log + `mms health --names` listing the culprit

Closes #261.


pandas-studio:
Pushed `6b99b9f` addressing the three actionable review items:

1. **Test assertions tightened** — `TestAddValidation` cases now match the actual error template (`is 43 chars`, `max for client server name 'memtomem-stm' is 42`, `Use a shorter --prefix`, `Register STM as 'mms'`) instead of OR-clauses that passed via the second branch. `test_hard_budget_loosens_when_client_server_name_short` no longer relies on a tautology (`"exceeds the budget" not in output` against a string that wasn't in any output) — replaced with a real negative (`"Error:" not in output`, `"max for client server name" not in output`) plus the positive that a relaxed warn still fires.

2. **`prefix_warn_threshold()` now dynamic** — derived from `client_server_name()` the same way `prefix_hard_limit()` is, so the threshold tracks the user's actual server name. Default 12-char server stays at 21 (preserves existing behavior); `mms` (3-char) relaxes to 30. New `test_warn_threshold_loosens_with_short_server_name` pins the relationship.

3. **Empty env var falls back to default** — `MMS_CLIENT_SERVER_NAME=""` (exported but empty) used to drop overhead to 9 bytes and silently permit overflows the user expected to be rejected. New `test_empty_env_var_falls_back_to_default` covers the guard.

Flake follow-up filed as #264 — orthogonal to #261, pre-existing fragility in `TestAutoIndexStartupWarning` that this PR's added test mass made more visible.

Re: review item 4 (mismatched-docs grep) — confirmed clean: `grep -rln '"memtomem-stm"\|memtomem-stm -s user' docs/ notebooks/ CONTRIBUTING.md` returns 0 hits, so the README swap doesn't strand any other registration sites.

CI re-running on the new commit.

tsdata (approved):
Review fixes in 6b99b9f address all three actionable items (test assertion drift, dynamic warn threshold, empty-env guard) plus the empty-env nice-to-have. Both commits green. lgtm.

PR #272, opened 2026-04-29
feat(proxy): instrument INDEX path with stm_index_stats counters

memtomem:
## Summary

- Adds `IndexObservability` — a per-tool counter aggregate for the `extract_and_store` (INDEX) pipeline. Mirrors `SurfacingObservability` shape: per-tool attempts (per call) plus 4-label outcomes (`stored` / `dedup_skip` / `extracted_zero_facts` / `error`). Until now, INDEX activity was visible only as per-call `ExtractOutcome.facts_stored` returns that the manager discarded; no aggregate view existed for operators to answer "is the auto-extraction pipeline actually producing LTM writes, or is it firing on tool calls that have nothing worth storing?".
- Adds `stm_index_stats` MCP tool gated by the existing `MEMTOMEM_STM_ADVERTISE_OBSERVABILITY_TOOLS` flag, advertised after `stm_surfacing_stats`. Default-on; advertised tool count goes from 11 to 12.
- Wires the counter at `extract_and_store` via an optional `observability=` kwarg defaulting to a no-op singleton, so any direct caller works unchanged. `ProxyManager` always instantiates a real `IndexObservability` — the snapshot's `any_call` flag is read at render time so zero-traffic deployments produce no output.

## Why 4 labels and not 2

The split between `stored` / `dedup_skip` / `extracted_zero_facts` / `error` is load-bearing. Fusing `extracted_zero_facts` into `stored=0` would lose the signal "extraction fired but the LLM/heuristic produced nothing", which is architecturally distinct from "facts were produced but were duplicates" (`dedup_skip`) and from "facts were produced and stored" (`stored`). The first is the most informative single observable for whether INDEX is doing useful work on the tool calls it sees — operators should be able to read it directly, not derive it.

## Why a separate observability module instead of extending `SurfacingObservability`

The shapes differ: surfacing tracks `skip_reasons + outcomes + cache` with per-tool sub-dicts on each. INDEX has `attempts` (flat per-tool int) + `outcomes` (per-tool dict). Forcing one struct to host both would either bloat the surfacing snapshot with INDEX-shaped fields or compromise INDEX into the surfacing shape. Keeping them as parallel modules with matching ergonomics (same `_TOTAL_KEY`, same `any_call` semantics, same no-op stand-in pattern) gives operators the symmetry without the coupling.

## Why no quality dimension

`SurfacingObservability` correlates with `surfacing-feedback` to ask "did the agent find this useful?". `IndexObservability` has no equivalent and intentionally none. Any quality signal for INDEX would have to choose between two non-equivalent observables — *positive-value* (does an extracted fact later get retrieved on a related query?) and *harm-prevented* (what fraction of writes would have leaked sensitive content absent redaction?). Bundling them in one signal would conflate distinct questions; deferring lets a future quality-signal PR scope to one or both deliberately. The schema-level absence is intentional and self-documents the open scope.

## Behavior change

- `ProxyManager.__init__` now constructs an unconditional in-memory counter on every instantiation (one `defaultdict` + one `threading.Lock`; negligible cost). Previously, no such state existed.
- `stm_index_stats` appears in `tools/list` for clients with `MEMTOMEM_STM_ADVERTISE_OBSERVABILITY_TOOLS=true` (the default). Pinned by `tests/test_server_tools.py::TestAdvertiseObservabilityFlagEndToEnd::test_default_advertises_all_twelve` (count assertion 11 → 12) and by the `_STM_UTILITY_TOOL_NAMES` exhaustiveness regression in `test_utility_tool_names_tuple_matches_registered_set`.
- `extract_and_store` gains a kwarg-only `observability=` parameter with a no-op default. The free function's existing direct call sites (only `ProxyManager._extract_and_store`) are updated to pass the manager-owned instance; no other callers exist in `src/` or `tests/`.

## Test plan

- [x] `tests/test_index_observability.py` — 12 new tests covering counter contract, snapshot copy independence, no-op singleton, format helper output, ASCII-order pin, descending outcome sort. All pass.
- [x] `tests/test_server_tools.py` — advertise-flag, advertise-order, and exhaustiveness regressions all updated and passing (43/43).
- [x] Full CI filter `uv run pytest -m "not ollama"` — 1763/1763 pass.
- [x] `uv run ruff check src` — clean.
- [x] `uv run ruff format --check src` — clean.
- [ ] `uv run mypy src` — 2 pre-existing errors at `manager.py:451` and `:829`, unrelated to this PR (verified via `git stash` round-trip on main).

### DIFF
--- a/src/memtomem_stm/proxy/index_observability.py
+++ b/src/memtomem_stm/proxy/index_observability.py
@@ -1,0 +1,128 @@
+"""In-memory counters for STM-driven LTM writes (the INDEX path).
+
+Mirrors ``surfacing/observability.py`` in spirit and shape so ``stm_index_stats``
+and ``stm_surfacing_stats`` stay symmetric at the operator surface — but
+**deliberately omits a quality dimension**. SURFACE has a quality signal
+(``surfacing-feedback``) that lets operators ask "did the agent find this
+useful?". INDEX has no equivalent. The absence is a design finding, not
+an oversight: a quality dimension for INDEX would have to choose between
+two non-equivalent observables — positive-value (does an extracted fact
+later get retrieved on a related query?) and harm-prevented (what fraction
+of writes would have leaked sensitive content absent redaction?). Bundling
+them in one signal would conflate two questions; deferring lets a future
+quality-signal PR scope to one or both deliberately. Until then, the
+schema-level asymmetry self-documents the architectural state.
+
+State is per-process in-memory only. Counters reset on restart, mirroring
+``SurfacingObservability``. The cross-call aggregate is the load-bearing
+addition over ``ExtractOutcome.facts_stored``, which is per-call only and
+discards as soon as the caller drops the return value.
+
+Snapshot output is a flat dict so ``server.py::stm_index_stats`` formats
+without coupling to internal data structures.
+"""
+
+from __future__ import annotations
+
+import threading
+from collections import defaultdict
+from typing import Literal
+
+# Per-fact OR per-call outcome labels (mixed granularity by design):
+# - ``stored`` / ``dedup_skip``: per fact processed inside one call.
+# - ``extracted_zero_facts``: per call, terminal — extractor returned ``[]``
+#   so the loop body never ran.
+# - ``error``: per error event. Outer ``extractor.extract`` failure counts
+#   once (per call); per-fact ``index_file`` failure counts per fact.
+#
+# Math contract for ``snapshot()``:
+#   sum(outcomes["__total__"][k] for k in ALL) is **not** equal to
+#   ``attempts["__total__"]`` because per-fact outcomes can multi-count
+#   one attempt. Useful operator ratios:
+#   - ``outcomes["__total__"]["stored"]`` = literal mem_add count into LTM
+#     per process — the raw write-volume signal across windows. Callers
+#     compute per-turn or per-second rates at the read site if needed.
+#   - ``outcomes["__total__"]["extracted_zero_facts"] / attempts["__total__"]``
+#     = fraction of calls where INDEX produced zero LTM writes — the
+#     direct measurement of "extraction fired on a tool call but found
+#     nothing worth storing," architecturally distinct from "facts existed
+#     but were duplicates" (``dedup_skip``) and from "facts existed and
+#     were stored" (``stored``).
+IndexOutcome = Literal[
+    "stored",
+    "dedup_skip",
+    "extracted_zero_facts",
+    "error",
+]
+
+_TOTAL_KEY = "__total__"
+
+
+class IndexObservability:
+    """Aggregate per-tool counters for the INDEX (extract_and_store) pipeline.
+
+    Thread-safe via a single coarse lock — same trade-off as
+    ``SurfacingObservability``: O(1) dict writes, contention dwarfed by
+    the LTM round-trip the counters observe.
+    """
+
+    def __init__(self) -> None:
+        self._lock = threading.Lock()
+        self._attempts: dict[str, int] = defaultdict(int)
+        self._outcomes: dict[str, dict[str, int]] = defaultdict(lambda: defaultdict(int))
+        self._any_call = False
+
+    def record_attempt(self, tool: str) -> None:
+        """Record one ``extract_and_store`` invocation. Per-call.
+
+        Increments per-tool counter and the ``__total__`` aggregate.
+        """
+        with self._lock:
+            self._any_call = True
+            self._attempts[tool] += 1
+            self._attempts[_TOTAL_KEY] += 1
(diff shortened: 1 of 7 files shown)


memtomem:
Thanks for the review. Addressed in two follow-up commits:

**#1 (medium) — wire-in unpinned: fixed** (`f7ccd8b`)
Added a snapshot assertion at the end of `TestExtractAndStore::test_dedup_skips_duplicate` that pins `record_attempt` per-tool, `__total__` aggregate, and the `dedup_skip` outcome. Removing the `observability=self.index_observability` kwarg from `manager._extract_and_store` now fails this test. Also fixed the `test_index_observability.py` module docstring to reference the actual covering test (the prior `test_extract_and_store*.py` reference pointed at a non-existent file) and tightened the fragile `index("  t:")` slice to `split("\n  t:\n", 1)[1]` per #6.

**#3 (low) — traced() span position: fixed** (`9aa2943`)
Moved the `snapshot()` read and `any_call` short-circuit inside `with traced(...)`. Both sibling tools now emit a Langfuse span on every invocation regardless of whether the body has data to render. The proxy-disabled early return stays outside the span (no payload to carry).

**#2 (low) — `_NoOpIndexObservability` private leak in public type annotation: deferred**
Identical pattern to `surfacing/observability` (`SurfacingObservability | _NoOpObservability`). Adding it to the existing tracing-Protocol-refactor follow-up note as another reopen trigger so the future Protocol pass batches both observability modules at once.

**#4 (low, optional) — runtime literal validation: hold**
Single internal call site, mypy already catches mistypes there. `get_args` membership check would add value once a second call site appears or after the first observed mistype. YAGNI for now.

**#5 (informational) — format helper near-duplication: hold**
Per CLAUDE.md "three similar lines is better than premature abstraction." Two render helpers, different snapshot shapes (`skip_reasons + cache` vs `attempts`). The shared per-tool ordering already lives in `_ordered_tool_keys`. Trigger for the helper extract is a third observability module appearing.

Full CI filter (`uv run pytest -m "not ollama"`): 1763/1763. Ruff src clean. Pre-existing mypy errors unchanged.

memtomem:
Round 2: addressed sibling review #1 (medium, gating).

**#1 — auto_index_response scope: fixed via option (b) full coverage** (`d0f7713`)

Verified two LTM-write paths fire on the same proxy hot path (stages 4 and 4b in `manager.call_tool`); the initial commit covered only `extract_and_store`. Extended `IndexObservability` to both paths:

- `record_attempt(tool, path)` with `path ∈ {"extract", "auto_index"}` — `attempts` snapshot becomes `{tool: {path: count}}` so per-path call distribution is visible per tool. `outcomes` stays flat 4-label across both paths (auto_index uses only `stored`/`error`; the other two are extract-specific). Raw mem_add count reads as `outcomes[__total__][stored]` regardless of which path drove the write — path attribution lives in `attempts`.
- `auto_index_response` gains the same kwarg-only `observability=` parameter (no-op default) and records `record_attempt(tool, "auto_index")` + `stored`/`error` based on the inner `index_file` try/except.
- `manager._auto_index_response` passes `self.index_observability`, symmetric to extract.
- `_format_index_observability_sections` renders per-path attempts breakdown under each tool block (descending by count, same sort key as outcomes).
- `stm_index_stats` docstring + empty-state message now name both paths explicitly.

Wire-in pins extended: `TestAutoIndex::test_writes_file_and_indexes` asserts `attempts[tool] == {"auto_index": 1}` + `outcomes[tool] == {"stored": 1}`; new `test_auto_index_records_error_outcome_on_index_file_failure` pins the inner-try/except `error` branch so a future refactor can't silently drop it.

**#2 — `index_engine is None` outcome-skip: held**
Manager guards at `manager.py:1725` (`self._index_engine is not None`) — the path is unreachable from production. Test fixtures that pass `index_engine=None` directly cause an empty `outcomes` row but a non-zero `attempts` row, which is the expected disparity for a structural-invariant violation. Adding a label or early-return here would either hide the disparity or change disk-write behavior. Documented via the comment in `extract_and_store`'s loop.

**#3 — Public-signature private type: deferred** (unchanged)
Already in the RecordingSink Protocol follow-up trigger list. Not addressing in this PR.

**Minor 1 — `__slots__` intent**: noted in the no-op class docstring (mirrors surfacing's reason; instance-dict bloat unnecessary for module-level singletons).

**Minor 2 — `attempts == 0` defensive branch**: confirmed YAGNI; `record_attempt` always increments by 1.

**Minor 3 — split-on-"\n  t:\n" sentinel fragility**: added an explicit comment in the test pointing future contributors at the sentinel and the conditions under which it would need bumping (a label or path key starting with "t").

Validation: 1767 tests pass (was 1763 — +4 new), ruff src clean, RFC-leak grep clean.

PR #282, opened 2026-05-01
feat(mms): W2 PR3 — `mms host status` (drift bucket + UX)

memtomem:
## Summary

- Adds `mms host status`, the read-only inspection surface for W2's host-sync work. Read-only, never mutates registry / sidecar / anywhere; exit code always 0 (drift is observation, not a CI failure signal).
- Classifies every registry entry into one of four buckets relative to the W2 sidecar baseline: `unchanged` / `changed` / `removed_at_host` / `no_baseline`. Lives in a new `host` click group reserved by RFC §7.3 for the upcoming `scan` and `sync` commands.
- `--json` output includes every state in `entries[]` and a four-key `summary` (zero counters explicit). Text output's main table only shows `unchanged` + `changed`; the other two are footer counts with neutral phrasing — PR4 promotes `removed_at_host` to the main table without renaming anything.

## Why four states (not two)

`import_state.entries[name]` can KeyError along three real paths that PR2's backfill alone doesn't close:

1. Manual sidecar edits (RFC §3 explicitly allows hand-editing these files; the sidecar isn't exempt).
2. Atomic-write race windows (`--apply` writes registry then sidecar; PR3 status called in between).
3. Pre-PR2 installs that upgrade to the wired mms but call `mms host status` before any `mms import --apply`.

Read-only inspection's first PR can't surface those as an unhandled traceback. `no_baseline` is the recovery bucket with an actionable hint (`run \`mms import --apply\` to stamp`). The same bucket also guards future `HASH_VERSION` bumps — PR3 is the first consumer of `drift_hash_version`, so without the gate here a v1 baseline vs v2 binary would silently miscompare as ""always changed"".

## source_label-preferred matching

When the same entry name appears in multiple host configs, the comparison candidate is chosen by `source_label` match against the sidecar baseline first, falling back to discover()'s `ALL_HOSTS` first-seen order. This keeps the comparison axis pinned to the import-time host — mirroring an entry into a second host doesn't flicker between `unchanged` and `changed` based on iteration order. Both arms (matched + fallback) are tested explicitly.

## What this is *not*

Each has an explicit reopen trigger documented either in the module docstring or in code comments:

- Per-key diff renderer (""which field changed""): wait-for-signal — user reports `"changed but I don't know what"` or one round of internal dogfooding.
- `changed` row actionable hints: PR3 is observation only; the `no_baseline` hint is invariant-violation recovery, a different category.
- `--strict` / exit-non-zero-on-drift flag: wait-for-signal.
- `mms host scan` / `mms host sync`: separate PRs in the same group.
- W3+ `--force` re-stamp existing sidecar rows.

## Test plan

- [x] `uv run pytest tests/cli/test_mms_host.py` — 14 passed, all four states explicitly covered.
- [x] `uv run pytest -m \"not ollama\"` — 2009 passed (full regression green).
- [x] `uv run ruff check src tests && uv run ruff format --check src/memtomem_stm/cli/mms_host.py tests/cli/test_mms_host.py src/memtomem_stm/cli/proxy.py` — clean.
- [x] `uv run mypy src/memtomem_stm/cli/mms_host.py` — clean (advisory; pre-existing `proxy/manager.py` errors unchanged).
- [x] Manual smoke: seed claude-code → `mms import --apply` → `mms host status` (all unchanged); edit host config → `mms host status` (one changed); delete entry from host → footer note + 0 changed at host; manually delete sidecar entry → footer note with actionable hint.

### DIFF
--- a/src/memtomem_stm/cli/mms_host.py
+++ b/src/memtomem_stm/cli/mms_host.py
@@ -129,8 +129,18 @@ def _classify(
     """
     cand_by_name = _select_candidate_by_name(candidates, import_state)
     rows: list[dict] = []
-    for name, server in registry.servers.items():
+    for name in registry.servers:
         baseline = import_state.entries.get(name)
+        cand = cand_by_name.get(name)
+        # ``current_hash`` is uniformly the host's view: the canonical
+        # hash of the matched candidate, or ``None`` if no host scanner
+        # finds the entry. This keeps the field's meaning the same
+        # across every bucket — downstream tooling that recomputes
+        # against the host config gets the same value back regardless
+        # of whether the row is unchanged / changed / removed / has no
+        # baseline.
+        current_hash = compute_drift_hash(cand.server) if cand is not None else None
+
         if baseline is None or baseline.drift_hash_version != HASH_VERSION:
             # ``no_baseline`` covers both the missing-row case and the
             # version-mismatch case — they both mean "the comparison
@@ -142,12 +152,11 @@ def _classify(
                     "state": "no_baseline",
                     "source_label": baseline.source_label if baseline else None,
                     "baseline_hash": baseline.drift_hash if baseline else None,
-                    "current_hash": compute_drift_hash(server),
+                    "current_hash": current_hash,
                     "last_imported": baseline.last_imported if baseline else None,
                 }
             )
             continue
-        cand = cand_by_name.get(name)
         if cand is None:
             rows.append(
                 {
@@ -160,7 +169,6 @@ def _classify(
                 }
             )
             continue
-        current_hash = compute_drift_hash(cand.server)
         rows.append(
             {
                 "name": name,
@@ -189,7 +197,10 @@ def _render_text(rows: list[dict]) -> None:
     comparison signal. Each is conditionally shown when its count > 0.
 
     Column widths follow ``mms project list`` (mms_project.py:285):
-    plain f-strings, no rich.
+    plain f-strings, no rich. Alignment is best-effort — names longer
+    than the NAME column overflow rather than truncate. Not a contract;
+    swap in textual/rich/whatever when MCP names regularly exceed 20
+    chars in practice.
     """
     main_rows = [r for r in rows if r["state"] in ("unchanged", "changed")]
     n_unchanged = sum(1 for r in main_rows if r["state"] == "unchanged")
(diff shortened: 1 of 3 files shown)


memtomem:
Thanks for the thorough review. Disposition:

## Fixed in `65319a6`

**Medium — `current_hash` semantic asymmetry.** Unified to host view across all four buckets: ``current_hash = compute_drift_hash(cand.server) if cand else None``, computed once before bucket selection. Now means the same thing regardless of state — host canonical hash, or null if no host has it. Two new tests pin both halves:

- `test_current_hash_is_host_view_across_buckets` — registry shape A, host shape B, sidecar dropped → `no_baseline` row's `current_hash` is host's hash (B), not registry's (A). Asserts the asymmetry the review flagged.
- `test_current_hash_null_when_host_absent` — sidecar dropped + no host has entry → `current_hash` is null in `no_baseline` (parallel to the existing `removed_at_host` null contract).

`test_baseline_hash_matches_canonical` only covered the unchanged case where both views happen to be equal, which is exactly why the asymmetry slipped through — kept as-is for the canonical-hash anchor, complemented by the two new tests for the asymmetry contract.

**Low — test 3's `main_table` split is fragile.** Dropped. The JSON assertion below already covers `state == "removed_at_host"` strongly; the text slice was redundant signal at the cost of brittleness to footer rewording (which PR4 may evolve).

**Very low — malformed host config exit-0 pin.** Added `test_status_malformed_host_config_exit_0` — corrupts `~/.claude.json` and asserts exit 0 in both text and JSON modes. Pins the inheritance from `_read_json_safely`'s catch against a future scanner refactor.

**Cosmetic — column overflow.** Added one-line note in `_render_text`'s docstring: "Alignment is best-effort — names longer than the NAME column overflow rather than truncate. Not a contract."

## Not fixing

**Low — empty-registry stream asymmetry.** `mms_host.py` writing the empty-registry message to stdout is the right sibling to match. The closer analogue is `mms project list` (`mms_project.py:281`), which also writes to stdout for the empty-state case and exits 0 — both are read-only "show me what's there" commands where empty is normal. The `mms_project.py:306` reference in the review is `enable_cmd` (paired with `Exit(1)`), which is the *error* path. The split "info → stdout, error → stderr" is followed correctly here.

**Very low — `_select_candidate_by_name` Pass 1 O(N×M).** Agreed it reads as draft; leaving as-is. At ≤4 hosts × handful of entries per host, dwarfed by `discover()` itself. Reopen trigger: any user-visible latency on `mms host status`, or registry growing past ~50 entries — whichever first. The `(name, source_label) → cand` index in one pass is the obvious refactor.

## Test count

PR3 ships 17 tests (was 14 in the original commit; +3 from this review).

PR #294, opened 2026-05-02
fix(surfacing): heal MCP session after outer-timeout cancellation

### LINKED ISSUE #290: surfacing: outer wait_for cancellation does not reconnect LTM MCP session
Found during the 2026-05-02 implementation review (\`docs/reports/memtomem-stm-implementation-review-2026-05-02.md\`, finding #3).

## Symptom

\`SurfacingEngine.surface()\` wraps the LTM call in \`asyncio.wait_for\`:

- \`src/memtomem_stm/surfacing/engine.py:184-198\`

When the outer timeout fires, the inner \`McpClientSearchAdapter._session.call_tool(\"mem_search\", ...)\` is cancelled mid-RPC. The adapter's reconnect path catches only:

- \`src/memtomem_stm/surfacing/mcp_client.py:262\` — \`_TRANSPORT_ERRORS = (OSError, ConnectionError, EOFError, BrokenPipeError, asyncio.TimeoutError)\`

\`asyncio.CancelledError\` is **not** in that tuple. So a cancelled \`mem_search\` propagates out without reconnecting, leaving the MCP session mid-message. Subsequent calls may hang or return stale data.

For comparison, the proxy main path handles cancel/timeout symmetrically with reconnect:

- \`src/memtomem_stm/proxy/manager.py:1162\`

## Recommendation

Move timeout handling into \`McpClientSearchAdapter.search()\` (so cancellation is local), or catch \`asyncio.CancelledError\` at the adapter boundary, reconnect, and re-raise. Apply the same pattern to \`mem_do(increment_access)\` and any other adapter methods that share \`_TRANSPORT_ERRORS\`.

## Test

Use a fake session whose \`call_tool\` blocks on \`asyncio.sleep\`, set \`timeout_seconds\` very low, assert that after the timeout (a) the session was reconnected (or torn down), (b) the next surfacing call succeeds.

## Scope

Small-to-medium (~20 lines + test).

memtomem:
Closes #290.

## Summary

- `SurfacingEngine.surface()` wraps the LTM call in `asyncio.wait_for(..., timeout=...)`. When that outer timeout fires, the inner `mem_search` `call_tool` is cancelled mid-RPC and the MCP session is left in a half-read state. The adapter's `_TRANSPORT_ERRORS` deliberately does not catch `asyncio.CancelledError` (cooperative cancellation must propagate so `wait_for` can surface its `TimeoutError`), so previously a cancelled call left the session unhealthy with no recovery — the next surfacing cycle could hang or see out-of-order responses on the same stream.
- `McpClientSearchAdapter` now sets a `_needs_reconnect` flag in `except asyncio.CancelledError` and re-raises; a new `_heal_if_needed()` helper called at the top of `search()` / `increment_access()` / `scratch_list()` lazily reconnects before issuing a fresh RPC. If the lazy reconnect itself fails, the call returns empty (matching the existing `_session is None` path) and the flag is preserved so a later attempt can retry.
- Five regression tests in `test_mcp_client_reconnect.py::TestOuterCancellationLazyReconnect` pin the round-trip.

## Behavior change

- After an outer-timeout cancellation, the **next** surfacing call (or `increment_access` / `scratch_list`) tears down and re-establishes the MCP session before its first RPC. Adds one stdio reconnect cycle to that next call only — typical reconnect cost is a single subprocess restart already exercised by the existing transport-error path.
- No change for happy-path calls or for transport-error-driven reconnects.

## Test plan

- [x] `uv run pytest tests/test_mcp_client_reconnect.py tests/test_surfacing_engine.py tests/test_remote_ltm_integration.py tests/test_core_format_contract.py -q` — 136 passed
- [x] `uv run pytest tests/ -m "not ollama" -q` — 2087 passed
- [x] `uv run ruff format --check` on touched files — clean


pandas-studio (approved):
LGTM

PR #305, opened 2026-05-06
test(windows): set_home helper + 10-site HOME-patch sweep (#302 P1b)

memtomem:
## Summary

- Adds `tests/helpers.py::set_home(monkeypatch, path)` patching both `HOME` and `USERPROFILE` (memtomem #714 idiom). On Windows, `Path.home()` and `os.path.expanduser("~")` consult `USERPROFILE` first, so fixtures that monkeypatched only `HOME` leaked the developer's real profile into temp-dir-based tests on the new `windows-latest` CI leg.
- Migrates the 10 sites enumerated in the [#302 P1b inventory](https://github.com/memtomem/memtomem-stm/issues/302) — `tests/cli/test_mms_w1_acceptance.py` (×3), `tests/cli/test_mms_host.py`, `tests/cli/test_mms_project.py`, `tests/cli/test_mms_import.py`, `tests/cli/test_proxy_cli.py`, `tests/mms/test_state.py`, `tests/mms/test_import_hosts.py`, `tests/test_progressive_reads.py`. No behavior change on POSIX; removes the leak on Windows.
- Adds `tests/test_helpers.py` with a direct contract test so a future regression that drops the `USERPROFILE` patch fails here rather than via a flaky Windows job.

Refs #302 (P1b in the suggested PR sequence).

## Test plan

- [x] `uv run ruff check src tests` — clean
- [x] `uv run ruff format --check src` — clean (CI-required scope)
- [x] `uv run pytest -m "not ollama"` — 2079 passed, 7 skipped, locally on macOS
- [ ] Watch the `windows-latest` matrix leg on this PR (continue-on-error from #304) — expect a meaningful drop in HOME-leak failures

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #302, opened 2026-05-06
Windows support: triage from memtomem core fixes


memtomem:
**Tags:** `windows` `meta`

## Context

The sibling repo `memtomem/memtomem` (LTM core) has merged 35+ PRs and tracked 17+ issues addressing Windows compatibility over the past quarter. This issue triages those findings against the current `memtomem-stm` codebase to (a) capture confirmed gaps that warrant follow-up PRs, (b) record categories that are structurally **not applicable** so future contributors don't re-investigate, and (c) flag upstream/out-of-scope items.

This issue itself is the single tracking source. Each P-row below should become its own focused PR (or be explicitly closed as wontfix).

## Inventory of memtomem core Windows fixes (reference)

| Category | Representative PRs/issues | Core change |
|---|---|---|
| File locking | #625, #652, #817, #818, #448, #456 | `fcntl` → `portalocker`, signal/atexit split |
| HOME/USERPROFILE | #644, #714 | `tests/helpers.py::set_home()` helper + 133-site sweep |
| Path separators | #316/#716, #647/#717, #643/#711/#713 | `os.sep`, `replace("\\", "/")`, `Path` / `as_posix()` for compares |
| Encoding | #728 (merged), #759 (open) | Windows console stdout/stderr `reconfigure(encoding="utf-8", errors="replace")`; explicit `encoding="utf-8"` on `read_text` / `subprocess` |
| Runtime dir / NTFS | #637/#738, #645/#739, #732/#733 | Gate `XDG_RUNTIME_DIR` on `os.name != "nt"`, relax NTFS mode-bit checks, mtime µs-ceiling |
| Signals | #817/#818 | POSIX guard around `signal.signal(SIGTERM)` + `atexit` fallback |
| Tests / CI | #634/#638, #657, #713/#721-726/#729 | `windows-latest` CI matrix, AST guard against POSIX-only imports, `skipif(sys.platform=="win32")` patterns |

(Issue numbers refer to `memtomem/memtomem`, not this repo.)

### P0 — `_run_claude_mcp` subprocess missing `encoding=`

- **Location:** `src/memtomem_stm/cli/proxy.py:142`
- **Current:** `subprocess.run(cmd, capture_output=True, text=True, timeout=timeout)`
- **Fix:** add `encoding="utf-8", errors="replace"`.
[...]

PR #311, opened 2026-05-06
fix(cli/proxy): backslash-safe `--args` tokenization on Windows (#302 P1e)

memtomem:
## Summary

- `entry["args"] = shlex.split(args_str)` in `add` / `init` ran POSIX shlex on the raw argument string, which on Windows ate `\a`, `\t`, `\_` etc. as escapes. A path like `D:\a\repo\repo\tests\_fake_memtomem_server.py` came out as `D:arepotests_fake_memtomem_server.py`, was persisted into the JSON config, and was then handed to `asyncio.create_subprocess_exec` as a relative script path — surfacing as `[Errno 2] No such file` in `tests/cli/test_proxy_cli.py::TestAddValidate::test_validate_success_writes_entry_with_tool_count` on the windows-latest leg added in #304.
- Adds `_split_args` in `src/memtomem_stm/cli/proxy.py`. POSIX still calls plain `shlex.split` (no behavior change). On Windows we drive a `shlex.shlex` instance with `posix=True` + `escape=""` so quoting (`"hello world"` → `hello world`) keeps working but backslashes are preserved verbatim. `ValueError` on unclosed quotes is preserved so the existing `except ValueError` branches in `add` / `init` still catch malformed input. Both call sites (`proxy.py:802`, `proxy.py:1808`) are migrated.
- Adds `tests/cli/test_proxy_cli.py::TestSplitArgs` covering the POSIX branch, the literal CI repro path, quoted-whitespace tokens on Windows, the `ValueError` contract, and an end-to-end `mms add --args <win path>` that asserts the saved JSON config keeps the path intact. `sys.platform` is monkeypatched so the Windows branch runs on POSIX too; the live windows-latest leg covers the real interpreter on top.

This is a production bug exposed (not introduced) by the windows-latest matrix in #304 — the affected test is `TestAddValidate::test_validate_success_writes_entry_with_tool_count`, which was untouched in #305 (P1b) and unrelated to the NTFS mode-bit assertions in #306 (P1c).

Refs #302 (suggested as P1e in the sequence).

## Test plan

- [x] `uv run ruff format --check src` — clean (CI-required scope)
- [x] `uv run mypy src/memtomem_stm/cli/proxy.py` — clean
- [x] `uv run pytest tests/cli/test_proxy_cli.py -m "not ollama"` — 199 passed
- [x] `uv run pytest -m "not ollama and not bench_qa_meta and not bench_qa_llm_judge"` — 2083 passed locally on macOS
- [ ] Watch `windows-latest` matrix (continue-on-error from #304): `TestAddValidate::test_validate_success_writes_entry_with_tool_count` should turn green; the new `TestSplitArgs` class should be all-pass on the real interpreter.

Generated with Claude Code.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -30,12 +30,25 @@ def test_set_home_path_home_resolves_to_sandbox(tmp_path: Path, monkeypatch):
 
 def test_set_home_is_undone_after_test(tmp_path: Path, monkeypatch):
     """monkeypatch must restore both env vars when the test exits — otherwise
-    the next test in the same session inherits the sandbox HOME."""
-    original_home = os.environ.get("HOME")
-    original_userprofile = os.environ.get("USERPROFILE")
+    the next test in the same session inherits the sandbox HOME.
+
+    Note that ``os.environ.get("VAR") == None`` would silently pass even if
+    monkeypatch had failed to *delete* a previously-absent key (since both
+    cases ``get`` returns ``None``). When the original is absent, assert
+    the key is not in ``os.environ`` so a regression that left a None-string
+    behind would still trip — important for bare CI shells where ``HOME``
+    or ``USERPROFILE`` may not be pre-populated.
+    """
+    snapshots = {
+        var: ("HAS" if var in os.environ else "ABSENT", os.environ.get(var))
+        for var in ("HOME", "USERPROFILE")
+    }
 
     set_home(monkeypatch, tmp_path)
     monkeypatch.undo()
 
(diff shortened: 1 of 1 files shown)


pandas-studio:
Wrong head branch — `windows-p1b-followup-recovery` was already merged via #310 and contains `tests/test_helpers.py` changes, not the intended P1e `shlex.split` fix. Reopening with the actual P1e diff (`src/memtomem_stm/cli/proxy.py` + `tests/cli/test_proxy_cli.py`) under a new PR.

PR #324, opened 2026-05-16
feat(surfacing): thread _context_query through SurfacingEngine (F1)

memtomem:
## Summary

`ProxyManager._call_tool_inner` extracted `_context_query` from upstream args for the compression path but **stripped it from the args dict passed to surfacing** (`src/memtomem_stm/proxy/manager.py:1146-1150`). `ContextExtractor.extract_query()` had a legacy branch that prefers `arguments["_context_query"]`, but in the proxy path that key was already gone — so surfacing fell back to heuristics (path tokenization, semantic keys, tool name) and searched LTM with thin queries instead of the agent's actual intent.

This PR threads the agent-provided context query to surfacing via an explicit `context_query` keyword, without re-coupling to `upstream_args`.

## Approach

Added optional `context_query: str | None = None` keyword parameter to:

- `ContextExtractor.extract_query()` — **new priority-2 branch**, between the per-tool template (priority 1) and the legacy in-arguments branch (now priority 3).
- `SurfacingEngine.surface()` — forwards to `extract_query`.
- `ProxyManager._apply_surfacing()` / `_apply_surfacing_on_progressive()` — forward to `surface()`.

Wired all four proxy call sites:

| Path | File:Line | How it gets `context_query` |
|---|---|---|
| Cache hit | `manager.py:_on_cache_hit` | Extracts from incoming `arguments` locally; does not mutate the args dict |
| Primary progressive | `manager.py:_call_tool_inner` | Already-extracted local |
| Progressive fallback | `manager.py:_call_tool_inner` | Already-extracted local |
| Non-progressive | `manager.py:_call_tool_inner` | Already-extracted local |

`upstream_args` is **not** modified — the new kwarg is the only delivery path. The control signal stays out of the args bag so it never leaks downstream to the proxied tool.

### Extractor priority (final order)

1. Per-tool `query_template` — unchanged top precedence.
2. **NEW** — explicit `context_query` kwarg, when non-empty after strip.
3. Legacy `arguments["_context_query"]` — kept as fallback for direct engine callers and tests.
4. Heuristic extraction.
5. Tool name fallback.

Backwards-compatible: the legacy branch stays so existing tests in `test_context_extractor.py` and `test_query_aware.py` pass unchanged.

## Bench cleanup

`tests/bench/test_bench_qa_scenarios.py::test_s10_surfacing_recall_at_k` previously passed both `path` (workaround to clear `min_query_tokens`) and `_context_query` (stripped before reaching extractor). After F1 the `_context_query` reaches the extractor directly, so the `path` workaround is dropped. Docstring updated to reflect the new wiring.

## Out of scope (deferred)

- **F2** — `min_score` default change (experiment-driven, separate PR).
- **F3** — truncation knobs (`result_content_max_chars`, `preview_max_chars`); shares files but a separate concern.
- **F4** — docs/compression.md fix; already shipped in #323 (`eaa5d30`).

## Test plan

- [x] `uv run pytest -m "not ollama" tests/test_context_extractor.py tests/test_surfacing_engine.py tests/test_query_aware.py tests/test_proxy_manager_pipeline.py` — 167 passed (new tests: 11).
- [x] `uv run pytest -m "not ollama" tests/bench/test_bench_qa_scenarios.py::test_s10_surfacing_recall_at_k` — 1 passed with `path` workaround removed.
- [x] `uv run pytest -m "not ollama"` — 2102 passed, 7 skipped.
- [x] `uv run mypy src` — 2 pre-existing errors (`manager.py:459, :862`, tracked in #164); no new errors from F1.

New tests:

- `TestExplicitContextQueryKwarg` in `test_context_extractor.py` — 7 priority-order assertions (explicit wins over legacy; empty/whitespace falls through; template still wins; whitespace stripped).
- `TestExplicitContextQueryKwarg` in `test_surfacing_engine.py` — 3 behavioral assertions on `adapter.search` receiving the explicit query.
- `TestContextQueryPlumbing` in `test_proxy_manager_pipeline.py` — 4 assertions covering both helpers + cache-hit extraction.

### DIFF
--- a/src/memtomem_stm/proxy/manager.py
+++ b/src/memtomem_stm/proxy/manager.py
@@ -666,6 +666,7 @@ class ProxyManager:
         text: str,
         *,
         trace_id: str | None = None,
+        context_query: str | None = None,
     ) -> str:
         """Apply proactive memory surfacing if eligible."""
         if self._surfacing_engine is None:
@@ -677,6 +678,7 @@ class ProxyManager:
                 arguments=arguments,
                 response_text=text,
                 trace_id=trace_id,
+                context_query=context_query,
             )
         except Exception:
             logger.warning(
@@ -695,6 +697,7 @@ class ProxyManager:
         text: str,
         *,
         trace_id: str | None = None,
+        context_query: str | None = None,
     ) -> tuple[str, bool | None, str | None]:
         """Surface on a progressive first-chunk when the formatter mode keeps
         the ``PROGRESSIVE_FOOTER_TOKEN`` concat invariant intact.
@@ -731,6 +734,7 @@ class ProxyManager:
                 arguments=arguments,
                 response_text=text,
                 trace_id=trace_id,
+                context_query=context_query,
             )
             return surfaced, True, None
         except Exception as exc:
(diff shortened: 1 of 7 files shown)


pandas-studio (approved):
LGTM

PR #325, opened 2026-05-16
feat(surfacing): expose result_content_max_chars + preview_max_chars (F3)

memtomem:
## Summary

Two truncation limits silently capped how much LTM content reached the agent's context, with no operator knob:

- **Parser** (`mcp_client.py:107, 115` Compact; `:157, 164` Structured): each LTM result content capped at 500 chars before reaching the formatter.
- **Formatter** (`formatter.py:48, 53`): rendered preview capped at 300 chars before reaching the agent.

This PR exposes both as `SurfacingConfig` fields, defaults unchanged:

```python
result_content_max_chars: int = Field(default=500, gt=0)
preview_max_chars:        int = Field(default=300, gt=0)
```

Out-of-the-box behavior is **identical** to before — F3 is purely additive.

## Parse-time, not constructor-time

The parser is created once in `McpClientSearchAdapter.__init__` AND replaced with a bare `CompactResultParser()` during version-negotiation fallback. A constructor-stored limit would either become startup-only (breaking hot-reload) or get lost across the fallback path. F3 threads the limit through the call signature instead:

```python
def parse(
    self,
    text: str,
    *,
    max_content_chars: int = 500,
) -> tuple[list[RemoteSearchResult], list[str]]:
```

The adapter passes `self._config.result_content_max_chars` at parse time, so hot-reloads propagate automatically via the existing live-config pattern. The module-level `_compact_parser` singleton (used by a backwards-compat static helper) keeps working unchanged — it doesn't pass the kwarg, so it falls back to the 500 default.

For the formatter side, `SurfacingFormatter` already holds `self._config`, so the change is a one-line slice: `chunk.content[:300]` → `chunk.content[: self._config.preview_max_chars]`. The 150-char window slices (`window_before` / `window_after`) are intentionally left alone — they're context, not the primary preview, and were not flagged.

## Docs

Added two rows to the Surfacing Controls table in `docs/surfacing.md`, alongside `context_window_size`.

## Test plan

- [x] `uv run pytest -m "not ollama" tests/test_formatter.py tests/test_core_format_contract.py tests/test_surfacing_engine.py tests/test_remote_ltm_integration.py` — 136 passed (8 new in F3).
- [x] `uv run pytest -m "not ollama"` — 2117 passed, 7 skipped.

New tests:

- `TestPreviewMaxCharsKnob` in `test_formatter.py` — 3 tests: default 300-cap preserved, lower truncates, higher allows longer.
- `TestParserMaxContentChars` in `test_core_format_contract.py` — 5 tests: both parsers honor the kwarg, both default to 500 (regression guard against silent drift), short content unaffected.

## Out of scope

- **F2** (`min_score` default change) — deferred to an experiment-driven follow-up PR per the original adaptive-waffle plan.
- The 150-char window snippet slices in `formatter.py:47, 50`.

## Status of the surfacing roadmap

| Item | Status |
|---|---|
| F1 — `_context_query` plumbing | ✅ #324 |
| F2 — `min_score` default | deferred (experiment-driven) |
| **F3 — truncation knobs** | **this PR** |
| F4 — compression.md docs | ✅ #323 |

### DIFF
--- a/docs/surfacing.md
+++ b/docs/surfacing.md
@@ -85,6 +85,8 @@ The injection mode is configurable: `prepend` (default), `append`, or `section`.
 | `include_session_context` | `true` | Include working memory (scratch) items |
 | `dedup_ttl_seconds` | `604800` (7d) | Cross-session dedup window; `0` to disable |
 | `context_window_size` | `0` | Expand ±N adjacent chunks around search hits; `0` to disable |
+| `result_content_max_chars` | `500` | Max chars retained per LTM result before the formatter sees it |
+| `preview_max_chars` | `300` | Max chars per result preview in the injected memory block |
 | `consumer_model` | `""` | Model name for auto-scaling `max_results` and `max_injection_chars` |
 | `feedback_db_path` | `~/.memtomem/stm_feedback.db` | SQLite store for events, feedback, and cross-session dedup |
 | `cache_ttl_seconds` | `60.0` | Internal surfacing result cache TTL |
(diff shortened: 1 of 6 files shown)


pandas-studio (approved):
LGTM

PR #326, opened 2026-05-16
fix(surfacing/formatter): apply preview_max_chars to assembled preview (F3 follow-up)

memtomem:
## Summary

Follow-up to #325. A P2 review comment landed on the original F3 PR after the squash-merge: when ``context_window_size > 0``, the formatter joins ±150-char window snippets around the matched chunk, so a deployment setting ``preview_max_chars=50`` could still render >350 chars per result — contradicting the knob's documented "max chars per result preview".

The fix commit (then `5d62e4f`, now `f9d5867` after rebase onto main) was pushed to the F3 branch but missed the squash window. This PR re-applies it on top of merged main.

## Change

`src/memtomem_stm/surfacing/formatter.py:52` — slice the joined preview:

```python
preview = " | ".join(parts)[:preview_cap]
```

This is option 1 from the review (enforce after assembling), chosen over option 2 (budget context snippets within the cap) because a hard ceiling matches the documented "max chars per result preview" most directly and degrades predictably for operators picking small values.

## Test

New regression test in `tests/test_formatter.py`: `TestPreviewMaxCharsKnob::test_cap_bounds_assembled_preview_with_context_windows`.

- Constructs a result with `window_before` and `window_after` of 400 chars each
- Sets `preview_max_chars=50`
- Asserts the rendered preview line is ≤ 50 chars

Reverting the one-line fix on this branch causes the test to fail with "got len=358".

## Verification

- [x] `uv run pytest -m "not ollama" tests/test_formatter.py` — 13 passed (+1 new).

### DIFF
--- a/src/memtomem_stm/surfacing/formatter.py
+++ b/src/memtomem_stm/surfacing/formatter.py
@@ -49,7 +49,10 @@ class SurfacingFormatter:
                 parts.append(chunk.content[:preview_cap].replace("\n", " "))
                 if ctx.window_after:
                     parts.append(ctx.window_after[0].content[:150].replace("\n", " ") + "...")
-                preview = " | ".join(parts)
+                # ``preview_max_chars`` is a hard ceiling on the per-result
+                # preview — without this, joining ±150-char window snippets
+                # would push a single result past the documented cap.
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

Issue #289, opened 2026-05-02
proxy/compression: LLM compression sends sensitive content (privacy_patterns not wired)


memtomem:
Found during the 2026-05-02 implementation review (\`docs/reports/memtomem-stm-implementation-review-2026-05-02.md\`, finding #2).

## Symptom

\`LLMCompressor.compress()\` accepts \`privacy_patterns: list[str] | None\` and short-circuits to \`TruncateCompressor\` when sensitive content is detected:

- \`src/memtomem_stm/proxy/compression.py:1166-1181\`
- \`src/memtomem_stm/proxy/privacy.py:41\` (\`contains_sensitive_content\`)

But \`ProxyManager._apply_compression()\` calls it without passing patterns:

- \`src/memtomem_stm/proxy/manager.py:637\` — \`compressor.compress(text, max_chars=max_chars)\`

When an upstream tool's response contains an API key, password, or JWT and \`compression\` resolves to \`llm_summary\`, that content is sent to the configured LLM provider (OpenAI / Anthropic / custom \`base_url\`) verbatim.

## Why this matters

The privacy module exists today specifically as a routing signal for compression (per memory \`reference_proxy_privacy_scope.md\`). It was added to prevent exactly this leak, but the wire-in step was missed. The non-standard \`base_url\` warning at \`compression.py:1154-1164\` already acknowledges the credential-exfiltration risk; the corresponding scan is one call site away.

## Recommendation

1. Add \`privacy_scan_enabled: bool = True\` and \`privacy_patterns: list[str] = Field(default_factory=lambda: DEFAULT_PRIVACY_PATTERNS)\` to \`LLMCompressorConfig\`.
2. In \`ProxyManager._apply_compression()\` (LLM_SUMMARY branch), pass \`privacy_patterns=llm_cfg.privacy_patterns if llm_cfg.privacy_scan_enabled else None\`.
3. Regression test: an upstream response containing an API key triggers \`compressor.last_fallback == \"privacy\"\` instead of an outbound call.

Default-on so an operator who flips \`compression: llm_summary\` doesn't have to remember a second knob to be safe.

## Scope

Small (~10 lines plus a test). Suitable for a single focused PR.

PR #327, opened 2026-05-16
fix(proxy/privacy): scan full payload instead of first 10K chars

memtomem:
## Summary

`contains_sensitive_content()` (used by `LLMCompressor`'s privacy guard before sending payloads to the external LLM provider) was sampling only the first 10K chars. Any sensitive token past that offset — JWT, AWS key, private key, password assignment — slipped past the scan and got forwarded to the provider.

Drops the truncation and scans the full text. The privacy scan runs **before** the LLM call, so the alternative trade-off is "send N KB of potentially-sensitive content to a third party for a small CPU savings" — not worth it.

## Behavior change

- Privacy scan now flags sensitive content anywhere in the payload, not just the first 10K chars.
- No new config knob — this is a correctness fix on an existing default-on guard (#289).
- Pattern compilation is unchanged; the regex set is bounded (≤ ~10 patterns per call site), so full-text scan over a few hundred KB is dominated by Python regex engine cost, which is linear and remains fast in practice.

## Test plan

- [x] `uv run pytest tests/test_privacy.py tests/test_compression.py -q` — 85 passed
- [x] `uv run pytest -m "not ollama" -q` — 2126 passed (+2 new tests for the past-10K case)

New tests pin both layers:
- `test_privacy.py`: unit test asserting `contains_sensitive_content` detects a sensitive token past the 10K mark
- `test_compression.py`: `LLMCompressor` fallback test asserting `_call_api` is never reached when the sensitive token sits past 10K

### DIFF
--- a/src/memtomem_stm/proxy/privacy.py
+++ b/src/memtomem_stm/proxy/privacy.py
@@ -39,13 +39,9 @@ def _compile_patterns(patterns: tuple[str, ...]) -> list[re.Pattern[str]]:
 
 
 def contains_sensitive_content(text: str, patterns: list[str] | None = None) -> bool:
-    """Check if text contains any sensitive patterns.
-
-    Scans only the first 10K chars for performance.
-    """
+    """Check if text contains any sensitive patterns."""
     effective = patterns if patterns else DEFAULT_PATTERNS
     if not effective:
         return False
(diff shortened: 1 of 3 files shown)


pandas-studio (approved):
LGTM

PR #329, opened 2026-05-17
feat(surfacing): bump min_score default 0.02 → 0.03 (F2)

memtomem:
## Summary

- F2 of the surfacing roadmap (deferred from #324/#325/#326 pending benchmark data). Re-runs the s11 min_score sweep harness from #328, reads the curve, and bumps `SurfacingConfig.min_score` from `0.02` → `0.03` — the F1 peak on the s11 fixture.
- AutoTuner clamps untouched; tightening the upper clamp at 0.05 (the cliff in the curve) is a separate decision that should wait for feedback-distribution data.
- Docs (`docs/surfacing.md` table, `docs/configuration.md` env var example) kept in sync.

## The curve

From `sweep_s11_min_score.json` (byte-identical on main and this branch — deterministic):

| threshold | precision | recall |   F1   | returned | tp / fp / fn |
|-----------|-----------|--------|--------|----------|--------------|
|   0.005   |   0.571   |  1.000 | 0.7273 |    7     |  4 / 3 / 0   |
|   0.010   |   0.667   |  1.000 | 0.8000 |    6     |  4 / 2 / 0   |
|   0.020   |   0.750   |  0.750 | 0.7500 |    4     |  3 / 1 / 1   | ← prior default |
| **0.030** | **1.000** |**0.750**|**0.8571**|  **3** |  3 / 0 / 1   | ← **new default** |
|   0.050   |   1.000   |  0.250 | 0.4000 |    1     |  1 / 0 / 3   |
|   0.070   |   0.000   |  0.000 | 0.0000 |    0     |  0 / 0 / 4   |

The current default sits in a dip between the recall-leaning (0.010) and precision-leaning (0.030) regimes. Bumping to 0.030 drops the lone false positive without losing any positives — recall holds at 0.75, precision goes from 0.75 → 1.0, F1 +14%.

## Why not also tighten AutoTuner clamps?

AutoTuner's upper clamp is `0.05` (`src/memtomem_stm/surfacing/feedback.py:116`). The curve shows F1 collapsing from 0.857 at 0.030 to 0.400 at 0.050 — that's a cliff, and the clamp sits on the wrong side of it. Worth revisiting, but:

- It's a different policy lever (per-tool adaptation, not the global default).
- It needs feedback-distribution data to calibrate properly (when would the auto-tuner actually push a tool's threshold up to 0.04 vs 0.05?).
- Single-purpose PR is easier to revert if production tells a different story.

Leaving it for a separate F2-follow-up once `surfacing_events` / `surfacing_feedback` data exists in production.

## Caveats

- One synthetic fixture (`s11`). The local `stm_feedback.db` still has no `surfacing_events` / `surfacing_feedback` tables, so feedback-distribution validation isn't available yet — the deferred-memory rationale is half-satisfied.
- Recoverable via `MEMTOMEM_STM_SURFACING__MIN_SCORE` for operators who preferred the recall-leaning behavior.
- No test changes: every test that pins a threshold sets `min_score` explicitly; the s11 sweep is threshold-driven and produces an identical curve on this branch.

## Test plan

- [x] `uv run ruff check src && uv run ruff format --check src` — passes.
- [x] `uv run pytest -m "not ollama and not bench_qa_sweep"` — 2126 passed, 7 skipped.
- [x] `pytest test_s11_min_score_sweep -m bench_qa_sweep` — passes; curve byte-identical to main (only the default changed, not the harness).
- [ ] CI: 9 jobs incl. Windows matrix.


pandas-studio (approved):
LGTM

PR #331, opened 2026-05-17
feat(surfacing): per-tool feedback + auto-tune readiness in stm_surfacing_stats

memtomem:
## Summary

Productionization pass on the surfacing feedback path. The `surfacing_events` / `surfacing_feedback` tables and the `stm_surfacing_stats` / `stm_surfacing_feedback` MCP tools already exist; this PR adds the observability needed to actually act on production data, plus a regression test pinning the real on-disk DB layout.

## What changed

- **`FeedbackStore.get_stats()`** per-tool rows now include `feedback_count` and `not_relevant_count` from a single JOIN against `surfacing_feedback`. The `since=` filter applies to feedback via its parent event's `created_at`, so the new counts honor the same window as `events_total`. No new index — the existing `idx_feedback_surfacing` covers the JOIN.
- **`SurfacingEngine.get_min_score_snapshot()`** exposes config default, AutoTune on/off, `auto_tune_min_samples` threshold, and per-tool adjustments AutoTuner has made this process. `AutoTuner.adjustments` gains a read-only property returning a snapshot copy so callers cannot mutate the tuner's `_adjustments` dict.
- **`stm_surfacing_stats` formatting**:
  - New `Min score:` block — `default 0.030 (auto-tune on, min 20 samples)` + `Per-tool adjustments:` sublist when AutoTuner has moved any tool off the default.
  - Per-tool rows annotate auto-tune state: `auto-tuned`, `auto-tune ready` (≥ min_samples but inside [0.2, 0.6] no-op band), or `need N more for auto-tune`; render negative ratio as `feedback N (negative R%)` when feedback exists.
- **`TestFeedbackStoreCoexistence`** — regression test: opens a shared DB previously initialized by `CompressionFeedbackStore` (matching real `~/.memtomem/stm_feedback.db` layout: `compression_feedback` + `progressive_reads` only), then runs `FeedbackStore.initialize()` and verifies surfacing tables get added without touching existing rows. Plus an idempotency test on existing surfacing schema.

## What did NOT change

- AutoTuner clamps `[0.005, 0.05]` and the 0.6 / 0.2 ratio thresholds. Tightening the 0.05 upper clamp is still the natural next step, but its right input is production `surfacing_feedback` rows — which this PR makes *visible*, not the synthetic bench.
- The `_SCHEMA` block. Tables are unchanged; the productionization concern is observability, not schema evolution.

## Test plan

- [x] `uv run ruff check src && uv run ruff format --check src`
- [x] `uv run pytest -m \"not ollama\"` — 2132 passed (+5 new), 7 skipped

### DIFF
--- a/src/memtomem_stm/server.py
+++ b/src/memtomem_stm/server.py
@@ -700,11 +700,17 @@ async def stm_surfacing_stats(
                 if tool_name in overrides:
                     detail_parts.append(f"pinned {overrides[tool_name]:.3f}")
                 elif auto_tune_active and min_samples_required > 0:
-                    if fb >= min_samples_required:
-                        if tool_name in adjusted_scores:
-                            detail_parts.append("auto-tuned")
-                        else:
-                            detail_parts.append("auto-tune ready")
+                    # Check ``adjusted_scores`` first: AutoTuner has a documented
+                    # cold-start fallback (feedback.py::maybe_adjust) that tunes a
+                    # tool from the global feedback pool even when the tool's own
+                    # feedback count is below ``min_samples``. So an "auto-tuned"
+                    # row can have fb < min_samples — gating on fb first would
+                    # contradict both the per-tool adjustment block and the
+                    # effective threshold the engine actually used.
+                    if tool_name in adjusted_scores:
+                        detail_parts.append("auto-tuned")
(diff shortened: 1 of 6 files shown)


pandas-studio (approved):
LGTM

PR #332, opened 2026-05-17
feat(surfacing): persist AutoTuner min_score adjustments across restarts

memtomem:
## Summary
- AutoTuner's per-tool `_adjustments` lived in-memory only — every server restart silently threw away tunings the tuner had earned from feedback. With the readiness stats shipped in #331, this loss is now visible to operators, which makes it worth fixing.
- Adds an `auto_tune_adjustments(tool PK, min_score, updated_at)` table to the existing feedback DB. `AutoTuner.__init__` rehydrates `_adjustments` via `FeedbackStore.load_adjustments()`; every threshold change calls `FeedbackStore.save_adjustment(tool, new_score)`. No new file, no new connection.
- `stm_surfacing_stats.adjusted_scores` now reflects the tuner's accumulated view across restarts rather than only the current process.

## Test plan
- [x] `uv run pytest tests/test_feedback.py` — 31 passed, including two new tests:
  - `test_adjustment_persists_across_restart` — full round-trip via two `FeedbackStore` lifetimes on one DB file
  - `test_save_and_load_adjustments_roundtrip` — upsert overwrite
- [x] `uv run pytest -m "not ollama"` — 2142 passed, 7 skipped
- [x] `uv run ruff check src tests` + `ruff format --check`
- [ ] Manual: bring the server up, induce an adjustment (or hand-write a row via sqlite3), restart, confirm `stm_surfacing_stats` still shows the tuning under `adjusted_scores`

## Notes for the reviewer
- The schema uses `CREATE TABLE IF NOT EXISTS`, so existing feedback DBs upgrade in place — no migration step needed.
- `AutoTuner` always saves the absolute min_score (not a delta). If an operator later changes `SurfacingConfig.min_score`, persisted tool-specific tunings still take precedence — which preserves the tuner's prior decision, consistent with the pre-existing per-process behavior. Clearing rows in `auto_tune_adjustments` is the explicit reset path.

### DIFF
--- a/src/memtomem_stm/surfacing/feedback.py
+++ b/src/memtomem_stm/surfacing/feedback.py
@@ -82,7 +82,9 @@ class AutoTuner:
     def __init__(self, config: SurfacingConfig, store: FeedbackStore) -> None:
         self._config = config
         self._store = store
-        self._adjustments: dict[str, float] = {}
+        # Resume tunings persisted from a previous process — without this
+        # every restart silently throws away the AutoTuner's view.
+        self._adjustments: dict[str, float] = dict(store.load_adjustments())
 
     def maybe_adjust(self, tool: str) -> float | None:
         """Check feedback ratio and adjust min_score for a tool.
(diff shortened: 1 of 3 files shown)


pandas-studio (approved):
LGTM

PR #333, opened 2026-05-17
feat(surfacing): expose bootstrap health

memtomem:
## Summary
- add a read-only surfacing feedback DB schema inspector
- show surfacing bootstrap readiness in `mms health` text and JSON output
- log feedback-store initialization and surfacing path wiring at startup

## Tests
- `uv run pytest tests/test_feedback.py tests/cli/test_proxy_cli.py::TestHealth tests/test_server_tools.py::TestLifespan`
- `uv run ruff check src/memtomem_stm/surfacing/feedback_store.py src/memtomem_stm/surfacing/feedback.py src/memtomem_stm/server.py src/memtomem_stm/cli/proxy.py tests/test_feedback.py tests/cli/test_proxy_cli.py`
- `uv run mypy src/memtomem_stm/surfacing/feedback_store.py src/memtomem_stm/surfacing/feedback.py src/memtomem_stm/server.py src/memtomem_stm/cli/proxy.py`
- `git diff --check`

PR #335, opened 2026-05-17
test(surfacing): pin auto_tune_adjustments in bootstrap readiness

memtomem:
## Summary
- Regression test pinning the contract that `inspect_feedback_db()`
  treats `auto_tune_adjustments` as a required table.
- Closes the P2 reviewer comment on #333: an older DB with only the
  pre-#332 three-table schema must report `initialized=false` so
  `AutoTuner.load_adjustments` / `save_adjustment` can't follow a
  false-positive readiness signal.

## Why test-only

The hardening commit folded into the #333 squash already derives
`_REQUIRED_TABLES` from `_SCHEMA` via regex, so live behavior is
correct (`_REQUIRED_TABLES` resolves to all four tables at import
time). What was missing is a test that pins the contract — without
it, a refactor that re-introduces a hand-maintained tuple would
silently regress the same bug the reviewer flagged.

## Test plan
- [ ] `uv run pytest tests/test_feedback.py -q` passes (new test
      `test_inspect_feedback_db_pre_auto_tune_schema` exercises a
      legacy three-table DB).
- [ ] `uv run ruff check src && uv run ruff format --check src`.

### DIFF
--- a/tests/test_feedback.py
+++ b/tests/test_feedback.py
@@ -51,6 +51,30 @@ class TestFeedbackStore:
         assert status["initialized"] is True
         assert status["missing_tables"] == []
 
+    def test_inspect_feedback_db_pre_auto_tune_schema(self, tmp_path: Path):
+        # Regression: pre-#332 DBs with only the three original tables must
+        # report not-initialized so AutoTuner SQL errors don't follow a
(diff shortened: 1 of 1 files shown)


pandas-studio (approved):
LGTM

PR #336, opened 2026-05-17
feat(surfacing): surface bootstrap readiness via stm_proxy_health

### LINKED ISSUE #334: feat(surfacing): expose bootstrap readiness on MCP surfaces (follow-up to #333)
## Background

#333 added bootstrap readiness for the surfacing feedback DB — a read-only
schema inspector (`inspect_feedback_db`) wired into the `mms health` CLI
(text + JSON). That PR deliberately scoped to one surface; the same signal
is useful on the MCP-tool side, which agents (not operators) hit.

## The gap

- `stm_proxy_health` (`server.py:555`) reports upstream connectivity +
  surfacing circuit-breaker state but says nothing about whether the
  feedback DB exists or its tables are initialized. An agent seeing a
  "healthy" proxy can still be running with surfacing silently degraded.
- `stm_surfacing_stats` (`server.py:620`) reads from `stm_feedback.db`;
  when the DB is missing/uninitialized it currently fails or returns an
  empty view with no readiness hint.

## Proposed change

1. Extend `stm_proxy_health` to include the same `surfacing` block that
   `mms health --json` now emits (reuse `_surfacing_bootstrap_status()` —
   move it out of `cli/proxy.py` into `surfacing/feedback_store.py` or a
   sibling so `server.py` doesn't import from the CLI module).
2. Make `stm_surfacing_stats` short-circuit with a clear "feedback DB not
   initialized" message when `inspect_feedback_db()` reports missing
   tables, instead of bubbling a SQL error.

## Also deferred (smaller, list-only)

- AutoTuner persistence DB (#332) has no equivalent inspector.
- Inspector reports "missing tables" but doesn't name the bootstrap path
  (e.g. "first surfacing call will create them") — a `next_action` hint
  would close the loop for operators.

## Files

- `src/memtomem_stm/cli/proxy.py` — `_surfacing_bootstrap_status` (move)
- `src/memtomem_stm/surfacing/feedback_store.py` — host the helper
- `src/memtomem_stm/server.py:555,620` — wire into MCP tools

## Tests

- `tests/test_server_tools.py` — `stm_proxy_health` includes surfacing
  block; missing DB renders as "not initialized" not as an error.
- `tests/test_server_tools.py` — `stm_surfacing_stats` returns readiness
  hint when DB is absent.

memtomem:
Closes #334 (first half — MCP-tool surfaces).

## Summary

- Extend `stm_proxy_health` to include a Surfacing Bootstrap section
  matching what `mms health` (CLI) emits since #333. Agents inspecting
  the proxy via MCP can now tell whether feedback tracking is wired,
  where the DB lives, and whether its tables are ready — without
  shelling out to the CLI.
- Reuse `FeedbackTracker.bootstrap_status()` so the inspection logic
  has one home in `surfacing/feedback_store.py` (also exercises the
  `auto_tune_adjustments` regression coverage pinned in #335).

## Out of scope (left for a follow-up)

- The original #334 also called for `stm_surfacing_stats` to short-
  circuit with a "DB not initialized" hint. Re-reading the lifespan
  path (`server.py:148-156`), `FeedbackTracker.__init__` failures are
  already caught and set `feedback_tracker=None`, so the existing
  "not enabled" branch already covers the bad-DB case — there is no
  raw SQL error path to short-circuit. Punting that scope rather than
  adding a message tweak that doesn't fix a real failure.

## Test plan

- [ ] `uv run pytest tests/test_server_tools.py::TestHealth -q` — six
      cases covering: no servers, with servers, tracker ready, tracker
      missing tables, tracker None with feedback_enabled
      (\"runtime init failed\"), surfacing disabled (no block).
- [ ] `uv run pytest -m "not ollama"` — full suite.
- [ ] `uv run ruff check src && uv run ruff format --check src`.
- [ ] `uv run mypy src/memtomem_stm/server.py` — advisory, passes.

### DIFF
--- a/src/memtomem_stm/server.py
+++ b/src/memtomem_stm/server.py
@@ -558,9 +558,12 @@ async def stm_proxy_health(
     app = _get_ctx(ctx)
     pm = app.proxy_manager
 
+    bootstrap_lines = _surfacing_bootstrap_lines(app)
+
     health = pm.get_upstream_health()
     if not health:
-        return "No upstream servers configured."
+        head = "No upstream servers configured."
+        return "\n".join([head, *bootstrap_lines]) if bootstrap_lines else head
 
     lines = ["Upstream Server Health", "====================="]
     for name, info in health.items():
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #337, opened 2026-05-17
feat(surfacing): type LTM adapter outcomes for distinct skip labels

### LINKED ISSUE #295: LTM adapter outcome typing: distinguish empty / transport_error / parse_error
## Background

`McpClientSearchAdapter.search()` returns `([], [])` for at least five different conditions:

- `src/memtomem_stm/surfacing/mcp_client.py:292` — no session (adapter never connected, or `stop()` already called)
- `src/memtomem_stm/surfacing/mcp_client.py:322` — `mem_search` failed after a transport-error reconnect attempt
- `src/memtomem_stm/surfacing/mcp_client.py:325` — generic `Exception` from `mem_search` call
- `src/memtomem_stm/surfacing/mcp_client.py:332` — call succeeded but response carried no text content
- `src/memtomem_stm/surfacing/mcp_client.py:335` — call succeeded, parser returned empty list

`SurfacingEngine` then collapses these into a single observability label:

- `src/memtomem_stm/surfacing/engine.py:498-501` — records `no_results_dedup` or `no_results_score` with no way to tell "LTM unavailable" from "LTM returned zero results" from "parser hit an edge case".

## Concrete operational gap

An operator looking at `stm_feedback.db` skip counters cannot answer:

- Is LTM actually unreachable?
- Did the namespace simply have nothing relevant?
- Is the result parser broken on a recent core release?

All three look the same.

## Proposal

Return a structured outcome from `search()` (status enum: `ok` / `no_session` / `transport_error` / `call_error` / `empty_content` / `empty_results`) **or** raise typed exceptions for the failure cases and keep `([], [])` for the genuine empty path. `SurfacingEngine` then records distinct skip labels (`ltm_unavailable`, `ltm_call_failed`, `ltm_parse_empty`, `no_results_score`).

## Why doesn't existing reconnect cover this

The `_TRANSPORT_ERRORS` reconnect at `mcp_client.py:311-322` already retries once. The gap is *post-retry*: when retry also fails, the operator silently gets the same label as a healthy empty namespace.

## Test plan

- Parametrised test that drives each failure mode through `search()` and asserts the distinct outcome label landing in `stm_feedback.db`.
- Mutation test confirming a regression in the dispatch (e.g. mapping `transport_error` to `no_results_score`) fails the assertion.

## Scope

~50–100 LOC across `mcp_client.py`, `engine.py`, observability skip labels, plus tests. Single PR.

## Out of scope

- Per-call timeout placement (tracked separately).
- Non-stdio transport (tracked separately).

memtomem:
Closes #295.

## Summary
- `McpClientSearchAdapter.search` now returns `(results, hints, outcome)` with a `SearchOutcome` `Literal` (`ok` / `no_session` / `transport_error` / `call_error` / `empty_content` / `empty_results`) so the engine can tell five previously-indistinguishable failure modes apart.
- `SurfacingEngine._do_surface_miss` reads the outcome and records three new skip labels — `ltm_unavailable` (no session / transport error after reconnect retry), `ltm_call_failed` (non-transport exception), `ltm_parse_empty` (call OK but no text content). `ok` and `empty_results` fall through to the existing `no_results_score` path so `min_score` tuning signal is preserved.
- Failure outcomes early-return before `cache.set` so an `ltm_unavailable` does not poison the cache with an empty entry — pinned by `test_failure_outcomes_do_not_populate_cache`.

## Test plan
- [x] `uv run pytest -m "not ollama"` — 2161 passed
- [x] `uv run mypy src/memtomem_stm/surfacing/{mcp_client,engine,observability}.py` — 11 pre-existing `union-attr` errors, identical to `main` (advisory per CLAUDE.md)
- [x] Engine-level dispatch: `TestSurfacingLtmOutcomeDispatch` covers each new label + the empty_results regression assertion + the cache-non-poisoning invariant
- [x] Adapter-level outcomes: `TestReconnectRetry*`, `TestNonTransportErrorsDoNotReconnect`, `TestNoneContentDefense`, `TestOuterCancellationLazyReconnect`, `TestTextNoneTolerance` all assert the right outcome string for their scenario

### DIFF
--- a/src/memtomem_stm/surfacing/engine.py
+++ b/src/memtomem_stm/surfacing/engine.py
@@ -492,13 +492,31 @@ class SurfacingEngine:
         search_kwargs: dict[str, Any] = {}
         if ctx_win:
             search_kwargs["context_window"] = ctx_win
-        results, hints = await self._mcp_adapter.search(
+        results, hints, outcome = await self._mcp_adapter.search(
             query=query,
             top_k=max_results * 2,
             namespace=namespace,
             trace_id=trace_id,
             **search_kwargs,
         )
+
+        # #295: branch on the adapter outcome before doing any further work
+        # so the operator-facing skip label distinguishes "LTM unavailable"
+        # (``no_session``/``transport_error``) from "LTM call raised"
+        # (``call_error``) from "core returned no text" (``empty_content``).
+        # ``ok`` and ``empty_results`` both fall through to the existing
+        # min_score / dedup / ``no_results_score`` path so that operators
+        # tuning min_score keep seeing the same signal for the genuine
+        # empty-namespace case.
+        if outcome in ("no_session", "transport_error"):
+            self._observability.record_skip(tool, "ltm_unavailable")
+            return response_text
+        if outcome == "call_error":
+            self._observability.record_skip(tool, "ltm_call_failed")
+            return response_text
+        if outcome == "empty_content":
+            self._observability.record_skip(tool, "ltm_parse_empty")
+            return response_text
 
         # Parent trust-UX hints (parent PR #231): log at INFO even when results
         # are empty or get filtered out below, since an operator may want to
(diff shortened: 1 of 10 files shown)


pandas-studio (approved):
LGTM

PR #338, opened 2026-05-17
fix(surfacing): defer LTM MCP client start to first RPC

memtomem:
## Summary

- Move `McpClientSearchAdapter.start()` out of `app_lifespan` and into `_heal_if_needed` so the proxy's MCP `initialize()` returns immediately instead of blocking on the LTM subprocess handshake.
- Add an `asyncio.Lock` to serialize concurrent first-callers and a sticky `_start_attempted` flag to preserve the prior single-shot lifespan semantics when LTM is permanently unreachable.
- Add three `TestLazyStart` tests pinning the new contract.

## Why

Eagerly awaiting `mcp_adapter.start()` in `app_lifespan` blocks the proxy's own MCP `initialize()` while the LTM subprocess spawns and runs initialize + `mem_do(version)` round trips. On hosts with short startup timeouts (e.g. codex's 60s `startup_timeout_sec`), this regularly exceeded the budget. The host then respawned the proxy mid-init, leaving two parallel LTM children — exact symptom users reported as `Starting MCP servers (2/3): memtomem-stm` hanging indefinitely with duplicate `mms` / `memtomem-server` processes.

Diagnosis details:
- `mms` direct-launch initialize completed in ~4s
- Inside codex, same `mms` exceeded the 60s budget and codex respawned, producing two MCP-server groups under one `codex` PID — observed live, both groups sleeping waiting for handshake
- The expensive step in `start()` is the chain `stdio_client → ClientSession → session.initialize() → _negotiate_format()`, none of which is needed before the first surfacing RPC

## Behavior delta

| scenario | before | after |
|---|---|---|
| Proxy `initialize()` latency | LTM full handshake (often >60s under load) | sub-second |
| First surfacing RPC | always ready | first caller triggers `start()`, others lock-wait |
| Permanently unreachable LTM | `surfacing_engine = None`, single attempt | `_start_attempted` sticky, single attempt — equivalent |
| Mid-RPC cancellation reconnect (#290) | unchanged | unchanged |

The connection is established lazily; from the engine's perspective the only observable change is that the first cycle pays a one-time start cost instead of paying it during proxy boot.

## Test plan

- [x] `uv run pytest -m "not ollama"` — 2164 passed, 7 skipped (no regressions)
- [x] `uv run ruff check && uv run ruff format --check` — clean
- [x] New `TestLazyStart` covers:
  - first `search()` triggers `start()` and returns real results (not the old silent `no_session`)
  - failed `start()` flips `_start_attempted` and the next call short-circuits (no thundering herd of LTM respawns)
  - three concurrent `search()` callers serialize through the lock — `start()` runs exactly once
- [ ] Manual verification with codex: `Starting MCP servers (2/3): memtomem-stm` no longer hangs; only one LTM child group per proxy instance

### DIFF
--- a/src/memtomem_stm/server.py
+++ b/src/memtomem_stm/server.py
@@ -127,14 +127,19 @@ async def app_lifespan(server: FastMCP) -> AsyncIterator[STMContext]:
             # MCP client adapter. The adapter spawns (or connects to) a
             # memtomem MCP server using
             # config.surfacing.ltm_mcp_command / ltm_mcp_args.
+            # The adapter's MCP connection is deferred to the first
+            # surfacing RPC (see ``McpClientSearchAdapter._heal_if_needed``);
+            # eagerly awaiting ``start()`` here used to block the proxy's
+            # own MCP initialize handshake long enough for hosts (e.g.
+            # codex with a 60s startup_timeout) to time out and respawn
+            # the proxy, leaving two parallel LTM children.
             if config.surfacing.enabled:
                 try:
                     from memtomem_stm.surfacing.mcp_client import McpClientSearchAdapter
 
                     mcp_adapter = McpClientSearchAdapter(config.surfacing)
-                    await mcp_adapter.start()
                     logger.info(
-                        "Surfacing engine connected via MCP client to %s",
+                        "Surfacing engine MCP client configured for lazy start: %s",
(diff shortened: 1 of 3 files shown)


pandas-studio (approved):
LGTM

PR #339, opened 2026-05-17
docs(proxy): warn when auto_index/extraction enabled without index engine (#288)

### LINKED ISSUE #288: proxy: auto_index/extraction inert in normal server (no FileIndexer adapter)
Found during the 2026-05-02 implementation review (`docs/reports/memtomem-stm-implementation-review-2026-05-02.md`, finding #1).

## Symptom

`server.py` constructs `ProxyManager` without `index_engine=`:

- `src/memtomem_stm/server.py:187`

Both the auto_index stage (`src/memtomem_stm/proxy/manager.py:1651-1653`) and the extraction stage (`src/memtomem_stm/proxy/manager.py:1745-1747`) are gated on `self._index_engine is not None`. With no engine, enabling `auto_index.enabled=true` or `extraction.enabled=true` in `stm_proxy.json` is a no-op for runtime ingestion.

`ProxyManager.start()` already logs a warning when `auto_index.enabled=true` and the engine is missing (`src/memtomem_stm/proxy/manager.py:206-210`), but the documented pipeline (CLEAN → COMPRESS → SURFACE → INDEX in CLAUDE.md) still implies INDEX runs in normal operation.

## Tension with invariants

- CLAUDE.md: `STM talks to the LTM server only through the MCP protocol. Don't import memtomem from src/.`
- `FileIndexer` (in core memtomem) writes directly to LTM's filesystem. STM cannot reuse it without violating the invariant.
- Today there is no MCP-protocol-only `FileIndexer` adapter.

## Options

1. **Implement a remote `FileIndexer` adapter** that calls LTM tools (`mem_add` / a future `mem_index_file`) over MCP, and wire it into `ProxyManager` when `auto_index` or `extraction` is enabled. This makes the documented pipeline real.
2. **Mark INDEX/EXTRACT as unavailable in the standalone server path** in README/CLAUDE.md and strengthen the startup warning so operators don't enable a config that silently does nothing.

## Recommendation

Status: design — needs decision before implementation. Likely RFC-scope due to interaction with the read-only-discovery invariant (separate decision from \"auto_index\").

## Reopen / advance trigger

A user reports "I enabled `auto_index` and nothing was stored" or expresses intent to write back to LTM via the proxy at runtime.

memtomem:
## Summary

- Standalone `mms` constructs `ProxyManager` without an `index_engine`, so Stage 4 (auto_index / extraction) is silently skipped. Only the global `auto_index.enabled=true` case emitted a warning — per-upstream (`UpstreamServerConfig.auto_index`) and per-tool override (`ToolOverrideConfig.auto_index`) paths were inert with no operator signal, including the documented `github` example in `docs/configuration.md` that sets `auto_index: true` per-server.
- `ProxyManager.start()` now scans every enable-site (global, server, tool override) for both `auto_index` and `extraction`, then emits one warning per stage with a `(site1, site2, …)` location list. Operators can grep for `inert` and find exactly which config sites point at a missing index engine.
- README pipeline diagram, top-of-README narrative, "What STM doesn't do" section, CLAUDE.md pipeline invariant, and `docs/configuration.md` auto_index block are all updated so the docs no longer imply INDEX runs in the standalone server today.

Resolves #288 in its docs+warning scope. The remote `FileIndexer` MCP adapter remains the deferred RFC follow-up; the reopen trigger ("I enabled auto_index and nothing was stored") still applies for that future work.

## Test plan

- [x] `uv run pytest -m "not ollama"` — 2164 passed, 7 skipped (was 2162 + 2 new tests)
- [x] `uv run pytest tests/test_proxy_manager.py::TestAutoIndexStartupWarning -v` — 6 passed, including:
  - `test_warns_auto_index_enabled_but_no_engine` (tightened wording)
  - `test_warns_extraction_enabled_but_no_engine` (new)
  - `test_warns_per_server_auto_index_without_engine` (new, covers `UpstreamServerConfig.auto_index=True`)
  - `test_warns_per_tool_override_extraction_without_engine` (new, covers `ToolOverrideConfig(extraction=True)`)
- [x] `uv run mypy src` — pre-existing errors only (#164); no new errors at the edited block

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -30,9 +30,13 @@ The `ollama` marker auto-skips when Ollama isn't running; CI always uses
   `sys.stdin.isatty()`: TTY → help, non-TTY → `server.main` (the MCP stdio
   server). Don't diverge behavior between the three names — registering
   any of them as an MCP client's `command` must work identically.
-- **Pipeline order is CLEAN → COMPRESS → SURFACE → INDEX** — comments in
-  `src/memtomem_stm/proxy/` are the source of truth for the per-stage
-  contracts; full architecture write-up lives in the private
+- **Pipeline order is CLEAN → COMPRESS → SURFACE → (INDEX)** — INDEX
+  only runs when `ProxyManager` is constructed with an `index_engine`;
+  the standalone `mms` server does not wire one today (see #288), so
+  `auto_index` / `extraction` config is inert in that path and logs an
+  `inert` warning at startup. Comments in `src/memtomem_stm/proxy/`
+  remain the source of truth for the per-stage contracts; full
+  architecture write-up lives in the private
   `memtomem-docs/memtomem-stm/guides-archived/pipeline.md`.
 - **Line length 100**, target `py312` (`tool.ruff`, `tool.mypy`).
 - `.claude/` and `scripts/` are gitignored — don't commit anything under them,
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ It sits between your AI agent and its upstream MCP servers, compressing tool res
 flowchart TB
     Agent["Agent<br/>(Claude Code, Cursor, …)"]
     subgraph STM["memtomem-stm (STM)"]
-        Pipe["CLEAN → COMPRESS → SURFACE → INDEX"]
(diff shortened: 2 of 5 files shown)


pandas-studio (approved):
LGTM

PR #340, opened 2026-05-18
test(docs): pin Stage 4 inert claim to server.py construction (#299)

### LINKED ISSUE #299: docs/configuration.md auto-index Stage 4 description does not match server runtime behaviour
## Background

`docs/configuration.md:73` advertises:

```
# Auto-indexing (Stage 4 — save large responses to LTM)
export MEMTOMEM_STM_PROXY__AUTO_INDEX__ENABLED=false
export MEMTOMEM_STM_PROXY__AUTO_INDEX__MIN_CHARS=2000
export MEMTOMEM_STM_PROXY__AUTO_INDEX__MEMORY_DIR=~/.memtomem/proxy_index
export MEMTOMEM_STM_PROXY__AUTO_INDEX__NAMESPACE=proxy-{server}
```

In the bundled `mms` server, these env vars have no observable effect.

`src/memtomem_stm/server.py:187` constructs `ProxyManager(...)` without `index_engine`. The auto-index branches at `src/memtomem_stm/proxy/manager.py:1661-1663` are gated on `self._index_engine is not None`, so they never execute in normal `mms` runs. An operator who flips `AUTO_INDEX__ENABLED=true` sees zero behaviour change and zero error message.

The `AUTO_INDEX__*` knobs do work when `ProxyManager(..., index_engine=indexer)` is constructed by a library caller (used in tests, e.g. `tests/test_auto_index_background.py:77`).

## Status: blocked / awaiting upstream design decision

The right wording depends on whether STM will eventually own a server-wired write path to LTM, or whether auto-index should be re-positioned as a library-only hook. That decision is being scoped separately and is not yet ready.

This issue tracks the docs alignment that follows from the decision; it should not be implemented in isolation, because the right prose is materially different in each branch.

## Reopen / unblock trigger

Resume this issue when:

- The write-integration scoping concludes (either direction).
- A user reports surprise that `AUTO_INDEX__ENABLED=true` does nothing — at which point even an interim "library-only mode" annotation is justified ahead of the full decision.

## Proposal sketch (for when unblocked)

Two prose paths depending on the upstream decision:

- **If server-wired remote write lands:** keep "Stage 4" framing; add a note about the LTM transport / namespace ownership.
- **If library-only:** move auto-index env-var documentation to a "Library hooks" section (or `docs/library-mode.md`); keep an operator-level note in `configuration.md` that `AUTO_INDEX__*` is currently a no-op for the bundled server.

Either way, include a docs-drift regression test pinning the prose against `server.py:187` (assert `proxy_manager._index_engine` is `None` after default-config startup) so accidental wiring does not silently re-validate stale docs.

## Scope

Likely ~30–60 LOC including the regression test. Single PR.

memtomem:
## Summary

- `docs/configuration.md`'s Stage 4 NOTE (added in #339) tells operators that `AUTO_INDEX__*` / `EXTRACTION__*` are inert because the bundled `mms` server constructs `ProxyManager` without an `index_engine`. That claim lives in two files (`docs/configuration.md` + `src/memtomem_stm/server.py`) that can drift independently — silent drift = the original #288 failure mode.
- Add two paired docs-drift guards in `tests/test_docs_sync.py`:
  - **`test_bundled_server_proxy_manager_omits_index_engine`** AST-parses `server.py`, locates the unique `ProxyManager(...)` call, and asserts no `index_engine=` kwarg, exactly 2 positional args (3rd positional would smuggle it past the kwarg check — `index_engine` is `__init__`'s 3rd param), and only allowlisted kwargs. Distinct failure messages distinguish "Stage 4 wired" from "unrelated new kwarg".
  - **`test_configuration_md_stage4_inert_note_pinned`** locates the `#`-comment paragraph above `AUTO_INDEX__ENABLED=`, lowercases the scope, and asserts `inert`, `index_engine`, and `#288` all survive.
- Static AST parsing (not driving `app_lifespan`) is hermetic — the assertion is a source-level claim about the construction site, and `app_lifespan`'s many side effects (sqlite opens, mcp adapter spawn, JSON config load) would each become a new drift surface. Runtime warnings remain covered by `TestAutoIndexStartupWarning`; cross-reference comment points readers to the new pair.

Closes #299. Continues the #288/#339 thread.

## Test plan

- [x] `uv run pytest tests/test_docs_sync.py -v` — 8 passed (6 baseline + 2 new)
- [x] **Negative check A**: temporarily added `index_engine=None` to `server.py:209` — Test A failed with the wired-engine remediation message. Reverted.
- [x] **Negative check B**: temporarily deleted the NOTE block from `docs/configuration.md` — Test B failed with the missing-note remediation message. Reverted.
- [x] `uv run pytest -m "not ollama"` — 2171 passed, 7 skipped (was 2169 + 2 new tests)

### DIFF
--- a/tests/test_docs_sync.py
+++ b/tests/test_docs_sync.py
@@ -8,6 +8,7 @@ in local testing and only surfaces through user confusion.
 
 from __future__ import annotations
 
+import ast
 import re
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #341, opened 2026-05-18
docs(surfacing): clarify LTM bypasses proxy pipeline (#298)

### LINKED ISSUE #298: docs/surfacing.md overstates 'same compression / cache / surfacing pipeline applies' to LTM
## Background

`docs/surfacing.md:191` reads:

> This makes memtomem just another MCP upstream as far as STM is concerned — the same compression / cache / surfacing pipeline applies, and a memtomem crash never takes down STM's other upstream connections.

The "same MCP protocol boundary" half is accurate. The "same compression / cache / surfacing pipeline applies" half is not — LTM is reached via the special-purpose `McpClientSearchAdapter` for `mem_search` / `mem_do`, not registered through the generic upstream proxy path. `mem_search` responses do **not** flow through STM's compression / cache / surfacing pipeline; they feed *into* surfacing as the source of memories.

Evidence:

- `src/memtomem_stm/server.py:126-158` — surfacing engine wired with a dedicated `McpClientSearchAdapter`.
- `src/memtomem_stm/surfacing/mcp_client.py:188` — the adapter only implements `mem_search` / `mem_do` calls, not the generic upstream tool dispatch.

## Proposal

Tighten the wording. Suggested:

> This makes memtomem reachable over the same MCP protocol boundary STM uses for other upstreams. Unlike a generic proxied upstream, LTM responses bypass the compression / cache pipeline — they are consumed by STM's surfacing engine to compose context for upstream calls. A memtomem crash never takes down STM's other upstream connections.

## Test plan

Per the docs-drift regression pattern: pin the wording against an authoritative source.

The authoritative source for "LTM is or isn't a generic upstream" is `server.py` startup wiring. A test in `tests/` that:

1. Loads a default `Config()`.
2. Asserts `Config.proxy.upstream_servers` does not include LTM by default.
3. Greps `docs/surfacing.md` for the now-deprecated phrase ("same compression / cache / surfacing pipeline applies") and fails if it returns.

Mutation check: reverting the prose change must trip the assertion.

## Scope

~20 LOC: prose edit + a focused regression test. Single PR.

## Out of scope

- Rewording `docs/configuration.md:73` "Stage 4 — save large responses to LTM" — that depends on a separate decision about whether server-mode auto-index will be wired at all. Tracked separately.

memtomem:
## Summary

- `docs/surfacing.md`'s `## LTM Connection` claimed LTM is *"just another MCP upstream … the same compression / cache / surfacing pipeline applies"*. The first half is true; the second is not — LTM is reached via `McpClientSearchAdapter` wired into `SurfacingEngine` (`src/memtomem_stm/server.py:138-150`), not via `ProxyConfig.upstream_servers`, so LTM responses bypass the compression/cache pipeline and feed *into* surfacing rather than through it.
- Misleading wording matters for operators tuning `MEMTOMEM_STM_PROXY__*` knobs expecting them to bite LTM traffic.
- Update the prose to keep the accurate MCP-protocol-boundary framing, name the bypass explicitly, point to `surfacing/mcp_client.py`, and keep the crash-isolation claim.
- Add `test_surfacing_md_ltm_connection_distinguishes_from_generic_upstream` in `tests/test_docs_sync.py` pinning three claims: (a) `server.py` still imports `McpClientSearchAdapter` — anchors docs to source; (b) the deprecated phrase doesn't return in the section; (c) replacement keeps `bypass`, `surfacing engine`, and `crash` keywords (lowercase scope).

Closes #298.

## Test plan

- [x] `uv run pytest tests/test_docs_sync.py -v` — 7 passed (6 baseline + 1 new)
- [x] **Negative check**: temporarily restored the deprecated phrase in `docs/surfacing.md` — the new test failed with the deprecated-phrase remediation message naming the import-check fallback path. Reverted.
- [x] `uv run pytest -m \"not ollama\"` — 2170 passed, 7 skipped (was 2169 + 1 new test)

### DIFF
--- a/docs/surfacing.md
+++ b/docs/surfacing.md
@@ -190,7 +190,7 @@ export MEMTOMEM_STM_SURFACING__LTM_MCP_COMMAND=memtomem-server
 export MEMTOMEM_STM_SURFACING__LTM_MCP_ARGS='["--config","/etc/memtomem.json"]'
 ```
 
-This makes memtomem just another MCP upstream as far as STM is concerned — the same compression / cache / surfacing pipeline applies, and a memtomem crash never takes down STM's other upstream connections.
+This makes memtomem reachable over the same MCP protocol boundary STM uses for other upstreams. Unlike a generic proxied upstream, LTM responses bypass the compression / cache pipeline — they are consumed by STM's surfacing engine (via `McpClientSearchAdapter`, see `src/memtomem_stm/surfacing/mcp_client.py`) to compose context for upstream calls, rather than being passed through to the caller. A memtomem crash never takes down STM's other upstream connections.
 
 > **Note**: prior versions supported an in-process mode that imported memtomem directly. That path was removed so STM has a single LTM retrieval path and so core internals can evolve without breaking STM.
 
--- a/tests/test_docs_sync.py
+++ b/tests/test_docs_sync.py
@@ -236,6 +236,82 @@ def test_surfacing_md_documents_phase_1_observability_sample() -> None:
         )
 
 
+def test_surfacing_md_ltm_connection_distinguishes_from_generic_upstream() -> None:
+    """``docs/surfacing.md``'s ``## LTM Connection`` section must not claim
+    that LTM responses flow through the proxy pipeline.
+
+    LTM is reached via the dedicated ``McpClientSearchAdapter``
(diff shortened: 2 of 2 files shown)


pandas-studio (approved):
LGTM

pandas-studio (approved):
LGTM

PR #342, opened 2026-05-18
feat(circuit_breaker): expose opened_at as public read-only property (#277)

### LINKED ISSUE #277: feat(circuit_breaker): expose opened_at as public read-only property
## Motivation

`CircuitBreaker._opened_at` (init at `src/memtomem_stm/utils/circuit_breaker.py:30`,
written at L71) is currently a private attribute, but two recent signals
show it has *effectively-public* semantics. The contract is already
documented — just in the wrong place.

### 1. Tests anchor mocks on `cb._opened_at`

Every time-mocking test in `tests/test_circuit_breaker.py` reads
`cb._opened_at` directly: 9 reads across 8 lines (L61, L71, L83, L93,
L101 ×2, L114, L131, L160, L170) — split between mock anchors
(`mock_time.monotonic.return_value = cb._opened_at + delta`) and
assertions on the recorded timestamp. There is no public alternative —
the private attribute is the de-facto contract surface.

### 2. PR #122 (`5337a1c8`) put the production contract in a test comment

After PR #39 a floating-point precision flake surfaced (caught on
unrelated PR #119 CI ~22h post-merge). The fix landed this paragraph in
`tests/test_circuit_breaker.py:172-178`:

> `time_until_reset` computes `reset_timeout - (now - opened_at)`.
> `opened_at` is a real `time.monotonic()` value (often ~1e5-1e9 seconds
> on CI runners), so `(opened_at + 10.0) - opened_at` loses a few low
> bits and the result is `~1e-14` rather than bit-exact `0.0`.

That paragraph is a **production contract** describing `opened_at`'s
units, magnitude range, and arithmetic precision. It should live as the
docstring of a public property, not as a comment in one test method.

### 3. Operational diagnostics

"When did this circuit trip?" is a routine ops question — useful for
oncall, log correlation, and observability dashboards. A read-only
public property removes the temptation for callers to read the private
attribute directly.

## Proposal

Add a property to `CircuitBreaker`:

```python
@property
def opened_at(self) -> float | None:
    """Monotonic timestamp of the most recent open transition,
    or None if the breaker has never opened.

    After the breaker has opened at least once, this returns the
    timestamp of the *most recent* open transition — even if the
    breaker is currently closed. Use ``state`` to distinguish
    "currently open" from "previously opened, now recovered".

    The value is a real ``time.monotonic()`` reading — typically in
    the ~1e5 to 1e9 second range on long-running processes. Callers
    that subtract two ``opened_at``-derived values must tolerate a
    few ULPs of float residue (see ``time_until_reset`` boundary
    handling).
    """
    return None if self._opened_at == 0.0 else self._opened_at
```

The PR #122 paragraph migrates verbatim into this docstring, plus the
explicit "most recent open transition" clarification.

### Why "most recent" semantics?

`record_success` at `circuit_breaker.py:61-63` does **not** reset
`_opened_at` — it only flips state and zeroes failure count. So a
breaker that opened, recovered, and closed still carries a non-zero
`_opened_at`. The property surfaces that as the historical "last
trip" timestamp, which matches the diagnostic motivation (#3 above).
A closed-then-None semantics would erase that history.

## Tasks

- [ ] Add `opened_at` property + docstring (move PR #122's contract paragraph in, add "most recent open transition" clarification).
- [ ] Verify `_opened_at` init sentinel (currently `0.0` at L30) holds — `time.monotonic()` is implementation-defined but never returns exactly `0.0` on CPython/Linux/macOS in practice. Note the assumption in the property docstring or migrate the field to `Optional[float]` for type clarity (latter is a no-behavior-change refactor since `record_success` already leaves `_opened_at` untouched).
- [ ] Drive-by: switch the 9 test reads from `cb._opened_at` to `cb.opened_at` (8 lines listed above).
- [ ] Add a focused test: `opened_at` is `None` initially → returns the recorded value after `record_failure` triggers open → still equals that value while currently `open`, after auto-transition to `half-open`, and after `record_success` closes the breaker (regression pin for "most recent" semantics).

## Non-goals

- No public-API behavior change. Pure read-only addition.
- Keep monotonic semantics — no wall-clock variant (would diverge from `time_until_reset`'s mental model).
- Not exposing `_state` or `_failures` — both already have public accessors (`state`, `failure_count`).

## References

- `src/memtomem_stm/utils/circuit_breaker.py:30, 71` — backing `_opened_at`
- `src/memtomem_stm/utils/circuit_breaker.py:61-63` — `record_success` (does not reset `_opened_at`, justifies "most recent" semantics)
- `tests/test_circuit_breaker.py:172-178` — contract paragraph to migrate
- PR #39 (added `time_until_reset` + alias tests, closed #28)
- PR #122 / commit `5337a1c8` (fp precision fix where the contract paragraph originated)

memtomem:
## Summary

- `CircuitBreaker._opened_at` was effectively-public already: tests read it directly across 10 lines (mock anchors + assertions), and the production contract for its units/magnitude/float-precision lived as a paragraph inside `test_time_until_reset_at_boundary` (PR #122). Per #277, a production contract belongs on the symbol it describes.
- Add `CircuitBreaker.opened_at -> float | None` with the PR #122 paragraph migrated into its docstring, plus the explicit *most recent open transition* clarification. `record_success` only flips state + zeros failure count — it does not clear `_opened_at` — so the property surfaces the historical \"last trip\" timestamp. Use `state` to distinguish currently-open from previously-opened.
- Switch all 10 test reads from `cb._opened_at` to `cb.opened_at`. Replace the inline contract paragraph with a one-line pointer to the property docstring (keeping the absolute-tolerance rationale visible inline).
- Add `test_opened_at_lifecycle_preserves_most_recent_open`: pins `None → record_failure → still equals while open → still equals after half-open auto-transition → still equals after record_success closes` so a future \"reset on close\" change cannot silently erase the diagnostic.

No public-API behavior change. `_opened_at` field and its `0.0` sentinel unchanged; sentinel assumption noted in docstring (`time.monotonic()` is implementation-defined but never returns exactly 0.0 on CPython in practice).

Closes #277.

## Test plan

- [x] `uv run pytest tests/test_circuit_breaker.py -v` — 21 passed (20 baseline + 1 new)
- [x] `uv run mypy src/memtomem_stm/utils/circuit_breaker.py` — no issues
- [x] `uv run pytest -m \"not ollama\"` — 2173 passed, 7 skipped (was 2172 + 1 new test)

### DIFF
--- a/src/memtomem_stm/utils/circuit_breaker.py
+++ b/src/memtomem_stm/utils/circuit_breaker.py
@@ -51,6 +51,36 @@ class CircuitBreaker:
         return self._failures
 
     @property
+    def opened_at(self) -> float | None:
+        """Monotonic timestamp of the most recent open transition,
+        or None if the breaker has never opened.
+
+        After the breaker has opened at least once, this returns the
+        timestamp of the *most recent* open transition — even if the
+        breaker is currently closed. ``record_success`` flips state
+        back to ``closed`` and zeroes ``failure_count`` but does not
+        clear this timestamp, so the historical "last trip" is
+        preserved for diagnostics. Use ``state`` to distinguish
+        "currently open" from "previously opened, now recovered".
+
+        The value is a real ``time.monotonic()`` reading — typically in
+        the ~1e5 to 1e9 second range on long-running processes.
+        ``time_until_reset`` computes ``reset_timeout - (now - opened_at)``
+        from this; ``(opened_at + 10.0) - opened_at`` loses a few low
+        bits at those magnitudes and the result is ``~1e-14`` rather
+        than bit-exact ``0.0``. Callers subtracting two
+        ``opened_at``-derived values must tolerate a few ULPs of float
+        residue (the test at ``test_time_until_reset_at_boundary``
+        pins an absolute-tolerance check).
+
+        The backing field uses ``0.0`` as a "never opened" sentinel.
+        ``time.monotonic()`` is implementation-defined and could in
+        principle return ``0.0``, but does not in practice on CPython
+        (Linux/macOS/Windows).
+        """
+        return None if self._opened_at == 0.0 else self._opened_at
+
+    @property
     def time_until_reset(self) -> float | None:
         """Seconds until open breaker transitions to half-open. None if not open."""
         if self._state != "open":
--- a/tests/test_circuit_breaker.py
+++ b/tests/test_circuit_breaker.py
@@ -58,7 +58,7 @@ class TestCircuitBreakerHalfOpen:
 
         # Simulate time passing beyond reset_timeout
         with patch("memtomem_stm.utils.circuit_breaker.time") as mock_time:
-            mock_time.monotonic.return_value = cb._opened_at + 11.0
+            mock_time.monotonic.return_value = cb.opened_at + 11.0
             # Should transition to half-open and allow one probe
             assert not cb.is_open
             assert cb._state == "half-open"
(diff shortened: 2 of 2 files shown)


pandas-studio (approved):
LGTM

PR #343, opened 2026-05-18
fix(server): make STM observability tools opt-in

### LINKED ISSUE #228: MCP tool surface: picker noise from STM utility tools vs proxied tools
## Problem

The `memtomem-stm` MCP server advertises STM's utility tools (11 by default) **before** the proxied upstream tools in the MCP surface. Clients that render a linear tool picker (Claude Code `/mcp`, similar pickers elsewhere) show the STM tools first, pushing the domain tools users actually reach for — `langchain__*`, `langfuse__*`, `fs__*`, etc. — to the bottom of the list.

Concrete reproduction from an 18-tool setup today:

```
9.  stm_compression_stats
10. stm_progressive_stats
11. stm_tuning_recommendations
12. langchain__search_docs_by_lang_chain
13. langchain__query_docs_filesystem_docs_by_lang_chain
…
```

Items 1-11 are STM utility tools. Items 12+ are the proxied tools the user is actually trying to browse. The picker preserves registration order (not alphabetical — otherwise `langchain__*` would precede `stm_*`).

## Why now (incident count)

First observation today. Filed before a second one because the existing partial mitigation (below) doesn't fully close the gap even when enabled, and the design space is wide enough that early discussion beats a late scramble.

## Existing partial mitigation — doesn't fully solve it

`docs/cli.md:263-297` documents `MEMTOMEM_STM_ADVERTISE_OBSERVABILITY_TOOLS=false` (default `true`). When set, the STM surface drops from 11 → 4 (hides 7 observability tools). The remaining 4 are model-facing pipeline tools that can't be turned off:

- `stm_proxy_select_chunks`, `stm_proxy_read_more` — progressive-delivery callbacks
- `stm_surfacing_feedback`, `stm_compression_feedback` — learning-signal channels

So even with the flag set to `false`, **4 STM tools still appear above the first proxied tool** in the picker. The flag covers the context-token cost angle but leaves the picker-UX angle open.

### A) Flip registration order — proxied tools first, STM utility tools last

Keep the existing flat tool surface; change the advertise order so proxied tools appear first. STM utility tools move to the end of the list.

- Scope: one change to the advertise path.
- Tradeoff: picker is now ordered by "what the user is reaching for" vs "what supports the pipeline". No behavior change, no flags, no migrations.
- Doesn't help eager-load MCP clients pay for fewer tool schemas — unchanged total count.

### B) Flip `ADVERTISE_OBSERVABILITY_TOOLS` default `true` → `false`

STM's 11 observability tools are (per docs) "accessible through this very CLI" — the canonical interface is `mms`, not MCP. Advertising them by default pays tool-surface cost on every session for rarely-called admin tools.

- Scope: one default change + CHANGELOG + migration note.
- Tradeoff: breaking change for any user who was calling observability tools via MCP; mitigated by documenting the flip and the opt-in flag (`=true`).
- Shrinks STM surface from 11 → 4 by default; combined with (A), proxied tools are first in the picker and STM surface is minimal.

### C) Group-as-one-tool, `mem-do` style

Collapse the flat tool surface into per-group meta-tools:

- `stm_util(op="stats|health|cache_clear|select_chunks|read_more|…", …)` — one tool exposing all STM operations under an `op` enum.
- One wrapper per proxied upstream: `langfuse_docs(op="search|query", …)`, `langchain_docs(op=…)`, `filesystem(op=…)`, etc., synthesized at proxy startup from the upstream's `list_tools` response.

- Scope: substantial. Needs a runtime schema-synthesis layer that reads each upstream's per-tool `inputSchema` and composes a single group-tool `inputSchema` where `op` gates a branching argument structure. Also needs an opt-in flag (`MEMTOMEM_STM_GROUP_TOOLS=true` or similar) so existing users don't break.
- Tradeoff: picker drops from 18 → ~4 items (massive UX win); eager-load MCP clients (OpenAI Codex, …) pay schema cost for a handful of group tools instead of the full list; **but** sub-tool schemas are no longer first-class in the picker — each op's arguments have to be described inline in the group tool's description/schema. Needs an end-to-end prototype to measure real model-side ergonomics.
- Standard MCP-client compatibility is unknown until tested across clients.

## Recommended sequencing

Single issue covers the whole design space (A/B/C are connected — picking C later probably removes the need for A/B; shipping A now does not block C), but the **work** ships incrementally:

1. **Ship (A) first.** Biggest fraction of the observed pain for the lowest risk. No flag, no migration, no behavior change for callers. ~30 min.
2. **Discuss (B).** Breaking-change judgment call — the default-flip might be the right move given the docs already steer users to `mms`, but a survey of callers who rely on MCP access is warranted first. ~1 h plus discussion.
3. **Prototype (C) in a separate follow-up issue.** Schema synthesis, opt-in flag, multi-client compatibility testing, and measurement (context tokens + model-side ergonomics) are a multi-day workstream. Don't block (A) on it.

## Out of scope

- Reordering **within** the proxied-tool bucket (cross-server ordering of `fs__*` vs `gh__*`). Currently follows the order servers are declared in `stm_proxy.json`; no user pain signal here.
- Client-side picker filters (Claude Code `ToolSearch`, Codex `disabled_tools`). Those are downstream levers the user already has — this issue is about the default STM serves.
- Renaming utility tools (`stm_*` → `_stm_*` or similar) to bias alphabetical picker clients. The non-alphabetical observation from the repro already rules out relying on client-side sorting.

## Acceptance criteria

Minimum viable set for **phase 1** (alt A):

- [ ] `memtomem_stm.server` (or whichever layer assembles `list_tools`) advertises proxied tools before STM utility tools.
- [ ] No change to STM utility tool set, names, or behavior.
- [ ] Test that pins the order invariant — regression guard so a refactor can't silently flip it back.

**Phase 2** (alt B) gated on a signal survey; **phase 3** (alt C) split to a separate issue once A/B are shipped.

## Reopen / resolution triggers

- **Close A once merged** with the order-pinning test. Keep the issue open for B discussion.
- **Reopen (C)-discussion** only if: (i) a second user reports picker noise after A ships, or (ii) an eager-load client (Codex CLI et al.) shows measurable context-token waste from the flat advertise list.
- **Close-as-wontfix (B)** acceptable if the default-flip survey surfaces significant MCP-access usage of observability tools.

## Priority

Medium — UX papercut for every session, but no functional blocker. `ADVERTISE_OBSERVABILITY_TOOLS` already exists as a partial escape hatch for users who care today.

memtomem:
## Summary
- default MEMTOMEM_STM_ADVERTISE_OBSERVABILITY_TOOLS to false so only model-facing STM tools advertise by default
- keep MEMTOMEM_STM_ADVERTISE_OBSERVABILITY_TOOLS=true as the MCP opt-in for admin/observability tools
- update docs, changelog, and registration-time regression coverage

Closes #228

## Tests
- uv run pytest tests/test_server_tools.py
- uv run pytest tests/test_docs_sync.py
- uv run ruff check src/memtomem_stm/server.py src/memtomem_stm/config.py tests/test_server_tools.py
- uv run pytest


memtomem:
Addressed the review:

- **Rebased onto `main`** to drop the carry-over commit (`3a56f14`, the #277 `opened_at` property — that work lives on its own PR #342). The PR now contains only the phase-2 work plus the prose polish below.
- **Tightened `docs/cli.md`** "Trimming the advertised MCP tool surface" paragraph — the original repeated "the default four are model-facing" after already saying "STM advertises four model-facing MCP tools by default". Folded the parenthetical into the opening sentence.

Skipped the optional follow-up (a `test_docs_sync.py` pin coupling `_OBSERVABILITY_TOOLS` to the docs tool count) — happy to add it as a separate PR if desired, since it's a drift-guard pattern shared with #298/#299 rather than #228 scope.

Verified locally:
- `uv run pytest tests/test_server_tools.py tests/test_circuit_breaker.py tests/test_docs_sync.py` — 85 passed
- `uv run pytest -m "not ollama"` — 2173 passed, 7 skipped
- `uv run ruff check src && uv run ruff format --check src` — clean

pandas-studio (approved):
LGTM

PR #346, opened 2026-05-18
test(guard): reject module-level POSIX-only stdlib imports (#302 P3)

memtomem:
## Summary

Closes the last P-row on the #302 Windows-support triage — preventive AST guard against module-level POSIX-only stdlib imports (`fcntl`, `pwd`, `grp`, `termios`, `resource`, `syslog`, etc.).

A bare top-level `import fcntl` crashes on Windows at import time even when no call site uses it, because Python evaluates `import` statements eagerly. The only safe way to depend on these modules is from inside an `if sys.platform != \"win32\":` block or a `try/except ImportError` wrapper — both of which are nested in the AST (not direct children of `Module.body`) and therefore correctly skipped by the walker.

`stm` has **zero** such imports today (verified during the triage — see #302's "Not applicable" section), so this PR is pure regression insurance. It mirrors the AST check that memtomem core added in #657.

## Why a pytest test instead of a separate CI step / ruff rule

- Runs in the existing `test (ubuntu-latest)` and `test (windows-latest)` matrix jobs — both gating since #320. No new CI step to wire up, no new `required_status_checks` axis to coordinate.
- Failure surfaces as an ordinary test failure on the offending PR with `path:lineno: import <name>` in the assertion message — directly actionable.
- Ruff doesn't ship a built-in rule for "ban these stdlib imports at module scope only"; the AST walk is ~15 lines and the rule list is self-documenting.

## Test plan

- [x] `uv run pytest tests/test_no_posix_only_imports.py -v` — 4 passed locally (the main guard plus three self-tests covering the positive case, the platform-guard case, and the `try/except ImportError` case).
- [x] `uv run ruff check tests/test_no_posix_only_imports.py && uv run ruff format --check tests/test_no_posix_only_imports.py` — clean.
- [ ] CI on this PR: both Ubuntu and Windows `test` axes pick up the new test and pass.
- [ ] Sanity: temporarily adding `import fcntl` to any module under `src/` should make this test fail with a clear message pointing at the offender (not exercised in the PR, but the self-tests prove the walker would catch it).

## Follow-up

Once green, #302 can be closed — all P-rows (P0 / P1a–g / P2 / P3) are now landed and Windows is a required matrix axis (#320).

### DIFF
--- a/tests/test_no_posix_only_imports.py
+++ b/tests/test_no_posix_only_imports.py
@@ -1,0 +1,92 @@
+"""POSIX-only stdlib imports must not appear at module level in ``src/``.
+
+Python evaluates ``import`` statements eagerly. A bare ``import fcntl`` at the
+top of a module raises ``ModuleNotFoundError`` on Windows the moment the
+module is loaded — even when no call site uses it. Conditional imports nested
+under a platform guard (``if sys.platform != "win32":``) or wrapped in
+``try/except ImportError`` only execute on POSIX and are safe.
+
+``stm`` has zero such imports today (verified during the #302 Windows triage).
+This guard exists to keep it that way; mirrors memtomem #657.
+"""
+
+from __future__ import annotations
+
+import ast
+from pathlib import Path
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+SRC_ROOT = REPO_ROOT / "src"
+
+# Stdlib modules that exist only on POSIX builds of CPython. A module-level
+# import of any of these crashes on Windows at import time.
+POSIX_ONLY_STDLIB_MODULES = frozenset(
+    {
+        "fcntl",
+        "grp",
+        "nis",
+        "ossaudiodev",
+        "posix",
+        "pwd",
+        "resource",
+        "spwd",
+        "syslog",
+        "termios",
+    }
+)
+
+
+def _module_level_imports(tree: ast.Module) -> list[tuple[int, str]]:
+    """Return ``(lineno, top_level_module_name)`` for imports at module scope.
+
+    Imports nested inside ``If`` / ``Try`` / function / class bodies are
+    excluded — those are the legitimate way to depend on a POSIX-only module
+    without breaking Windows.
+    """
+    found: list[tuple[int, str]] = []
+    for node in tree.body:
+        if isinstance(node, ast.Import):
+            for alias in node.names:
+                found.append((node.lineno, alias.name.split(".")[0]))
+        elif isinstance(node, ast.ImportFrom) and node.module:
+            found.append((node.lineno, node.module.split(".")[0]))
+    return found
+
+
+def test_no_posix_only_top_level_imports_in_src() -> None:
+    paths = sorted(SRC_ROOT.rglob("*.py"))
+    assert paths, f"no python sources discovered under {SRC_ROOT}"
+
+    offenses: list[str] = []
+    for path in paths:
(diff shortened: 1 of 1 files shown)


memtomem:
Addressed both external-review P2 comments in a11bc69:

**(1) `pty` and `tty` added to the POSIX-only set.** Both are "Availability: Unix" in CPython docs and transitively pull in `termios` — they belonged in the original frozenset.

**(2) Walker no longer blanket-exempts every nested `If`/`Try`.** The original implementation only inspected `tree.body`, so anything nested inside an `if`/`try` was skipped — including `if True: import fcntl` and `try: import fcntl except OSError`, both of which still crash on Windows at load time.

The new walker recurses into `If`/`Try` at module scope and flips `exempt=True` only when the construct matches a recognized pattern:

| construct | exempts when |
|-----------|-------------|
| `if <test>:` | `<test>` is `sys.platform != "win32"`, `os.name != "nt"`, or `os.name == "posix"` |
| `try: ... except <T>:` | `<T>` includes `ImportError`, `ModuleNotFoundError`, `Exception`, `BaseException`, or is bare `except:` |

Everything else (`if True:`, `if some_var:`, `if sys.platform == "darwin":`, `except OSError:`, etc.) now correctly flags. Function/class bodies are still skipped — those don't run at module load, and a call-time POSIX import is a separate bug class outside this guard's scope.

Self-tests expanded from 3 → 7 to pin both the positive case (each recognized form exempts) and the negative case (each unrecognized form flags), so a future refactor of `_is_posix_platform_guard` or `_try_catches_import_error` can't silently loosen the rule. All 8 tests green locally; ruff clean.

pandas-studio (approved):
LGTM

memtomem:
Addressed in e8013fc.

`ast.ClassDef.body` is now walked with inherited `exempt` (no `class` construct itself confers exemption). Function bodies — including `async def` and methods nested inside the now-walked class bodies — remain skipped, since those don't run until called and represent a separate (call-time) bug class.

Self-tests added to pin both directions:
* `test_guard_flags_class_body_imports` — the previously missed positive case (`class C: import fcntl` now flags)
* `test_guard_skips_class_method_body_imports` — the negative case, ensuring the new ClassDef recursion doesn't accidentally start flagging method-body imports too

`test_guard_skips_function_body_imports` also extended to cover `async def` explicitly. 10/10 tests pass locally; ruff clean.

Note: there's an adjacent gap class the walker still silently skips — `With`, `For`, `While`, `Match` at module top level all execute at load time. Most likely real pattern is `with contextlib.suppress(ImportError): import fcntl`. Flagging here so it's on record; happy to roll that into this PR if preferred, otherwise will spin a small follow-up after merge.

memtomem:
Pushed a9da0be — closes the remaining load-time blind spot (`With` / `For` / `While` / `Match`).

| construct | exempts when | notes |
|-----------|-------------|-------|
| `with` | `with contextlib.suppress(ImportError, ...):` (incl. bare `suppress`) | analog of `try/except ImportError`; covers the practical case |
| `for` / `while` | nothing | loops at module scope are recursed but never exempt; canonical guard is `if sys.platform != "win32":` |
| `match` | nothing | platform dispatch via `match sys.platform: case "linux": ...` is not recognized; rewrite as `if` |

`async with` / `async for` are SyntaxErrors outside `async def`, so they can't appear at module load — no handler needed.

Self-tests grew 7 → 12 to pin both directions for each new construct. 15 total, all green; ruff clean.

PR #354, opened 2026-05-19
Fix already_known auto-tune direction

### LINKED ISSUE #347: Surfacing AutoTuner: already_known feedback inverts learning direction
## Background

`AutoTuner` adjusts per-tool `min_score` from the ratio of `not_relevant` feedback over total feedback. The numerator is filtered to a single rating, but the denominator includes **all** ratings — including `already_known`:

- Numerator: `WHERE rating = 'not_relevant'` (`src/memtomem_stm/surfacing/feedback_store.py:443-473`, `get_tool_not_relevant_ratio`)
- Denominator: `total feedback rows` (same method)
- Tuner branches: `ratio > 0.6` → raise threshold; `ratio < 0.2` → lower threshold (`src/memtomem_stm/surfacing/feedback.py:110-153`)

`already_known` is intended as a negative signal — the agent telling STM "stop surfacing this, I know it already." Cache invalidation handles the in-session case (`src/memtomem_stm/surfacing/engine.py:268-310`), but for AutoTuner learning the rating is silently bucketed as non-negative.

## Effect

Concrete scenario: 100 `already_known` + 0 `helpful` + 0 `not_relevant` over the auto-tune window:

- ratio = 0 / 100 = 0.0
- `ratio < 0.2` branch fires → `min_score` decreases by `auto_tune_score_increment`
- Next surfacing for the same tool pulls **more** memories above the lower threshold — the opposite of the agent's signal.

Long-term effect: a tool that the agent repeatedly tells "I already know these memories" gets *more aggressive* surfacing over time, with the same memories re-passing the score cutoff in subsequent sessions. The cache invalidation set (`_invalidated_ids`) is in-memory per session, so it does not survive process restart and does not push back against the tuner.

## Proposal

Two reasonable options — pick one and document the rationale:

**Option A — count `already_known` as negative.** Include it in the numerator alongside `not_relevant`:

\`\`\`sql
WHERE f.rating IN ('not_relevant', 'already_known')
\`\`\`

Rename `get_tool_not_relevant_ratio` → `get_tool_negative_ratio` and update the AutoTuner log to read \"negative ratio\".

**Option B — three-way ratio.** Track `helpful`, `negative` (= not_relevant + already_known), and `neutral` separately, branch on `helpful / total` instead. Better calibration but a wider change.

Option A is the minimal correct fix and matches the existing increment/decrement shape — propose starting there.

## Out of scope

- Per-memory LTM demotion when negative feedback accumulates (separate issue).
- Migration of historic per-tool adjustments — process-local adjustments are already rebuilt from store on restart, but operators with persisted adjustments arrived at via the broken ratio may want a one-time reset.

## Test pin

Add a regression test that submits N `already_known` ratings for a tool and asserts `min_score` does **not** decrease (instead either stays put or increases, depending on Option chosen).

---

pandas-studio:
## Summary
- count already_known with not_relevant as negative feedback for surfacing AutoTuner
- expose negative_count in surfacing stats so reported ratios match tuner behavior
- update docs and add regression coverage for already_known raising min_score

## Verification
- uv run pytest tests/test_feedback.py tests/test_server_tools.py::TestSurfacingStats
- uv run pytest tests/test_stm_integration.py::TestAutoTunerLoop tests/test_effectiveness.py::TestAutoTunerConvergence tests/test_effectiveness.py::TestAutoTunerColdStart
- uv run ruff check src/memtomem_stm/surfacing/feedback_store.py src/memtomem_stm/surfacing/feedback.py src/memtomem_stm/server.py tests/test_feedback.py
- git diff --check

Closes #347


tsdata (approved):
LGTM

PR #355, opened 2026-05-19
Surface memories on progressive path by default (#348)

### LINKED ISSUE #348: Surfacing: injection_mode='prepend' default silently skips surfacing on progressive path
## Background

`SurfacingConfig.injection_mode` default is `\"prepend\"` (`src/memtomem_stm/surfacing/config.py:46`). On the progressive-delivery path, `ProxyManager._apply_surfacing_on_progressive` short-circuits and returns the response unchanged whenever the engine is in prepend mode (`src/memtomem_stm/proxy/manager.py:794-799`):

\`\`\`python
if self._surfacing_engine.injection_mode == \"prepend\":
    if self._log_once(...):
        logger.warning(\"Progressive surfacing skipped: injection_mode='prepend' ...\")
\`\`\`

The reason is real — prepend would shift character offsets and break the `PROGRESSIVE_FOOTER_TOKEN` concat invariant `stm_proxy_read_more` depends on (`docs/compression.md:181`). The problem is the **default** combination: a fresh `mms init` produces a config where surfacing is enabled and progressive triggers are wired, but every progressive-strategy call (or ratio-guard progressive fallback) silently surfaces nothing. The operator only sees one WARNING on the first progressive call.

## Effect

- Largest responses (the ones progressive was designed for) are exactly the responses surfacing is most useful for, and exactly the path that default settings disable.
- `stm_surfacing_stats` does not count this skip. There is no `record_skip(tool, \"progressive_prepend\")` call at `manager.py:794`, so operators looking at the stats see zero evidence that surfacing was bypassed. The closest signal is the single WARNING log line — easy to miss after process restart.

## Proposal

Two parts, tracked as one issue because they are coupled:

**Part 1 — change default.** Flip `injection_mode` default to `\"append\"`. Append preserves the progressive invariant and the agent-facing UX is similar (most agents read full responses; placement does not change which memories surface).

- Tradeoff: existing operators relying on prepend get a different default. Make this a single CHANGELOG entry and surface the new default in `mms init` output for transparency. No deprecation needed because `\"prepend\"` remains valid — only the default moves.

**Part 2 — count the skip.** Add an observability hook on the manager-side prepend-skip path so it shows up in `stm_surfacing_stats`:

- New `SkipReason` enum value, e.g. `\"progressive_mode_conflict\"` (label kept generic so a future per-mode constraint can reuse it).
- `record_skip(tool, \"progressive_mode_conflict\")` at the early-return in `manager.py:794`.
- Document in `docs/surfacing.md` so the skip label appears in the operator-facing tour.

## Out of scope

- Designing a prepend-safe progressive offset rewrite. The reason the skip exists is the concat invariant; that is a structural redesign, not a default-flip.
- Removing the `\"prepend\"` mode entirely. It remains useful for non-progressive responses where the agent should see memories first.

## Test pin

- Default-change test: instantiate `SurfacingConfig()` and assert `injection_mode == \"append\"`.
- Observability test: drive a progressive call with `injection_mode='prepend'`, assert the observability snapshot contains the new skip reason for the tool.

---

pandas-studio:
## Summary

- Flip `SurfacingConfig.injection_mode` default from `prepend` to `append` so the progressive-delivery path actually surfaces memories under a fresh `mms init` config.
- Add `progressive_mode_conflict` `SkipReason` and call `record_skip` at `ProxyManager._apply_surfacing_on_progressive`'s prepend short-circuit so the bypass shows up in `stm_surfacing_stats` instead of relying on a one-shot WARNING log.
- Update `docs/surfacing.md` example + default table to match.

Closes #348.

## Why now

`_apply_surfacing_on_progressive` short-circuits when `engine.injection_mode == "prepend"` because prepending memories would shift character offsets and break the `PROGRESSIVE_FOOTER_TOKEN` concat invariant that `stm_proxy_read_more` depends on (see `docs/compression.md` Stage 3 note). That guard is correct. The problem is that `"prepend"` was *also* the default, so every progressive call under a fresh `mms init` silently skipped surfacing — exactly the largest responses for which surfacing is most useful. Operators only saw one WARNING per process and no `stm_surfacing_stats` counter, so the bypass was easy to miss after a restart. The project's own bench harness (`tests/bench/bench_qa/runner.py:191`) already overrides to `append` for this reason.

## Behavior change

- Operators relying on the **implicit** `prepend` default now get memories appended to the response instead of prepended.
- Operators with an **explicit** `injection_mode: prepend` in their config see no change.
- `prepend` remains a valid choice and is still useful for non-progressive responses where the agent should see memories first.
- New `progressive_mode_conflict` skip reason starts appearing in `stm_surfacing_stats` for any operator who keeps `injection_mode: prepend` while using progressive delivery. The label is mode-agnostic so a future per-mode constraint can reuse it without a new enum value.

## Test plan

- [x] `tests/test_config_constraints.py::TestSurfacingNumericConstraints::test_surfacing_injection_mode_default_is_append` pins the new default.
- [x] `tests/test_proxy_manager_pipeline.py::TestApplySurfacingOnProgressive::test_prepend_mode_records_skip_in_observability` drives a progressive call in `prepend` mode with a real `SurfacingObservability` and asserts the snapshot has per-tool + `__total__` counts for `progressive_mode_conflict`.
- [x] `test_prepend_mode_skip_works_without_observability` covers the engine-without-observability path so the new `record_skip` site is null-safe.
- [x] Existing `test_prepend_mode_skips_and_warns_once` still passes — WARNING-once contract preserved.
- [x] `uv run pytest tests/ -m "not ollama"` — 2200 passed, 1 pre-existing flake unrelated to this PR (`test_search_returns_empty_when_no_session`, same failure on main `24df9ee` before my changes).
- [x] `uv run ruff check` + `uv run ruff format --check` on touched files clean.


tsdata (approved):
LGTM

Issue #349, opened 2026-05-19
Surfacing: mms init/health does not probe LTM MCP reachability — silent ltm_unavailable


pandas-studio:
## Background

`SurfacingConfig.enabled` defaults to `True` (`src/memtomem_stm/config.py:58`), but neither `mms init` nor `mms health` verifies that the configured `ltm_mcp_command` (default `\"memtomem-server\"`) can actually be spawned.

- `mms init` prompts for upstream servers, language preset, and pruning — nothing for surfacing (`src/memtomem_stm/cli/proxy.py:1715-...`).
- `mms health`'s surfacing bootstrap only inspects the feedback DB at `~/.memtomem/stm_feedback.db` (`src/memtomem_stm/cli/proxy.py:2189-2209`). It does not check `which $ltm_mcp_command` or attempt a probe spawn.

## Failure mode

When `memtomem-server` (or whatever `ltm_mcp_command` resolves to) is not on PATH or fails to start, `McpClientSearchAdapter` flips `_start_attempted=True` on first failure and returns `outcome=\"no_session\"` for every subsequent call (`src/memtomem_stm/surfacing/mcp_client.py:343`). The engine maps that to `record_skip(tool, \"ltm_unavailable\")` (`src/memtomem_stm/surfacing/engine.py:511`). Surfacing is permanently disabled for the process lifetime with no error surfaced — the only signal is the `ltm_unavailable` counter inside `stm_surfacing_stats`, which the operator has to know to look at.

This is distinct from the circuit-breaker path (which fires on repeated mid-call failures); the sticky-no-session path is silent at startup.

## Proposal

**Part 1 — `mms init` LTM prompt.** Add a step at the end of the init flow:

\`\`\`
Enable LTM surfacing? (requires an MCP-compatible memtomem server)
  [auto]  detect — probe `memtomem-server`, fall back to disabled with a note
  [yes]   always enabled (default; record path you used)
  [no]    skip surfacing
\`\`\`

The `auto` branch runs `shutil.which(ltm_mcp_command)` and a 1-second `--help`-style probe; on failure it writes `surfacing.enabled = false` with a one-line config comment explaining how to re-enable. TTY-only — non-TTY callers keep the current default.

**Part 2 — `mms health` LTM probe.** Extend `_surfacing_bootstrap_status` to spawn `ltm_mcp_command` once with a short timeout and report:

\`\`\`
Surfacing Bootstrap
==============================
  config: enabled
  feedback tracking: enabled
[...]

PR #358, opened 2026-05-20
feat(cli/health): probe LTM MCP reachability in mms health (#349 part b)

pandas-studio:
## Summary

- Add a stdio MCP probe (`ClientSession.initialize` + best-effort `mem_do(action="version")`) to `_surfacing_bootstrap_status`, gated by a fast `shutil.which` PATH check.
- Render the result in `mms health` as a new `ltm server:` row: `connectable (cmd, version X)`, `UNREACHABLE — <reason>`, or `skipped (surfacing disabled)`.
- Closes part (b) of #349. Part (a) (warn-once) shipped in #356. Part (c) (`mms init` LTM prompt) is deferred wait-for-signal per the #356 decision (operators commonly bootstrap STM before LTM; an init-time check would force a re-enable later).

## Why this is the next layer on top of #356

#356 added a single WARNING the first time the adapter returns `no_session` / `transport_error` — but the operator still has to read the log stream during a real call. `mms health` is the proactive diagnostic surface; failing to probe LTM there is the gap the issue calls out (`src/memtomem_stm/cli/proxy.py:2189-2209`). The probe is fast (PATH check first, then a stdio handshake bounded by `surfacing.timeout_seconds`), best-effort (failures are a warning row, not a non-zero exit — consistent with how upstream probes already report), and reuses `_root_cause_message` so anyio `TaskGroup` wrappers don't leak `"unhandled errors in a TaskGroup (1 sub-exception)"` into the line.

## Behavior change

- New health output row when surfacing is enabled:
  - `ltm server: connectable (memtomem-server, version 0.3.0)` — initialize succeeded; version is best-effort, omitted on parse failure.
  - `ltm server: UNREACHABLE — memtomem-server not on PATH` — fast-reject from `shutil.which`.
  - `ltm server: UNREACHABLE — <display>: timeout (3s)` — handshake exceeded `surfacing.timeout_seconds`.
  - `ltm server: UNREACHABLE — <display>: <root cause>` — other spawn/initialize errors.
  - `ltm server: skipped (surfacing disabled)` — when `surfacing.enabled=false`.
- `mms health --json` adds `surfacing.ltm_server` with `command`, `args`, `display`, `connected`, `version`, `error`, and optional `skipped`.
- No change to exit code, `mms health` does not gate on LTM reachability.
- No change to the WARNING-once behavior added in #356 — that path still fires the first time a real surfacing call hits an unreachable LTM.

## Test plan

- [x] `tests/cli/test_proxy_cli.py::TestHealth::test_health_reports_unreachable_ltm_server` — `__missing_ltm__` on PATH, asserts the `UNREACHABLE` line and `not on PATH` reason.
- [x] `tests/cli/test_proxy_cli.py::TestHealth::test_health_reports_connectable_ltm_server` — drives the probe against `tests/_fake_memtomem_server.py`, asserts the `connectable` line and `version 0.3.0-fake`.
- [x] Existing `TestHealth` cases extended: `test_health_default_surfacing_block` now also asserts `ltm_server.connected is False` under the autouse `__missing_ltm__` env, locking the JSON shape.
- [x] `uv run pytest tests/cli/test_proxy_cli.py -q` → 205 passed.
- [x] `uv run ruff check src/memtomem_stm/cli/proxy.py tests/cli/test_proxy_cli.py` → clean.
- [x] `uv run pytest tests/test_no_posix_only_imports.py -q` → 15 passed (the new `import shutil` is stdlib-portable; guard would catch a slip).

### DIFF
--- a/src/memtomem_stm/cli/proxy.py
+++ b/src/memtomem_stm/cli/proxy.py
@@ -8,6 +8,7 @@ import logging
 import os
 import re
 import shlex
+import shutil
 import subprocess
 import sys
 import tomllib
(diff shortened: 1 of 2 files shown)


pandas-studio:
Fixed in 4d354fd. Reproduced locally — `CliRunner`'s in-memory stderr (no `fileno()`) caused `stdio_client` → `create_subprocess_exec` to reject with `UNREACHABLE — …: fileno`. Rewrote `test_health_reports_connectable_ltm_server` to spawn through `subprocess.run` with the same `from memtomem_stm.cli.proxy import cli; cli()` shim that `TestAddValidate::test_validate_success_writes_entry_with_tool_count` (test_proxy_cli.py:879-918) already uses for the same constraint. Unreachable-path test stays on `CliRunner` since its `shutil.which` reject never spawns a child. `uv run pytest tests/cli/test_proxy_cli.py::TestHealth -q` → 16 passed.

memtomem:
Both addressed in 1abbcb5.

**P2 — verify `mem_search` before reporting connectable.**
Probe now calls `list_tools()` after `initialize()`. If `mem_search` isn't advertised, returns `connected=False` with a specific error pointing at the misconfiguration:

```
ltm server: UNREACHABLE — <cmd>: server initialized but does not expose
  'mem_search' (surfacing adapter calls this tool — config likely points
  at a non-memtomem MCP server)
```

Version probe is now gated on `mem_do` being advertised first, so older servers that don't recognize `action="version"` log a single debug line and don't poison the result. New test `test_health_flags_ltm_server_missing_mem_search` drives a minimal FastMCP server that exposes only an unrelated tool — asserts the UNREACHABLE branch and the `mem_search` mention in the error.

**P3 — silence LTM probe stderr.**
LTM probe path now wraps `asyncio.run` with `_silenced_mcp_sdk_logs()` (parity with `_probe_servers`) and passes `open(os.devnull, "w")` as `stdio_client`'s `errlog`. `io.StringIO` was the obvious first choice but lacks `fileno` — `create_subprocess_exec` rejects it, the same trap that 4d354fd worked around for `CliRunner`. The devnull sink is fd-backed so the subprocess accepts it, and child stderr / SDK exception logs no longer leak into `mms health` or `mms health --json`.

`uv run pytest tests/cli/test_proxy_cli.py -q` → 206 passed (16 in `TestHealth`, +1 new). `uv run ruff check` clean.

memtomem:
Fixed in da90f25.

`--timeout` now threads from `health(timeout=...)` → `_surfacing_bootstrap_status(timeout)` → `_ltm_mcp_status(surfacing, timeout)` → both `asyncio.wait_for` calls in the probe. `surfacing.timeout_seconds` is no longer read by the probe path — the CLI flag is the single bound.

While writing the regression test I also caught a latent dispatch bug the previous design hid: when `asyncio.wait_for` times out inside the probe, anyio's `TaskGroup` (wrapped by `stdio_client`) re-raises the `TimeoutError` as an `ExceptionGroup` leaf, so the prior `except asyncio.TimeoutError` branch silently missed it and the error rendered as the fallback type name `TimeoutError` (visible in the first run of the new test before the dispatch fix). Extracted `_root_cause_exc` from `_root_cause_message` and now dispatch on `isinstance(root, TimeoutError)` — wrapped and bare cases share one path. Output is now `timeout (Ns)` consistently.

New test `test_health_ltm_probe_honors_cli_timeout` runs a server that just `time.sleep(60)`s, sets `MEMTOMEM_STM_SURFACING__TIMEOUT_SECONDS=30`, and asserts `mms health --timeout 1` returns within the subprocess's 15s watchdog with `UNREACHABLE — …: timeout (1s)`. Pre-fix this would have blocked for 30s and tripped the outer timeout.

`uv run pytest tests/cli/test_proxy_cli.py -q` → 207 passed. `uv run ruff check` clean.

memtomem:
Fixed in 47deab0.

Probe now tracks a single deadline at entry (``deadline = loop.time() + timeout``) and hands each ``wait_for`` ``deadline - now``. After ``list_tools`` confirms ``mem_search``, the version probe is skipped entirely if less than ~50ms of budget remains so we don't fire a ``wait_for`` we know will trip immediately; if any remains, ``mem_do(action="version")`` runs against that residual. A version-phase ``TimeoutError`` still gets swallowed and ``version`` reported as ``None`` — ``connected`` stays true because ``mem_search`` connectivity was already proven by ``list_tools``.

New regression test ``test_health_ltm_probe_shares_timeout_across_phases`` runs an inline FastMCP server that handshakes + lists tools fast but ``await asyncio.sleep(60)``s inside ``mem_do(action="version")``. With ``--timeout 1`` and ``MEMTOMEM_STM_SURFACING__TIMEOUT_SECONDS=30``, asserts:

- ``ltm_server.connected is True`` (mem_search proven)
- ``ltm_server.version is None`` (budget-starved → silently dropped)
- ``elapsed < 5.0s`` (pre-fix would have spent ~30s on the version stall and tripped the outer 15s subprocess timeout)

Local measurement: 1.71s wall time, mostly child Python + MCP startup. `uv run pytest tests/cli/test_proxy_cli.py::TestHealth -q` → 19 passed. `uv run ruff check` clean.

Issue #352, opened 2026-05-19
Surfacing privacy: full query persisted in feedback DB + query[:50] at INFO log


pandas-studio:
## Background

The surfacing pipeline currently leaks user-derived query text in two places:

1. **Feedback DB persistence** — `surfacing_events.query TEXT NOT NULL` (`src/memtomem_stm/surfacing/feedback_store.py:22`). `record_surfacing` writes the full extracted query verbatim (`feedback_store.py:138-144`). Stored indefinitely; subject only to the operator manually pruning `~/.memtomem/stm_feedback.db`.

2. **INFO log lines** — `engine.py:579-580` and `relevance.py` log `query[:50]` at INFO level on the surfacing hot path. The truncation reduces but does not eliminate sensitive content — file paths, partial commit messages, or the first sentence of a user-provided description all fit in 50 chars.

`proxy/privacy.py` exists but is routing-only (LLM-vs-hybrid compression decision) and does not redact (`reference_proxy_privacy_scope.md`). There is no current redaction layer for surfacing queries.

## Why this matters

The extracted query is whatever the `ContextExtractor` synthesizes from tool arguments — typically file paths, search terms, the first sentence of a description argument, or an explicit `_context_query`. For an agent running against a private codebase or internal documents, this text frequently includes:

- Internal file paths revealing project structure.
- The first 200 chars of issue / ticket / commit text (via `_first_sentence`).
- User-typed search queries when the tool is something like `search_issues` or `read_file`.

Persisting this in a local SQLite store is defensible — the store is per-user. The combination of (a) **indefinite retention**, (b) **no documentation that the DB contains query text**, and (c) **INFO-level log emission** is the gap.

## Proposal

Three parts. They are independent — file as one issue, ship as one or more PRs as appropriate.

**Part 1 — log level.** Drop the two `query[:50]` INFO lines to DEBUG (`engine.py:579`, the equivalent in `relevance.py`). Operators who need them for tracing can flip the surfacing logger to DEBUG; the default install no longer prints query previews to stderr.

**Part 2 — DB retention.** Add `surfacing.query_retention_days` config (default 30, `0` for off). Extend `cleanup_expired` (`feedback_store.py:512`-ish) to null out the `query` column on events older than the retention window while keeping the event row for stats. Document the new column in `docs/surfacing.md` so the data lifecycle is explicit.

[...]

PR #368, opened 2026-05-20
fix(surfacing/feedback_store): null aged query text via query_retention_days (#352 part 2)

memtomem:
## Summary

- Add `SurfacingConfig.query_retention_days` (default `30`, `0` disables).
- Add `FeedbackStore.cleanup_expired_queries(retention_seconds)` that UPDATEs `query = NULL` on aged rows while preserving the row itself for stats.
- Wire it into `SurfacingEngine._maybe_cleanup_expired` as an independent sub-task alongside the existing `seen_memories` cleanup.
- Migrate legacy `query TEXT NOT NULL` DBs to the relaxed schema so the UPDATE can land.
- Document the new column + data lifecycle in `docs/surfacing.md`.

Closes part 2 of #352. Part 3 (`persist_query_text=False` opt-in hashing — write-side counterpart) ships separately. Part 1 (#367) demoted the INFO log query preview to DEBUG.

## Why

Every successful surfacing call writes one `surfacing_events` row containing the verbatim extracted query — typically file paths, the first sentence of a description argument, or an explicit `_context_query` from the agent. The legacy schema declared the column `query TEXT NOT NULL`, so the text accumulated indefinitely until the operator manually pruned `~/.memtomem/stm_feedback.db`. Combined with (a) no documentation that the DB contains query text and (b) the INFO log leak addressed by part 1, this gave a user-derived corpus that never aged out.

## Behavior change

Operators who relied on the raw `query` column being readable indefinitely lose it after 30 days by default. Set `query_retention_days=0` to keep the legacy behavior. The event row itself is never deleted by this knob — only the `query` column is cleared, so `stm_surfacing_stats` aggregates (per-tool counts, skip-reason categorization, helpfulness) are unaffected.

## Migration

`_relax_surfacing_events_query_notnull` runs from `FeedbackStore.initialize()` and detects the legacy constraint via `PRAGMA table_info`. On a hit it runs the standard SQLite recipe (CREATE new table, INSERT SELECT, DROP, ALTER RENAME, recreate `idx_events_tool`) inside `BEGIN IMMEDIATE`. Fresh DBs see no-op behavior: `CREATE TABLE IF NOT EXISTS` against an existing table is a no-op so the relaxed `_SCHEMA` only takes effect on brand-new files; the helper short-circuits when the pragma reports `notnull = 0`. `tune_connection` does not enable foreign keys, so the table swap does not trip on the `surfacing_feedback.surfacing_id REFERENCES surfacing_events(id)` declaration (it's a declared-only reference in SQLite without `PRAGMA foreign_keys=ON`).

## Independence from dedup cleanup

The previous `_maybe_cleanup_expired` guard short-circuited on `dedup_ttl_seconds <= 0` and would have suppressed retention for operators who disabled cross-session dedup. The two branches now run independently under the shared hourly interval, with separate try/except so a failure in one does not skip the other. `test_retention_runs_even_when_dedup_disabled` pins the contract.

## Test plan

- [x] `uv run pytest tests/test_feedback_store_query_retention.py tests/test_surfacing_engine.py::TestMaybeCleanupExpired tests/test_cross_session_dedup.py -x -q` — 11 new tests, all green; existing tests still pass.
- [x] `uv run pytest -m "not ollama" -q` — 2231 passed, 1 pre-existing flake (`test_search_returns_empty_when_no_session` anyio task-scope race, fails on `main` too).
- [x] `uv run pytest tests/test_docs_sync.py -q` — docs sync unaffected.


memtomem:
Force-pushed an amended commit (`ecbce41` → `3e088d3`) addressing the review concern: `FeedbackStore.get_stats` previously did `len(query)` unconditionally, which would crash on a NULL row after retention had swept anything. The recent-block loop now renders `query_preview = "<expired>"` when `query is None` instead. Two new regression tests:

- `test_get_stats_renders_placeholder_for_nulled_query` — mixed raw + nulled rows in the same recent window.
- `test_get_stats_does_not_crash_on_all_nulled` — every row swept, `get_stats` still returns a well-formed dict.

The smoke-test pass on the prior commit was misleading: the cleanup ran in the same process before any caller SELECTed a nulled row in the same recent window, so the gap only surfaced ~30 days post-merge in a real deployment. Same-PR fix is the right shape (the new invariant "`query` may be NULL" was introduced here).

Test counts updated in the commit message + CHANGELOG (11 → 13).

PR #369, opened 2026-05-20
fix(surfacing): opt-in sha256 hashing of persisted query column (#352 part 3)

memtomem:
## Summary

- Add `SurfacingConfig.persist_query_text` (default `True`, backward compatible).
- When set to `False`, the engine substitutes `"sha256:" + sha256(query)[:16]` (23 chars total) before calling `FeedbackTracker.record_surfacing` — both cache-miss and cache-hit call sites apply the substitution.
- `FeedbackStore.get_stats` passes the 23-char hash form through verbatim (no 80-char clip — the digest tail is the disambiguating bit), legacy clip preserved for raw text.
- `stm_surfacing_stats` emits a one-line legend when any recent row carries the prefix.
- Document the new knob in `docs/surfacing.md`.

Closes part 3 of #352. Part 1 (#367) demoted the INFO log query preview to DEBUG; part 2 (#368) added the `query_retention_days` TTL sweep. With this PR, **#352 is closed**.

## Why opt-in (not default-on)

Default-on hashing would be a non-trivial behavior change for operators who have built around the raw `query` column (e.g. ad-hoc SQL like `WHERE query LIKE '%search_issues%'`). Part 2's bounded TTL keeps the legacy contract with privacy bounded by time; this part adds the orthogonal full-opt-out path for deployments that need it. Operators pick the layer matching their threat model.

## In-process vs persisted

Only the persisted column is touched. The in-process surface call keeps the raw query for:

- `RelevanceGate` cooldown — Jaccard similarity against recent queries.
- The MCP `mem_search` request — hashing here would defeat surfacing entirely.
- `SurfacingFormatter` rendering — agent-visible block continues to show the raw query.

## Cache-hit parallel path

The engine has two `record_surfacing` call sites — `_render_cached` (cache hit) and `_do_surface` (cache miss). Both apply `_persistable_query`. Without the parallel cache-hit branch, every repeat query under the same surfacing cache TTL would leak the raw text. `test_cache_hit_also_hashes` pins the contract.

## Test plan

- [x] `uv run pytest tests/test_surfacing_persist_query_text.py -x -q` — 9 new tests across engine boundary, store stats, and formatter legend.
- [x] `uv run pytest tests/test_surfacing_engine.py tests/test_feedback.py tests/test_server_tools.py tests/test_observability.py tests/test_docs_sync.py -q` — 226 tests, all green.
- [x] `uv run pytest -m "not ollama" -q` — 2229 passed, 1 pre-existing flake (`test_search_returns_empty_when_no_session` anyio task-scope race, fails on `main` too).


memtomem:
Pushed a follow-up commit (`85b3349`) addressing both review findings on `feedback_store.py:396-397`. The bypass-clipping branch and the formatter legend both used `query.startswith("sha256:")`, which would misclassify a legitimate raw query starting with the literal prefix (e.g. an agent searching for a checksum) — under the default `persist_query_text=True` such a row would have skipped the 80-char clip and triggered the legend.

Both gates now use `re.fullmatch(r"sha256:[0-9a-f]{16}")` against the exact 23-char shape the engine writes:

- `_HASHED_QUERY_RE` in `feedback_store.py` — used inside `get_stats` to decide pass-through-vs-clip.
- `_HASHED_QUERY_PREVIEW_RE` in `server.py` — used inside `stm_surfacing_stats` to decide whether to emit the legend.

Kept separate because each module reads the pattern for a different purpose and the alternative import direction (`server → feedback_store` for the constant) would cycle through the surfacing engine.

Three new regression tests:

- `test_raw_query_with_sha256_prefix_is_clipped` — 200-char adversarial raw query starting with `sha256:` is clipped to 80 chars by `get_stats` and does not match the legitimate-hash regex.
- `test_invalid_hash_shape_is_clipped` — 15-hex, 17-hex, and non-hex variants all take the raw-text branch.
- `test_no_legend_for_raw_sha256_prefix_lookalike` — an 80-char-clipped lookalike does not trigger the formatter legend.

CHANGELOG updated to call out the `re.fullmatch` exact-shape gate.

Issue #353, opened 2026-05-19
Feedback model upgrade: batched per-memory ratings, finer rating granularity, LTM demotion signal


pandas-studio:
## Background

The current feedback surface (`stm_surfacing_feedback`, `src/memtomem_stm/server.py:633-659`) accepts one rating at a time from a 3-value enum (`helpful`, `not_relevant`, `already_known`). Three structural limits on the loop:

### 1. One call per (surfacing, memory) pair

`stm_surfacing_feedback(surfacing_id, rating, memory_id?)` takes a single rating. If a surfacing event returned 3 memories and the agent wants to rate them differently, it must issue 3 separate tool calls. Agents commonly drop late calls or stop mid-batch.

### 2. Bimodal rating range

3 values force a coarse signal. The agent either says `helpful` or `not_relevant`; there is no `partially_helpful` / `weakly_related` middle. Since AutoTuner (`feedback.py:110-153`) reads `not_relevant_count / total`, mid-relevance memories pushed into `helpful` skew the ratio toward over-surfacing.

### 3. Negative feedback stops at STM

`handle_feedback` invalidates the in-memory cache for negative ratings (`engine.py:268-310`) but never signals demotion to LTM. The chunk's `access_count` boosts on `helpful` via `mem_do(action=\"increment_access\")` (`engine.py:285-292`), no symmetric path for negatives. Same memory passes the score cutoff in the next session and resurfaces.

## Dependencies

- This issue overlaps with the `already_known` AutoTuner classification bug (#347). Fix that first; it is a 1-line SQL change. The rest of this issue assumes #347 has landed.
- Formatter rating-hint update (#350) needs to enumerate whatever rating values this issue settles on. Coordinate so the rating set in #350's hint matches whatever ships here.

## Proposal

Three parts. Each can ship independently; sequencing matters.

**Part 1 — batched ratings.** New tool signature (additive — keep the single-call form for compat):

\`\`\`python
async def stm_surfacing_feedback(
    surfacing_id: str,
    rating: str | None = None,           # legacy single-rating path
    memory_id: str | None = None,
    ratings: list[dict] | None = None,   # new batched path: [{memory_id, rating, note?}]
)
\`\`\`

The agent issues one call per surfacing event. The engine fan-outs internally to `record_feedback` and the boost/invalidation routines. Tests assert that mixing the legacy and batched fields raises a clear validation error.

**Part 2 — rating granularity.** Extend the enum to 4 values:

[...]

Issue #350, opened 2026-05-19
Surfacing formatter: agent-facing output bundle (source basename / score / rating hint / surfacing_id position)


pandas-studio:
## Background

Four formatter UX issues in `src/memtomem_stm/surfacing/formatter.py` that collectively shape what the agent sees and how feedback loops back. Grouped because the fixes interact (header redesign, ID placement, and per-result line format all touch the same render path).

### 1. Source path collapses to basename (`formatter.py:42`)

\`\`\`python
source = str(meta.source_file.name) if meta.source_file else \"\"
\`\`\`

`notes/2026-q1/auth.md` and `archive/old/auth.md` render identically as `auth.md`. Agent loses the parent-directory signal that helps disambiguate. Namespace badge is asymmetric — only rendered when not `\"default\"` (`formatter.py:40`).

### 2. Score is exposed raw (`formatter.py:61`)

\`\`\`python
lines.append(f\"- **{source}**{ns_badge} (score={r.score:.2f}): {preview}\")
\`\`\`

With `min_score=0.03` default, scores look high relative to the cutoff but low relative to a mental 0–1 scale. The agent has no way to calibrate this and the number is not actionable — if the agent should not surface a memory below 0.03, the engine has already filtered it. Exposing the raw score risks the agent pattern-matching on it (e.g. ignoring sub-0.5 memories) at thresholds the operator did not set.

### 3. Feedback hint omits valid rating values (`formatter.py:71-74`)

\`\`\`
_Surfacing ID: abc123def456 — call `stm_surfacing_feedback` to rate_
\`\`\`

Valid ratings are `helpful`, `not_relevant`, `already_known` (`src/memtomem_stm/surfacing/feedback.py:13`), but the injected hint does not enumerate them. Agents (Claude / GPT) inconsistently follow italic markdown directives, and rating values like `\"good\"` or `\"useful\"` reach `record_feedback` and get rejected. The feedback loop is the input to AutoTuner, so dropped feedback directly degrades learning.

### 4. surfacing_id position fragility under truncation (`formatter.py:80-81`)

\`\`\`python
if max_chars and len(memory_block) > max_chars:
    memory_block = memory_block[:max_chars] + \"\\n... (memory block truncated)\"
\`\`\`

[...]

PR #370, opened 2026-05-20
feat(surfacing): batched per-memory ratings (#353 part 1)

memtomem:
## Summary

Resolves part 1 of #353. `stm_surfacing_feedback` previously accepted one `rating` per call, so a surfacing event with N memories the agent wanted to rate differently required N tool invocations — agents commonly drop late calls or stop mid-batch, and the AutoTuner only saw whatever subset survived.

- New `ratings: list[dict]` parameter — entries `{memory_id, rating}` — mutually exclusive with the legacy `rating` / `memory_id` fields. Mixing both shapes (or supplying neither) returns a clear validation error.
- `SurfacingEngine.handle_feedback_batch` fans out per entry to the same `record_feedback` / `_invalidate_cache_for_feedback` routines the single-call path uses, so `helpful` boost and `not_relevant` / `already_known` cache invalidation keep identical side-effect semantics.
- `helpful` boosts collapse to one `increment_access` RPC across the (de-duped) helpful subset so the per-event `_boosted_event_ids` guard stays enforceable when single and batched calls interleave for the same `surfacing_id`.
- Malformed entries (non-dict / missing `memory_id` / missing `rating`) fail-fast before any write; per-entry `VALID_RATINGS` rejection surfaces in the summary string without blocking surrounding entries.
- Engine-absent fallback (`feedback_tracker` only) fans out via the tracker without boost/invalidation — the engine owns those side effects.

Parts 2 (rating granularity — `partially_helpful`) and 3 (LTM demotion signal) of #353 ship separately; part 2 needs the formatter rating-hint update from #350 to enumerate whatever values land, and part 3 needs cross-repo coordination with LTM for a `decrement_access` action.

## Behavior change

The MCP signature gains an optional `ratings` parameter and defaults `rating` to `None`. Existing single-rating callers are unaffected — `rating` is still accepted positionally / by keyword, and the dispatch through `SurfacingEngine.handle_feedback` is unchanged. Clients that previously sent `rating=null` over MCP (relying on the prior required-string contract to reject) now get a clearer `Error: either rating or ratings is required.` message instead of a schema-level rejection.

## Test plan

- [x] 7 new server-level tests (`TestSurfacingFeedback`): batched dispatch, mutual-exclusion validation (both directions), no-rating/no-ratings rejection, engine-absent batched fan-out, engine-absent malformed-entry short-circuit
- [x] 9 new engine-level tests (`TestHandleFeedbackBatch`): mixed-rating fan-out + collapsed boost, multi-helpful dedup → single boost, cross-shape guard sharing, empty-list rejection, malformed-entry no-partial-write, per-entry invalid-rating partial recovery, disabled-tracker message
- [x] Full suite: `uv run pytest -m "not ollama"` → 2260 passed (the one pre-existing unrelated `test_search_returns_empty_when_no_session` flake reproduces on main as well)
- [x] `uv run ruff check src tests` + `uv run ruff format --check` on touched files clean


memtomem:
Thanks — good catch. Fixed in 1eaa0c4.

`_record_batched_via_tracker` now mirrors `SurfacingEngine.handle_feedback_batch`: a two-pass loop that shape-validates every entry into a `parsed` list before any `record_feedback` call, so a malformed entry mid-batch fail-fasts without committing the prefix.

Regression pin added — `test_batched_fallback_malformed_after_valid_persists_nothing` — encodes exactly your scenario (`[{valid}, {missing rating}]`) and asserts `record_feedback` was never called.

PR #378, opened 2026-05-25
feat(cli): add `mms hook` to surface LTM memories for built-in tools

memtomem:
## What

Adds a `mms hook` subcommand that bridges a host's **built-in** editor tools into STM's surfacing pipeline. Hosts like Claude Code run a `PostToolUse` hook as a command that gets the tool call as JSON on stdin and replies on stdout; `mms hook` reads that payload and, for read-like built-in tools (`Read`/`Grep`/`Glob`/`Bash`), runs STM's existing surfacing over the tool output and returns relevant LTM memories via `additionalContext`.

This extends STM's value beyond MCP-proxied tools — until now, a host's built-in tool calls bypassed STM entirely.

## Why surfacing-only (no `updatedToolOutput`)

Appending context via `additionalContext` is the side-effect-free hook play. Compression-style output *replacement* is a separate, riskier stage: the replacement value must match each tool's output shape, and compressing a `Read` would break a later `Edit`'s verbatim `old_string` match. That work is deliberately deferred.

## Safety (a hook must never disrupt the host)

- **Read-like allowlist enforced in the hook itself** (`Read`/`Grep`/`Glob`/`Bash`, overridable via `MEMTOMEM_STM_HOOK_SURFACE_TOOLS`) — not relying on the host's matcher. A too-broad matcher therefore can't feed `Write`/`Edit` content in as a search query. The surfacing gate's own write-tool block can't be relied on here because `fnmatch` is case-normalizing (identity on POSIX), so `*write*` never matches `Write`.
- **No feedback store on this path** — extracted queries (e.g. secret-bearing `Bash` commands) are never persisted, and there's no SQLite contention with the server. The feedback loop, cross-session dedup, and an explicit privacy gate arrive with the planned daemon stage.
- **Never raises, always exits 0** — malformed input, disabled surfacing, timeout, or any internal error degrades to `{}` so the original tool result passes through untouched. The "never raises" contract holds for direct callers (a future daemon), not just the CLI's outer guard.

## Known limitation

A fresh `mms hook` process cold-starts the LTM MCP child each call (~6s embedding/reranker load, measured locally) which exceeds the default `surfacing.timeout_seconds` of 3.0 — so surfacing is **near-no-op with stock defaults**. Raise `MEMTOMEM_STM_SURFACING__TIMEOUT_SECONDS` and lower `MEMTOMEM_STM_SURFACING__MIN_RESPONSE_CHARS` to make it fire pre-daemon. A persistent local daemon (warm LTM connection) is the real fix and the planned next stage.

## Register in Claude Code

```json
{"hooks": {"PostToolUse": [
  {"matcher": "Read|Grep|Glob|Bash",
   "hooks": [{"type": "command", "command": "mms hook"}]}]}}
```

## Testing

- New `tests/cli/test_hook_cmd.py` (23 tests): pure helpers (`_tool_response_to_text`, `_extract_surfaced_block`); engine-seam end-to-end with a mock LTM adapter; allowlist rejection of `Write`/`Edit`/`MultiEdit`; allowlist env override; "never raises" on engine error; CLI degradation (empty / invalid JSON / non-PostToolUse / disabled → `{}`).
- `ruff check` / `ruff format` clean; `mypy` clean.
- Full suite: 2275 passed (2 pre-existing local-only failures from a live LTM on the dev box are unrelated and pass in CI).

### DIFF
--- a/src/memtomem_stm/cli/hook_cmd.py
+++ b/src/memtomem_stm/cli/hook_cmd.py
@@ -1,0 +1,288 @@
+"""``mms hook`` — bridge a host's built-in tool calls into STM's surfacing.
+
+Claude Code (and compatible hosts) invoke a ``PostToolUse`` hook as a plain
+command: the hook payload arrives as JSON on **stdin** and the hook replies
+with JSON on **stdout**. This subcommand reads that payload, runs STM's
+proactive memory surfacing over the built-in tool's output, and — when
+relevant LTM memories are found — returns them through ``additionalContext``
+so the host appends them next to the tool result the model reads.
+
+Why surfacing-only (no ``updatedToolOutput``): appending context is the
+side-effect-free hook play. Compression-style *replacement* of a built-in
+tool's output is a separate, riskier stage (it must match each tool's output
+shape, and compressing ``Read`` breaks a later ``Edit``'s verbatim
+``old_string`` match) and is intentionally out of scope here. See the Stage B
+section of the project plan / ``project_stm_builtin_tool_hooks`` memory.
+
+Tool scope: we only surface for **read-like** built-in tools
+(:data:`_DEFAULT_SURFACE_TOOLS` — Read/Grep/Glob/Bash, overridable via
+``MEMTOMEM_STM_HOOK_SURFACE_TOOLS``). This allowlist is enforced here rather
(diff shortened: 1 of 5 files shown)


pandas-studio (approved):
LGTM

PR #379, opened 2026-05-25
feat(daemon): warm surfacing daemon for `mms hook` (Stage 2, PR 1)

memtomem:
## Why

`mms hook` (#378) surfaces LTM memories for built-in tools, but spawns a fresh
LTM MCP child **per call** — the ~6s cold start (embedding + reranker load)
exceeds the 3s surfacing timeout, so with stock defaults surfacing usually
times out and the hook is a near-no-op. The fix the code already pointed at:
a persistent local daemon holding a **warm** LTM connection.

This is **Stage 2, PR 1** — the transport + foreground daemon + the warm
engine, opt-in. Auto-spawn from the hook and ownership hardening land in PR 2.

## What

**Opt-in, additive.** With `MEMTOMEM_STM_HOOK__USE_DAEMON` unset, `mms hook`
behaves exactly as today (cold in-process path). When set, it does one bounded
loopback round trip to the daemon and degrades cleanly — `{}` (default) or the
cold path (`MEMTOMEM_STM_HOOK__FALLBACK=cold`) — if the daemon is unreachable.
The hook still always exits 0.

- **Transport** (`daemon/protocol.py`, `discovery.py`, `client.py`): loopback
  TCP (`127.0.0.1`, OS-assigned port) + per-start random token, newline-JSON
  framing, `0o600` handshake file. One code path across macOS/Linux/Windows
  (Windows is a required CI target → no UDS/named-pipe split). The token gates
  *use* of the port; it's checked with `secrets.compare_digest` and never
  logged.
- **Warm engine, dedup-only** (`daemon/server.py`): one long-lived
  `SurfacingEngine` + lazily-started LTM adapter. A `FeedbackTracker` is wired
  **only** for cross-session `seen_memories` dedup — a new
  `SurfacingEngine(record_feedback_events=...)` flag (default `True`, proxy
  path unchanged) lets the daemon keep dedup while persisting **no** query text
  and emitting **no** `stm_surfacing_feedback` rating prompt (the pure-hook
  path has no in-band feedback channel; `Bash` queries may carry secrets). The
  daemon also disables the result cache (so `_surfaced_ids`/`seen_memories`
  dedup is authoritative, not short-circuited by a cache hit) and AutoTuner
  (so ranking isn't nudged by feedback rows it never records), and forces
  `persist_query_text=False`.
- **CLI** (`cli/daemon_cmd.py`): `mms daemon run/start/stop/status/restart`
  (detached spawn, `--json`), registered in `proxy.py`. Adds
  `python -m memtomem_stm` for a path-free detached spawn.
- **Lifecycle**: idle shutdown, graceful teardown that terminates the warm LTM
  child, cross-platform spawn lock + signal handling. The handshake carries a
  broadened config fingerprint so a daemon started under different config is
  treated as stale.
- **Config**: new `HookConfig` (`MEMTOMEM_STM_HOOK__use_daemon` /
  `daemon_timeout_seconds` / `fallback` / `record_feedback_events`) and
  `DaemonConfig` (`MEMTOMEM_STM_DAEMON__host` / `idle_timeout_seconds`).

## Verification

- `ruff check src` + `ruff format --check src` + `mypy src` — clean.
- `pytest -m "not ollama"` — passes (the 2 unrelated failures are the
  documented live-LTM dev-box flakes; they fail identically on `main`).
- New tests: engine dedup/feedback-event decoupling, daemon protocol round
  trip + token rejection, discovery/locking primitives, idle shutdown, dedup
  -only wiring, broadened fingerprint, single-deadline client timeout, and the
  hook degradation ladder.
- Manual: `mms daemon start` → `status --json` (running, `ltm=cold`, `0o600`
  handshake) → `stop` (clean teardown, handshake removed).

## Deferred (PR 2+)

Auto-spawn-on-first-hook (lock-guarded detached spawn + readiness poll),
lock-held-for-lifetime ownership, optional POSIX UDS fast-path. Behavioral
feedback / rating loop remain out of scope (no in-band channel in the
pure-hook path).

### DIFF
--- a/src/memtomem_stm/__main__.py
+++ b/src/memtomem_stm/__main__.py
@@ -1,0 +1,15 @@
+"""``python -m memtomem_stm`` → the Click CLI.
+
+Primarily so the daemon can re-spawn itself detached with a stable, path-free
+invocation (``[sys.executable, "-m", "memtomem_stm", "daemon", "run"]``) on
+every platform, rather than hunting for a console-script path. Because the CLI
+group only dispatches the bare-invocation MCP-server path when *no* subcommand
+is given, passing ``daemon run`` here invokes that subcommand directly.
+"""
+
+from __future__ import annotations
+
+from memtomem_stm.cli.proxy import cli
+
+if __name__ == "__main__":
+    cli()
--- a/src/memtomem_stm/cli/daemon_cmd.py
+++ b/src/memtomem_stm/cli/daemon_cmd.py
@@ -149,7 +149,12 @@ def start_cmd() -> None:
             click.echo(_ok("daemon already running"))
             return
         _spawn_detached()
-    hs = _wait_ready(config, timeout=8.0)
+        # Hold the lock through readiness. Releasing it here would let a
+        # concurrent `start` acquire it, see ping return None (our child hasn't
+        # published its handshake yet), and spawn a *second* daemon — which,
+        # with ephemeral ports + last-writer-wins handshake, orphans an extra
+        # warm LTM process.
+        hs = _wait_ready(config, timeout=8.0)
     if hs is not None:
         click.echo(_ok(f"daemon started (pid={hs.get('pid')} port={hs.get('port')})"))
     else:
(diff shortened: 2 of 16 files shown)


pandas-studio (approved):
LGTM

PR #381, opened 2026-05-25
feat(hook): default daemon mode on (use_daemon=True) (Stage 2, PR 3)

memtomem:
## What & why

Stage 2 PRs #379 (warm surfacing daemon) and #380 (lock-guarded auto-spawn + lifetime-lock) built everything needed to make daemon mode safe on by default. This PR flips the switch: `mms hook` now routes built-in-tool surfacing through the warm daemon by default.

Previously `use_daemon` was opt-in (`False`), so the default path spawned a fresh LTM child per call — a ~6s embedding+reranker cold start that exceeds the 3s surfacing timeout, meaning the default effectively surfaced nothing. With the daemon as the default, the first read-like tool call returns `{}` while fire-and-forget spawning the daemon, and subsequent calls are sub-second warm hits (idle-timing-out after 15 min).

## The non-obvious part: two sources of truth

`use_daemon` is read in two places:
- `HookConfig.use_daemon` (Pydantic field) — only `mms daemon status` reads this.
- `_daemon_enabled()` (the hook hot path) — reads `MEMTOMEM_STM_HOOK__USE_DAEMON` from the env **directly** (to skip a config load), and returned `True` only for explicit truthy values.

So flipping the field default alone would be **inert** for runtime behavior. This PR inverts `_daemon_enabled()` so unset → on, using a falsy-token set (`0/false/f/no/n/off`) that **matches Pydantic's bool parsing** — so the hook and `mms daemon status` never disagree on a value Pydantic accepts. (A Codex review caught an earlier draft that missed `n`/`f` and would have silently split the two.)

`fallback=skip` and `auto_spawn=True` are unchanged — both correct under default-on.

## Tests

- **Hermeticity fixture** (`tests/cli/test_hook_cmd.py`): with daemon-on, any `mms hook` invocation without a live daemon would read the real `~/.memtomem` handshake and `Popen` a detached daemon. An autouse fixture isolates `DATA_DIR` to tmp and neutralizes `request_spawn`, keeping CLI tests off the dev box's daemon and out of CI subprocess leaks. (Spawn behavior itself stays covered in `tests/daemon/test_server.py`.)
- Parsing test now pins the hook↔config agreement (Pydantic-recognized tokens only) + a separate unset→`True` test.
- New `test_hook_use_daemon_default_true`; the fingerprint-exclusion test now toggles `use_daemon` **off** the default so it still proves the field is excluded.

## Verification

End-to-end on an isolated temp data dir:
- unset → `hook will use daemon: yes`; `=0` → cold path
- first `mms hook` → `{}` + daemon auto-spawned (status `running`)
- warm `mms hook` → ~0.27s round trip (vs ~6s cold)

`ruff` + `mypy` clean; 2339 passed (the 2 failures are the pre-existing live-LTM-on-dev-box failures, reproduce with this change stashed).

### DIFF
--- a/src/memtomem_stm/cli/daemon_cmd.py
+++ b/src/memtomem_stm/cli/daemon_cmd.py
@@ -249,7 +249,7 @@ def status_cmd(as_json: bool) -> None:
         )
     else:
         click.echo("stopped")
-    hint = "yes" if use_daemon else "no (set MEMTOMEM_STM_HOOK__USE_DAEMON=1)"
+    hint = "yes" if use_daemon else "no (opted out via MEMTOMEM_STM_HOOK__USE_DAEMON=0)"
     click.echo(f"hook will use daemon: {hint}")
 
 
--- a/src/memtomem_stm/cli/daemon_cmd.py
+++ b/src/memtomem_stm/cli/daemon_cmd.py
@@ -203,7 +203,10 @@ def status_cmd(as_json: bool) -> None:
     from memtomem_stm.daemon.discovery import handshake_path, is_pid_alive, read_handshake
 
     config = _load_config()
-    use_daemon = config.hook.use_daemon
+    # The hook only reaches the daemon for eligible surfacing calls, so global
+    # surfacing-off means it never will — regardless of the use_daemon knob.
+    surfacing_on = config.surfacing.enabled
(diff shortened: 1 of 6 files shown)


pandas-studio (approved):
LGTM

PR #382, opened 2026-05-25
feat(daemon): key handshake+lock files by config fingerprint (Stage 2, PR 4)

memtomem:
## What

Key the daemon's handshake and lifetime-lock **filenames** by the config fingerprint:
`stm-daemon.json` / `stm-daemon.lock` → `stm-daemon-<fp>.json` / `stm-daemon-<fp>.lock`.

## Why

The handshake already carried a `config_fingerprint` and the client (`_live_handshake_candidate`) already rejected a fingerprint-mismatched daemon. But both files were still **single shared paths**, so config-drift coexistence was broken:

- A daemon started under **config A** holds the one shared `stm-daemon.lock` for its lifetime.
- A hook running under **config B** reads the shared handshake, sees the fingerprint mismatch, and rejects it (good) — then tries to auto-spawn its own daemon. `request_spawn` probes the **one** lock, finds it held by daemon A, and defers. **Forever.** Under the default `fallback=skip`, hook B returns `{}` on every call and never gets a daemon.
- Even if it could spawn, daemon B would publish to the **same** `stm-daemon.json`, clobbering A's handshake and breaking discovery for both.

## What changed

- `discovery.handshake_path(data_dir, fingerprint)` and `locking.lock_path(data_dir, fingerprint)` now take the fingerprint and produce per-config filenames (shared `stm-daemon` prefix, differing only by extension). Fingerprint is a 16-hex digest → always filesystem-safe.
- `DaemonServer` freezes the fingerprint once in `__init__` (`self._fingerprint`) and reuses it for the lock path, handshake path, and handshake content — so the lock held, file published, and fingerprint advertised can never disagree (and a mid-run env change can't repoint the daemon at a different file).
- `request_spawn` / `_live_handshake_candidate` / all `mms daemon` subcommands thread the current config's fingerprint through.
- **Removed** `mms daemon start`'s now-dead "a daemon with a different config is running — run `restart` to replace it" branch. Under coexistence a different-config daemon owns a *different, invisible* lock+handshake and is no longer a conflict; `start` simply reconciles toward its own config's daemon. `stop`/`status`/`restart` act only on the current config's keyed files.

## Result

Exactly **one daemon per distinct config**; daemons for different configs coexist, and a hook always reaches (or spawns) the daemon matching its own effective config.

## Tests

- New: `test_handshake_and_lock_paths_keyed_by_fingerprint`, `test_request_spawn_unblocked_by_different_config_lock`, `test_start_coexists_with_different_config_daemon`.
- Replaced the three obsolete different-config misreport tests (the behavior they guarded is gone).
- Updated all daemon test call sites to the keyed signatures.
- `ruff check` + `ruff format --check` clean; `mypy src` clean; `pytest -m "not ollama"` green except the 2 pre-existing local-LTM failures (also fail on `main`, pass in CI).

---

## Review (Codex)

One **P2** raised — *"keep legacy daemon paths discoverable"*: on upgrade, an old daemon owning the pre-#382 shared `stm-daemon.{json,lock}` becomes invisible to new fingerprint-keyed clients (`status`/`stop` see nothing, `start` spawns a duplicate, and a `idle_timeout=0` orphan can't be stopped via the CLI).

**Declined — no migration surface exists:**
- The entire daemon feature (#378–#381) is **unreleased** — it lands after the latest tag `v0.1.23`. No shipped version ever had the old shared-path daemon, so no end-user upgrade crosses that boundary; only a dev tracking `main` could, and `pkill -f "memtomem_stm daemon"` resolves it.
- Default `idle_timeout_seconds = 900.0`, so an orphaned old daemon self-terminates within 15 min anyway; old/new use different lock filenames, so they coexist without contention.
- A legacy read/stop/migration path would be dead code at release (no shipped `stm-daemon.json` format to be compatible with), so it's deliberately out of scope.

### DIFF
--- a/src/memtomem_stm/cli/daemon_cmd.py
+++ b/src/memtomem_stm/cli/daemon_cmd.py
@@ -96,22 +96,19 @@ def run_cmd(detached: bool, foreground: bool) -> None:
 
 @daemon_group.command(name="start")
 def start_cmd() -> None:
-    """Spawn the daemon detached if it isn't already running.
+    """Spawn the daemon detached if one isn't already running for this config.
 
-    The daemon self-guards via its lifetime lock, so ``start`` never holds the
-    lock — it reconciles toward a running daemon. The lock may currently be held
-    by a daemon that is *shutting down* (e.g. right after ``mms daemon stop``):
-    it won't answer ping and will soon release the lock, so we keep retrying the
-    spawn until the lock frees rather than polling ping once and giving up. A
-    *different-config* daemon stays alive and won't release — that we report.
+    Reconciles toward a running daemon *for the current config*. The daemon
+    self-guards via its (per-config) lifetime lock, so ``start`` never holds the
+    lock. The lock may briefly be held by a *same-config* daemon that is
+    *shutting down* (e.g. right after ``mms daemon stop``): it won't answer ping
+    and will soon release the lock, so we keep retrying the spawn until it frees
+    rather than polling ping once and giving up. A daemon started under a
+    *different* config owns a different lock + handshake (config-drift
+    coexistence), so it never blocks this spawn and is left untouched — there is
+    no cross-config conflict to report.
     """
     from memtomem_stm.daemon import client
-    from memtomem_stm.daemon.discovery import (
-        config_fingerprint,
-        handshake_path,
-        is_pid_alive,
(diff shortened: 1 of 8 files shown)


pandas-studio (approved):
LGTM

PR #391, opened 2026-05-30
fix(surfacing): hash secret-bearing queries on the proxy persist path

memtomem:
## Problem

The proxy/server surfacing path builds `SurfacingEngine` with the default config, so `persist_query_text` defaults `True` and the raw query is written to `surfacing_events.query` verbatim. A Bash `command` argument — or a tokenized `WebFetch` URL — can therefore land an **inline secret** (API key, bearer token, password) on disk.

The hook cold path (`record_feedback_events=False`) and the daemon path (`persist_query_text=False`) already mitigate this. Only the proxy path was exposed. `#352` shipped an opt-in hashing knob, not a secret guard.

## Fix

Add the guard inside `SurfacingEngine._persistable_query` — the single chokepoint both persist sites (`_render_cached` cache-hit + `_do_surface_miss` miss) already route through:

```python
if not self._config.persist_query_text:
    return self._hashed_query(query)
if contains_sensitive_content(query):
    return self._hashed_query(query)
return query
```

- **Detect-then-hash, all tools.** Reuses `proxy.privacy.contains_sensitive_content` (api_key / secret_key / `password=` / bearer / private-key header / JWT / AWS / `sk-…` / `ghp_…`). Tool-agnostic — secrets can ride in `WebFetch`/`Write` too, not just Bash.
- **No new config.** privacy.py's module docstring already anticipates this reuse.
- **No import cycle.** privacy.py imports only stdlib.
- **Disk-only.** Only the *persisted* value changes; the in-flight query used for cooldown, cache keys, formatting, webhooks and session dedup is untouched.
- **Clean queries persist as-is**, so feedback ratings keyed on `(tool, query)` still correlate (the 16-hex `sha256:` digest is deterministic for secret-bearing ones too).
- The digest logic is factored into `_hashed_query`, shared by the existing `persist_query_text=False` branch and the new guard.

## Tests

`tests/test_surfacing_persist_query_text.py` — `TestSecretGuard` + `TestPersistSitesRouteThroughGuard`:
- Parametrized over all seven secret-pattern families (each verified to match the real detector) driven end-to-end through `engine.surface()` with **default config** → hashed.
- Clean queries (verified not to match) → persisted literal.
- Unit assertion directly on the chokepoint.
- Source-inspection pin: both persist sites wrap the query in `_persistable_query`.

`uv run pytest -m "not ollama"` surfacing+privacy sweep: 667 passed. Ruff check + format clean.

### DIFF
--- a/src/memtomem_stm/surfacing/engine.py
+++ b/src/memtomem_stm/surfacing/engine.py
@@ -10,6 +10,7 @@ import uuid
 from typing import Any
 
 from memtomem_stm.observability.tracing import traced
+from memtomem_stm.proxy.privacy import contains_sensitive_content
 from memtomem_stm.surfacing.cache import SurfacingCache
 from memtomem_stm.surfacing.config import SurfacingConfig
 from memtomem_stm.surfacing.context_extractor import ContextExtractor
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #392, opened 2026-05-30
feat(surfacing): configurable + validated auto-tune bounds, purge pinned adjustments

memtomem:
## Problem

The AutoTuner had three gaps around its min_score bounds:

1. **Hard-coded literals.** `maybe_adjust` clamped to `0.05` (ceiling) and `0.005` (floor) inline — unvalidated, not configurable.
2. **No validation** tying the bounds to `min_score`, nor any guard against an increment large enough to jump the whole range in one step.
3. **Stale pinned adjustments.** When an operator pins a per-tool `min_score`, the engine already skips `maybe_adjust` for that tool — but `__init__` still resurrected the persisted adjustment, so `get_min_score_snapshot` / `stm_surfacing_stats` reported a value the engine never applies.

## Changes (one coupled `SurfacingConfig` + `AutoTuner` path)

**1. Configurable bounds.** New `SurfacingConfig.auto_tune_score_ceiling` (default `0.05`) and `auto_tune_score_floor` (default `0.005`); `feedback.py` clamps against these. Defaults are **unchanged** — prior calibration notes flag the ceiling as having no empirical basis, so this makes them tunable without re-tuning.

**2. `model_validator`.** Enforces `floor <= min_score <= ceiling`. A bound left at its default **widens** to admit a raised/lowered `min_score`, so no previously-valid config breaks on upgrade; an explicit bound on the wrong side of `min_score` is rejected. The increment guard rejects only a step **larger than the whole band** (a single jump across the entire range) — sub-band increments, including the existing `0.01` test default, stay valid.

**3. Pin purge.** `__init__` drops in-memory adjustments for pinned tools (**in-memory only** — the persisted row is left intact so unpinning restores the learned value) and logs the suppressed count. Bounds are logged once at construction.

## Tests

New `tests/test_surfacing_autotune_bounds.py` (self-contained):
- `TestConfigurableBounds`: ceiling/floor caps replace the literals, default bounds unchanged.
- `TestBoundsValidation`: adaptive widening (raised + lowered min_score), wrong-side bound rejected, over-band increment rejected, in-band increment accepted.
- `TestPinPurge`: pinned adjustment suppressed, unpinned kept, store left untouched (no save/delete round-trip), purge count logged.

The existing `test_surfacing_autotune.py` stays green. `uv run pytest -m "not ollama"` surfacing+feedback+config sweep: 658 passed. Ruff check + format clean.

### DIFF
--- a/src/memtomem_stm/surfacing/config.py
+++ b/src/memtomem_stm/surfacing/config.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from pathlib import Path
 from typing import Literal
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, Field, model_validator
 
 from memtomem_stm.proxy.config import MODEL_CONTEXT_WINDOWS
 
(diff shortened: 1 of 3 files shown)


pandas-studio (approved):
LGTM

PR #394, opened 2026-05-30
fix(compression): make SCHEMA_PRUNING + SKELETON query-aware

memtomem:
## What

Extends the #386 SELECTIVE / Hybrid-TOC query-awareness to the two structural compressors that still ranked purely by position. When a `context_query` is present, each one steers *what it keeps* toward the query; with no query (or no BM25 signal) both fall back to **byte-for-byte identical** output.

- **`SchemaPruningCompressor`** — for a top-level object whose keys score against the query, prune relevant subtrees gently and irrelevant ones hard (longer strings / more array items for the matching keys). The first tier keeps full default detail everywhere, so a payload that already fits is never pruned more than the uniform path. **Every key survives and key order is preserved** — mirrors `TruncateCompressor._json_key_truncate`.
- **`SkeletonCompressor`** — weight the per-section content budget by BM25 relevance (every section keeps a 60-char floor; the remainder is handed out by score) so relevant sections retain more body lines. **Section order and every heading are preserved** — mirrors `TruncateCompressor._section_aware_truncate`. The weighted total never exceeds the even-split total, so query-awareness adds no overshoot.
- **manager** — `SCHEMA_PRUNING` / `SKELETON` get explicit dispatch branches in `_apply_compression` that inject the manager's relevance scorer and forward `context_query`, exactly like `TRUNCATE` / `SELECTIVE`. The generic `get_compressor` fallback (NONE, PROGRESSIVE, EXTRACT_FIELDS) is unchanged — `NoopCompressor.compress` takes no `context_query`. This is the sole compression dispatch site (the main pipeline and the AUTO branch both route through it).

## Design notes

- **Load-bearing invariant:** with `context_query=None`, an empty/whitespace query, a no-signal query, a single-key object, or a non-object root, both compressors run the exact pre-query path. Pinned by `test_*_is_byte_identical` tests.
- **Reorder vs. reallocate:** these compressors emit a reduced *document* (not an index like the SELECTIVE TOC), so reordering content would corrupt document/key order. Instead we reallocate the keep-budget by relevance — the established in-repo pattern for these exact content types (`_section_aware_truncate`, `_json_key_truncate`).
- **`field_extract` is intentionally left position-ranked** — BM25 over short keys is a weak signal. Reopen if a user reports a dropped query-relevant key. Top-level arrays under schema-pruning are likewise deferred (no keys to weight).

## Out of scope (pre-existing, deferred)

`SchemaPruningCompressor`'s uniform final tier (`result[:max_chars] + "... (pruned)"`) both overshoots `max_chars` by the suffix length and can emit invalid JSON. This is a **pre-existing** gap (predates this change) that the manager retention guard mitigates in production; the proper fix for schema-pruning is non-trivial because it can't drop keys (the schema invariant). The query path here does **not** duplicate that handling — `_compress_weighted` returns `None` at a too-tight budget and delegates to the single uniform site, so at tight budgets the query output equals the no-query output.

## Tests

New `tests/test_query_aware_structural.py`: byte-identity (incl. single-key dict and tight-budget delegation), relevant-key/section enrichment, all-keys/all-headings survival, section-order preservation, no-waste-when-it-fits, non-object-root deferral, scorer **consultation** (recording stub), and manager forwarding + scorer injection for both strategies. The bench harness now forwards `context_query` to every query-aware compressor (no-op for current callers; removes a silent-drop trap).

Full suite: no new failures past the 2 known pre-existing py3.13 stdio-teardown flakes (`test_double_start_no_leak`, `test_search_returns_empty_when_no_session`).

### DIFF
--- a/src/memtomem_stm/proxy/compression.py
+++ b/src/memtomem_stm/proxy/compression.py
@@ -1177,12 +1177,16 @@ class SchemaPruningCompressor:
         handling (which would duplicate the uniform path's truncation).
         """
         # (relevant_max_str, irrelevant_max_str, irrelevant_max_array)
+        # irrelevant_max_array stays >= 2: _prune samples "first 2 + last 1", so
+        # max_array=1 on a 2-item array duplicates the tail and emits a bogus
+        # "(-1 items omitted)" marker (oversized, not pruned). The value lever
+        # for irrelevant keys is max_string, not the array cap.
         tiers = (
             (self._max_string, self._max_string, self._max_array),
             (self._max_string, 40, 2),
-            (self._max_string, 20, 1),
-            (40, 12, 1),
-            (20, 10, 1),
+            (self._max_string, 20, 2),
+            (40, 12, 2),
+            (20, 10, 2),
         )
         for rel_str, irr_str, irr_arr in tiers:
             pruned = {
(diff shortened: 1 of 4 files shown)


pandas-studio (approved):
LGTM

PR #395, opened 2026-05-30
fix(compression): valid-JSON, within-budget final tier for SchemaPruning

memtomem:
## Problem

`SchemaPruningCompressor`'s contract is *"JSON schema-preserving"*, but its
minimal final tier handled an overflow with `result[:max_chars] + "\n... (pruned)"`.
That **both overshot the budget by the 13-char suffix AND cut the JSON
mid-token, producing INVALID JSON**. It is not a corner case — a 200-key dict at
a routine 500-char budget already triggered it:

```
compress(<200-key dict>, max_chars=500)
-> len 513 (+13 over budget), tail: '..."key_016": "valu\n... (pruned)'  # JSONDecodeError
```

This is the same `result[:max_chars]+suffix` family `TruncateCompressor` fixed
in #384; the manager retention guard only masked it (it raises tiny budgets), so
direct/bench callers and tight final-tier budgets still hit it.

## Fix

Replace the hard slice with `_fit_minimal`, which **always emits valid JSON
within budget** and degrades gracefully instead of corrupting the output:

- **Tier 1** keeps EVERY top-level key, progressively collapsing the deepest
  nested levels to compact shape stubs (`{N keys}` / `[N items]`) so the agent
  still sees the full schema and nested detail survives in proportion to the
  budget. A new `max_depth` arg on `_prune` drives this — `None` (every existing
  caller) means full depth and is byte-identical to before.
- **Tier 2** drops whole trailing keys/items into a valid marker, only when the
  top-level keys *themselves* overflow (a wide object/array).
- **Floors** are type-preserving and always valid: `{}` / `[]` / `""` / `null`.

New contract: output is valid JSON **always**, and `len(out) <= max(4, max_chars)`
(`null` floor = 4; `{}`/`[]`/`""` = 2). Above a 4-char budget the cap always holds.

## Behavior change

At a budget too small to hold the full schema, the *all-keys* invariant now
**yields to validity + budget** (recording the loss in a marker), matching
`TruncateCompressor`'s `_truncated` precedent. Pre-fix, the output "kept" key
names only as substrings inside **invalid, over-budget** JSON — so this trades a
broken guarantee for an honest, parseable one. The existing
`test_information_loss` all-keys/nested-keys tests on `NESTED_CONFIG` pass
**unchanged** (depth-collapse keeps all top-level + nested keys at their budgets)
and now emit valid JSON where they previously relied on substrings in a corrupt
slice.

## Marker fidelity (caught in adversarial review, folded into this PR)

- the dropped-keys marker uses a **collision-safe key**, so a real top-level
  `_pruned` field is never clobbered and the omitted count stays consistent;
- the dropped-items marker counts against the **original** list length, not the
  already-sampled intermediate form (which understated it by ~30–300×);
- non-string scalar / empty-container roots at sub-token budgets now emit a
  **valid token** (`null`, or type-preserving `{}`/`[]`) instead of a mid-token slice.

## Out of scope (still deferred, distinct contracts)

`FieldExtract` / `Skeleton` / `Hybrid` share a surface-similar `result[:max_chars]`
pattern but with different contracts (looser-JSON / markdown-heading / non-JSON).
The `_fit_minimal` depth tiers are intentionally coarse — they can return under
budget at the floor; that is accepted floor behavior (valid + within budget), not
a regression.

## Test plan

New `tests/test_schema_pruning_output_invariants.py` (116 tests) pins: valid JSON
+ within `max(4, max_chars)` across object/array/scalar roots and budgets; the
keep-all-top-level-keys depth-collapse path; each depth rung; the marker-fidelity
guarantees; and the pathological-budget floors.

- `uv run pytest -m "not ollama"` → 2595 pass / 2 pre-existing py3.13 anyio
  stdio-teardown flakes (confirmed pre-existing on clean `main` via stash-test;
  unrelated to compression).
- `uv run ruff check src && uv run ruff format --check src` → clean.
- `uv run mypy src/memtomem_stm/proxy/compression.py` → clean.

### DIFF
--- a/src/memtomem_stm/proxy/compression.py
+++ b/src/memtomem_stm/proxy/compression.py
@@ -1240,7 +1240,12 @@ class SchemaPruningCompressor:
             omitted = n - min(2, n) - 1
             return head + [f"... ({omitted} items omitted)"] + tail
         if isinstance(data, str) and len(data) > max_str:
-            return data[:max_str] + "..."
+            # Never let the "..." ellipsis make the value LONGER than the
+            # original (e.g. a 2-char string at max_str=1). This keeps a very
+            # small max_str (used by the final tier to shrink values before
+            # dropping keys) from ever inflating short strings.
+            truncated = data[:max_str] + "..."
+            return truncated if len(truncated) < len(data) else data
         return data
 
     def _fit_minimal(self, data: object, max_chars: int) -> str:
@@ -1250,10 +1255,11 @@ class SchemaPruningCompressor:
         keeps every key and never drops whole items (the all-keys schema
         invariant). Degrades gracefully, validity and budget always held:
 
-        1. Keep EVERY top-level key, progressively collapsing the deepest nested
-           levels to compact shape stubs (``{N keys}`` / ``[N items]``) until the
-           object fits — so the agent still sees the full top-level schema, and
-           nested detail survives in proportion to the budget.
+        1. Keep EVERY top-level key, escalating value compression until it fits:
+           first collapse the deepest nested levels to shape stubs (``{N keys}``
+           / ``[N items]``), then shrink scalar strings — so the agent still sees
+           the full top-level schema, with nested/string detail surviving in
+           proportion to the budget.
         2. Only when the top-level keys *themselves* cannot fit (a wide object /
            array, where the key names alone exceed the budget) drop whole
            trailing keys/items into a valid marker.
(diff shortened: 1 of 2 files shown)


memtomem:
## Codex review — addressed in `80498f3`

Ran `codex review --base main`. It surfaced two P2 issues in the new `_fit_minimal` fallback; both are fixed in the follow-up commit:

- **[P2] Shrink scalar values before dropping keys.** Depth-collapse handled nested containers but left top-level strings at `max_str=10`, so a flat object of long strings dropped *every* key (`{"a":"x"*1000,"b":"y"*1000}` @ 30 → `{}`). Tier 1 now escalates `max_str` (10 → 4 → 0) after collapsing depth, with a no-expand guard in `_prune` so a tiny `max_str` never inflates already-short strings. Dropping a key is now strictly a last resort. → `{"a": "...", "b": "..."}` (both keys kept).
- **[P2] O(n²) prefix scan when dropping keys.** The wide-object drop loop re-serialized every prefix (a 20k-key object @ 500 chars stalled for seconds). Candidate length is monotonic in the kept-key count, so it now binary-searches the largest fitting prefix — O(log n) serializations (~40ms for 20k keys).

Also split scalar/empty roots out of the container path (a scalar has no keys to preserve, so it shrinks its value directly instead of being mangled to `"..."` by Tier 1). New tests cover the keep-keys-by-shrinking path, the binary-search behavior, the no-expand guard, and the corrected marker-fidelity assertions.

Full suite: 2595 pass / 2 pre-existing py3.13 anyio stdio-teardown flakes (unrelated). `ruff check` + `ruff format --check` + `mypy` clean.

pandas-studio (approved):
LGTM

PR #396, opened 2026-05-31
fix(compression): valid-JSON, budget-filling final tier for FieldExtract

memtomem:
## Background

`FieldExtractCompressor` was the deferred sibling of #395 (SchemaPruning final-tier fix). Its overflow handling — `result[:max_chars] + "\n... (truncated)"` at three sites (`_compress_json` dict + scalar, `_compress_text`) — both **overshot the budget** by the 16-char suffix and, on the JSON path, **cut JSON mid-token → invalid JSON**. Separately, a top-level array appended its omitted-count as a **trailing text marker after `json.dumps`**, so any array of >5 items was unparseable JSON regardless of budget.

The reopen trigger ("invalid output / budget breach on a real payload through this compressor") is now empirically reproduced.

## Concrete gap

```
FieldExtractCompressor().compress(<40-key config dict>, max_chars=500)
  before: len 516 (+16 over budget), json.loads -> JSONDecodeError
  after : len 481 (<= 500), valid JSON, all 40 top-level keys present
```

A naive "valid + within budget" fix is **not** enough: matching the old invalid output's content density requires *filling* the budget. On the bench fixture `s02` (paginated JSON whose budget sits barely above the retention floor) a uniform string-cap or depth-collapse drops below the ratio floor and strips QA-relevant values. So the final tier fills the budget with leading full-detail content.

## Change

- **`_take(value, budget)`** — greedy prefix-fill, **strict** (`compact len <= budget`) for containers, keeping leading full-length scalars (names/ids/roles) and truncating only the boundary string; the tail drops into a valid omitted-count marker counted against the **original** length and collision-safe against **every** original key (incl. a real `_truncated` in the dropped tail).
- **`_fit_extracted`** — a top-level dict lays down an **all-key skeleton** (preserves key structure), then **enriches keys in document order** to fill the budget; list / scalar roots call `_take` directly. Floors to `{}` / `[]` / `""` and emits a non-string scalar root **verbatim** (never corrupts `true`/`42` → `null`).
- Top-level array output uses an **in-array** omitted marker (valid JSON), not a trailing text suffix.

Dict budget utilisation is now ~98–100% (vs the 15–22% a depth-collapse cliff leaves); `s02` keeps its full QA values within `[floor, budget]`.

## Behavior change

- The JSON final tier now yields **all-keys-or-valid-marker** instead of an invalid mid-token slice; a top-level array's omitted-count is an in-array element rather than a trailing text suffix. `NESTED_CONFIG`/`API_RESPONSE` info-loss "all top-level keys visible" tests pass unchanged.
- `test_json_array_shows_first_items`'s budget was raised (a sub-100 budget could only surface `item0` via the old over-budget invalid slice).

## Tests

`tests/test_field_extract_output_invariants.py` (104 tests): valid-JSON, within-budget, budget-fill (no cliff), skeleton all-keys, scalar non-corruption, collision-safe markers — including regressions for findings surfaced by an adversarial-review workflow (list-root never floors to `[]`; non-string scalar roots never corrupt to `null`; marker never clobbers a real `_truncated` in the dropped tail). Full suite green (2 pre-existing py3.13 stdio-teardown flakes only).

## Deferred (separate PR)

Non-finite floats (`NaN`/`Infinity`) serialize to RFC-invalid bare tokens — a **pre-existing** class shared with SchemaPruning/Truncate (`json.dumps` default `allow_nan`), not introduced here. Tracked separately.

### DIFF
--- a/src/memtomem_stm/proxy/compression.py
+++ b/src/memtomem_stm/proxy/compression.py
@@ -960,24 +960,37 @@ class FieldExtractCompressor:
 
     def _compress_json(self, data: object, max_chars: int) -> str:
         if isinstance(data, dict):
-            summary = self._extract_dict(data, max_chars)
-            result = json.dumps(summary, ensure_ascii=False, indent=2)
+            preview: object = self._extract_dict(data, max_chars)
         elif isinstance(data, list):
-            preview_n = min(5, len(data))
-            preview: list[object] = []
-            for item in data[:preview_n]:
-                if isinstance(item, dict):
-                    preview.append(self._preview_dict(item))
-                else:
-                    preview.append(item)
-            result = json.dumps(preview, ensure_ascii=False, indent=2)
-            if len(data) > preview_n:
-                result += f"\n... ({len(data)} items total, showing first {preview_n})"
+            preview = self._preview_list(data)
         else:
-            result = json.dumps(data, ensure_ascii=False)
-        if len(result) > max_chars:
-            result = result[:max_chars] + "\n... (truncated)"
-        return result
+            preview = data
+        indent = 2 if isinstance(data, (dict, list)) else None
+        result = json.dumps(preview, ensure_ascii=False, indent=indent)
+        if len(result) <= max_chars:
+            return result
+        # Final tier: emit VALID JSON within budget. The old
+        # ``result[:max_chars] + "\n... (truncated)"`` overshot the budget by the
+        # suffix length AND cut the JSON mid-token (invalid output); a top-level
+        # array additionally appended a non-JSON text marker. ``_fit_extracted``
+        # re-derives a valid, within-budget form from the original data.
+        return self._fit_extracted(data, max_chars)
+
+    def _preview_list(self, data: list) -> list[object]:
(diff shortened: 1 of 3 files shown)


memtomem:
Follow-up `c75275e` addresses two P2s from a Codex review of the final-tier rewrite:

1. **Scalar-first enrich** — `_extract_dict` orders scalars before collections so they survive a tight budget, but `_fit_extracted`'s enrich walked keys in document order. For `{"big1":{…},"big2":{…},"big3":{…},"name":"Alice","role":"admin"}` the leading collections ate the budget and `name`/`role` stayed at their `""` stub. Now enriches scalars before collections (output key order unchanged) → leading scalar values survive.
2. **Bounded non-string scalar root** — a scalar whose literal exceeds `max_chars` (e.g. a 1000-digit number) was emitted verbatim, blowing the hard budget. Now degrades to a truncated *string* of the literal: valid JSON, within budget, visibly truncated (verbatim still returned when it fits).

Tests split accordingly; full suite green (2 pre-existing py3.13 stdio-teardown flakes only).

pandas-studio (approved):
LGTM

PR #397, opened 2026-05-31
fix(compression): valid, heading-preserving final tier for SkeletonCompressor

memtomem:
## Problem

`SkeletonCompressor` documents its core guarantee — *"preserves ALL headings … every heading survives"* — but its final tier was a raw `result[:max_chars]` byte slice. When the assembled skeleton overflowed the budget (the per-section content budget floors each section at 60 chars, so `sum(budgets) = 60·n` ignores `max_chars` once there are many sections) the slice **dropped trailing headings**, **cut the last heading mid-line**, and **sliced away the `(skeleton — …)` footer**. Empirically, an 80-section doc at a routine 2000-char budget already lost the footer and mid-cut a heading.

This is the last of the four sibling compressors that shared the `result[:max_chars]` overshoot defect (Truncate #384, SchemaPruning #395, FieldExtract #396 already fixed). This PR closes the family.

## Fix

Replace the slice with graceful degradation in a new `_fit_skeleton` helper, ranking **heading preservation above the metadata footer**:

- **All headings fit** (`all_headings_len <= max_chars`): keep every heading (the primary contract), refill body lines, append the footer only if it still fits.
- **Partial** (the bare heading lines overflow): keep the longest **whole-heading** prefix, reserving room for a `… (N of M sections omitted)` marker (a *constant* reserve — never the footer — so the kept-heading count is **monotonic** in the budget); append the footer too if it additionally fits.
- **Floor** (a budget below one heading + marker): the whole first heading when it fits, else a hard slice only below a single heading's width — a budget the retention ladder never produces.

Body refill is **layered** (every section's first body line before any section's second), so the skeleton keeps "heading + first content line per section" instead of letting early sections hog the slack. Reserves use their widest possible width, so `len(output) <= max(0, max_chars)` holds by construction. The fits-case build is untouched → **byte-identical** when the skeleton already fits (the existing query-aware / byte-identity tests run in that path). Negative `EmbeddingScorer` scores stay clamped to ≥ 0.

## Behavior change

- When headings overflow, the all-headings invariant now yields to budget **gracefully** (whole-heading prefix + `… (N of M sections omitted)` marker) instead of a destructive slice — matching the #395 "all-keys yields to budget" precedent.
- In the partial regime the footer is **best-effort** (dropped when it would crowd out headings), since heading preservation outranks the metadata footer.

## Tests

New `tests/test_skeleton_output_invariants.py` (127 tests) pins: `len(out) <= max(0, budget)`; no heading cut mid-line above the single-heading floor; **kept-heading count monotonic** across a full descending budget sweep; footer present (and counts accurate, verified independently) when roomy; every dropped section accounted for in the marker; first-body-line distributed before extras; byte-identity in the degraded path; and the floor/pathological budgets. Existing `test_query_aware_structural.py` / `test_information_loss.py` pass unchanged.

## Review

Two adversarial-review workflows + three `codex review` passes; all confirmed findings folded in:
- non-monotonic retention across tier boundaries → unified the reserve so a smaller budget never shows more headings;
- footer-vs-heading priority inversion → heading preservation first;
- document-order refill starving later sections → layered first-line-per-section refill;
- one suggested fix (restore a reserve-0 multi-heading floor) was **adjudicated and rejected** — simulated, it re-introduces the non-monotonicity, trading a general property for retention at pathological budgets the ladder never produces.

## Out of scope (separate follow-ups)

The identical Hybrid slice, `SelectiveCompressor`'s budget-blind TOC, and NaN/Infinity→null sanitization are tracked separately. Pre-existing `mypy` advisories at `compression.py:1186-1187` are in #396's `FieldExtractCompressor` (already on `main`), untouched here.

## Verification

`uv run pytest -m "not ollama"` → full suite green (only the 2 known pre-existing py3.13 stdio-teardown flakes); `ruff check` + `ruff format --check src` clean; bench `s05` (the only SKELETON scenario) unaffected (fits-case, byte-identical).

### DIFF
--- a/src/memtomem_stm/proxy/compression.py
+++ b/src/memtomem_stm/proxy/compression.py
@@ -1803,8 +1803,10 @@ class SkeletonCompressor:
         frees reserved bytes for *more* headings).
 
         * **All headings fit** (``all_headings_len <= max_chars``): keep every
-          heading (the primary contract), refill body lines up to the budget,
-          and append the footer only if it still fits.
+          heading (the primary contract), refill body lines up to the RAW budget
+          (the footer is NOT reserved, so it never preempts a body line — else a
+          larger budget that first admits the footer would drop content), then
+          append the footer only if it fits in whatever slack is left.
         * **Partial** (the bare heading lines overflow): keep the longest prefix
           of WHOLE heading lines — reserving room for the ``... (N of M sections
           omitted)`` marker but NOT the footer — then append the marker, and the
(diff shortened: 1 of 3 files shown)


pandas-studio (approved):
LGTM

PR #399, opened 2026-05-31
fix(surfacing): render disambiguating source paths

### LINKED ISSUE #359: Surfacing: render 2-segment source path to disambiguate same-basename memories
## Background

`SurfacingFormatter.inject` (`src/memtomem_stm/surfacing/formatter.py:66`) renders the source as:

```python
source = str(meta.source_file.name) if meta.source_file else ""
```

So `notes/2026-q1/auth.md` and `archive/old/auth.md` both render as `auth.md` in the surfaced-memories block. When two surfaced memories share a basename, the agent loses the parent-directory signal that would let it disambiguate which one to act on.

This was part 1 of #350. Parts 3+4 shipped via #357; CHANGELOG (line 16) explicitly deferred parts 1 and 2 as separate wait-for-signal items.

## Reopen trigger

**User report (or QA scenario) where the agent acted on the wrong memory because two surfaced results shared a basename.** Until we have such a signal, the basename collapse is theoretical — changing the rendered source string is a user-visible behavior change that touches every surfaced memory, so we want the change to be paid back by a real disambiguation problem.

Adjacent low-cost signals that would also count:
- An operator opens an issue saying their surfaced output is ambiguous because of basename collisions.
- A `bench_qa` scenario where the judge marks the agent's choice wrong because of basename ambiguity.

## Proposal (when triggered)

Render the last 2 path segments instead of only the basename:

- `Path("/notes/2026-q1/auth.md")` → `2026-q1/auth.md`
- `Path("/auth.md")` → `auth.md` (top-level file, only 1 segment available)

Plus the asymmetric namespace badge cleanup from the original #350 proposal: drop `[default]` only when the operator-configured `default_namespace` matches `meta.namespace`, instead of the current hardcoded `"default"` magic string at `formatter.py:65`. With `default_namespace=None` (current default), the badge would always render — this is a behavior change for default deployments and should land in the same PR so the source-line format stabilises in one step.

## Test pin (when implemented)

- Multi-segment path renders as `parent/basename`.
- Top-level path (1 segment) renders as just the basename.
- `default_namespace=None` + `namespace="default"` → `[default]` badge visible.
- `default_namespace="work"` + `namespace="work"` → no badge.
- `default_namespace="work"` + `namespace="default"` → `[default]` badge visible.

## Out of scope

- Score abstraction — split into a separate follow-up with a different trigger (sub-X raw score ignored signal).
- Per-result soft budget for the bullet list.

Follow-up to #350 / #357.

memtomem:
## Summary
- render surfaced memory source labels as parent/basename when possible
- honor SurfacingConfig.default_namespace when deciding whether to hide namespace badges
- add formatter regression tests and changelog coverage

Closes #359

## Tests
- uv run pytest tests/test_formatter.py tests/test_progressive.py tests/cli/test_hook_cmd.py tests/test_surfacing_engine.py -q
- uv run ruff check src/memtomem_stm/surfacing/formatter.py tests/test_formatter.py
- git diff --check


pandas-studio (approved):
LGTM

PR #400, opened 2026-05-31
fix(surfacing): render relevance buckets

### LINKED ISSUE #360: Surfacing: replace raw score with relevance bucket label after sub-X-ignored signal
## Background

`SurfacingFormatter.inject` (`src/memtomem_stm/surfacing/formatter.py:86`) renders each result as:

```python
lines.append(f"- **{source}**{ns_badge} (score={r.score:.2f}): {preview}")
```

The raw float is exposed to the agent. With `min_score=0.03` default, a score of `0.18` is already well above the cutoff the engine enforces, but on a mental 0–1 scale it reads as "low" — the agent has no way to calibrate against the operator-set floor. The number is also not actionable: anything below `min_score` was filtered before reaching the formatter, so the agent has no decision to make at that boundary.

The concrete risk is that agents pattern-match on the raw float (e.g. "ignore anything < 0.5") at thresholds the operator never set, silently muting surfacings the engine and AutoTuner already approved.

This was part 2 of #350. Parts 3+4 shipped via #357; CHANGELOG (line 16) explicitly deferred this part as wait-for-signal, noting it "needs measured 'agent ignores sub-X scores' signal before changing the score display contract."

## Reopen trigger

**Measured evidence that an agent is ignoring surfaced memories whose raw score is below some self-imposed threshold.** Concretely:

- A `bench_qa` scenario where the judge shows the agent failed to use a surfaced memory whose score was above `min_score` but below ~0.3.
- A user report where they see a surfaced memory being ignored and trace it to the raw score in the rendered output.
- A correlation in `stm_surfacing_feedback` data showing `not_relevant` rate jumps below a particular raw-score band that has no corresponding precision drop in the engine.

Without one of these, the change is speculative — the rendered display contract is user-facing, and reversing a bucket abstraction once agents start parsing it is more expensive than keeping the float for now.

## Proposal (when triggered)

Replace `(score={r.score:.2f})` with one of three bucket labels derived from the score's position between `min_score` and `1.0`:

- `[strong]`
- `[related]`
- `[weak]`

Operators retain access to exact scores via `stm_surfacing_stats` (per-tool breakdown already exposes raw distributions). Bucket boundaries should be tied to `min_score` (e.g. thirds of `[min_score, 1.0]`) so an operator who tightens or loosens the floor sees consistent bucket meaning without re-tuning their mental model.

## Test pin (when implemented)

- Bucket label appears on each result line; raw float (`score=`) does not.
- Score just above `min_score` → `[weak]`.
- Score in middle third → `[related]`.
- Score near 1.0 → `[strong]`.
- Bucket boundaries shift correctly when `min_score` is overridden (e.g. via per-tool config).

## Out of scope

- Source path rendering — split into a separate follow-up with a different trigger (same-basename confusion signal).

Follow-up to #350 / #357.

memtomem:
## Summary
- replace agent-visible raw surfacing scores with [weak], [related], and [strong] buckets
- derive bucket boundaries from the active min_score floor
- update formatter tests, bench injected-memory counting, docs, and changelog

Fixes #360

## Tests
- uv run pytest tests/test_formatter.py -q
- uv run pytest tests/test_bench_pipeline.py -q
- uv run pytest tests/test_docs_sync.py -q
- uv run pytest tests/test_surfacing_engine.py -q
- uv run pytest tests/test_stm_integration.py tests/test_remote_ltm_integration.py -q
- uv run ruff check src/memtomem_stm/surfacing/formatter.py tests/test_formatter.py tests/bench/harness.py
- git diff --check


pandas-studio (approved):
LGTM

PR #401, opened 2026-05-31
fix(compression): sanitize NaN/Infinity to null so tier output is valid JSON

memtomem:
## Problem

Python's `json.loads` accepts the JSON-extension tokens `NaN` / `Infinity` / `-Infinity` (turning them into `float('nan')` / `float('inf')`), and `json.dumps` re-emits those same bareword tokens — which RFC 8259 JSON forbids and strict parsers (browser `JSON.parse`, Go `encoding/json`, `json.loads(parse_constant=...)`) reject. Upstream tool responses that dump numpy/pandas/ML metrics routinely carry them, so any compression tier that parsed such a payload and re-serialized produced **invalid JSON** to the MCP client.

The existing tier tests missed this because plain `json.loads` silently accepts `NaN`/`Infinity`.

Reproduced before the fix (budget-exceeding payloads, strict parse):

```
FieldExtract  -> INVALID: NaN token survives
SchemaPruning -> INVALID: NaN token survives
Truncate      -> INVALID: NaN token survives (config-dict path)
```

## Fix

Sanitize at the single ingest point: a recursive `_sanitize_nonfinite` that maps non-finite floats to `null`, applied via `_mm_json_loads` (a thin `json.loads` wrapper) at every JSON parse site in `compression.py`.

Sanitizing **once at parse time** means every downstream serialization — including the budget-search probe loops that measure `len(json.dumps(candidate))` — sees only finite values, so there is **zero per-serialization overhead** (the alternative, wrapping all 27 `json.dumps` sites, would add a traversal to every probe). `_sanitize_nonfinite` returns the input unchanged (same identity) when no non-finite float is present, so the common case allocates nothing; `bool` (an `int` subclass) is left untouched.

`allow_nan=False` is deliberately **not** used: it makes `json.dumps` raise mid-search, which would break the tiers that serialize speculative candidates.

## Behavior change

Compressed JSON that previously contained `NaN`/`Infinity`/`-Infinity` tokens now contains `null` in those positions. This only changes output bytes when the upstream payload carried non-finite floats; ordinary payloads are byte-identical (the no-copy fast path).

## Tests

New `tests/test_compression_nan_inf.py` asserts strict validity via `json.loads(parse_constant=...)` across the FieldExtract / SchemaPruning / Truncate (config-dict path) / Selective tiers — with non-finite values at top level, nested in dicts, and in list elements — plus direct unit coverage of the sanitizer (non-finite → null, finite/bool/str untouched, container recursion, no-copy fast path, and the parse-time wrapper). A `test_strict_loads_helper_rejects_naninf` guard ensures the strict-parse helper itself rejects the tokens, so the suite can't silently no-op.

All compression suites pass (341 tests); `ruff check` + `ruff format` clean.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


pandas-studio (approved):
LGTM

PR #402, opened 2026-05-31
fix(compression): JSON-safe Hybrid overshoot fallback instead of raw slice

memtomem:
## Problem

`HybridCompressor.compress` assembled `head + separator + tail_compressed` and, on overshoot, did a raw `result[:max_chars]` slice. In the default TOC tail mode the tail is a `json.dumps`'d `SelectiveCompressor` envelope, so the slice commonly severed the TOC mid-object and emitted **invalid JSON** — the same raw-slice bug class the final-tier family (#394/#395/#396) removed from every other strategy.

The overshoot is reachable on any multi-section document because the `SelectiveCompressor` TOC envelope is budget-blind and can far exceed its requested tail budget. Reproduced: a 40-section doc at `max_chars=600` produced a 7555-char assembled result, sliced to 600 chars mid-object → `JSONDecodeError`.

## Fix

Fall back to a whole-text `TruncateCompressor` pass (mirroring the budget guards already in `compress`) so the output is structurally sound rather than a half-cut envelope. `TruncateCompressor`'s section-aware markdown path can itself overshoot by a few chars, so the fallback is clamped with a final codepoint slice: it carries no JSON contract (the TOC is abandoned), and for JSON / plain-text input the fallback already fits so the clamp is a no-op. The JSON-config `TruncateCompressor` path was verified within budget across 144 adversarial cases, so the clamp never corrupts JSON.

**Not** done: reshaping the Selective TOC envelope from inside Hybrid. It is a control structure bound to a stored `PendingSelection` (`selection_key` / `hint`); re-fitting its entries here would desync it from the stored chunks and break the two-phase `select()` round-trip.

The sibling raw slice in `SkeletonCompressor` (markdown, not JSON) is a separate strategy and is left for a separate PR.

## Behavior change

On TOC-mode overshoot, output is now a within-budget whole-text truncation instead of a budget-capped (but invalid-JSON) sliced TOC. When the TOC fits, it is preserved unchanged.

## Tests

`test_tail_contains_toc_or_truncation` previously passed only because the buggy sliced output kept the `"Remaining content"` separator; it now asserts within-budget, an accepted truncation/TOC marker, and — via a `_toc_tail` helper that detects a real JSON TOC tail by its leading `{` rather than the always-present separator label — that any surviving TOC parses. A new `test_toc_overshoot_stays_valid_json_within_budget` pins the overshoot fallback (within budget, no severed JSON) across a budget sweep.

All compression suites pass (313 tests); `ruff check` + `ruff format` clean.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


pandas-studio (approved):
LGTM

PR #403, opened 2026-05-31
fix(compression): budget-aware SELECTIVE TOC previews on the standalone path

memtomem:
## What

`SelectiveCompressor._store_and_build_toc` built one entry per section with a fixed 80-char preview and **ignored `max_chars` entirely**, so a many-section document produced a TOC far larger than the requested budget on the **standalone SELECTIVE path** — only the `HybridCompressor` layer guarded the budget; the standalone path forwarded the oversized TOC unclamped.

## Fix

Thread `max_chars` through `_store_and_build_toc` and shrink the per-entry preview **uniformly** via a budget-filling binary search (`build(preview_cap)`), keeping **every entry**:

- the two-phase `select()` protocol needs the full index (every section stays addressable by key);
- the manager already exempts SELECTIVE from the ratio guard for exactly this reason.

`len(build(cap))` is non-decreasing in `cap`, so the search finds the largest cap that fits; `build(0)` (empty previews) is the floor. At extreme section counts even the zero-preview TOC may exceed the budget — **accepted by design** (entries are never dropped; the agent retrieves full content via `stm_proxy_select_chunks`).

## Tests

New `tests/test_selective_toc_budget.py` (5 tests): within-budget + valid JSON + every-entry-retained, length-monotonic-in-budget, `select()` still resolves preview-omitted sections, full-preview unchanged when it already fits, and the tight-budget over-floor trade-off.

The two budget tests target the window where shrinking actually applies — **above** the zero-preview structural floor (~5.3k for 50 entries, never dropped) and **below** the raw-doc passthrough threshold (`compress` returns raw text when `len(text) <= max_chars`).

## Verification

- `tests/test_selective_toc_budget.py` — 5/5 pass
- full suite `-m "not ollama"` — no new failures (2 pre-existing py3.13 stdio flakes unrelated to this change)
- `ruff check src tests` + `ruff format --check src` — clean


pandas-studio (approved):
LGTM

PR #405, opened 2026-05-31
fix(compression): monotone list-root final tier for FieldExtract

memtomem:
## What

`FieldExtractCompressor`'s final tier for a **list root** (`_fit_extracted`) was content **non-monotone in the budget** — a larger `max_chars` could preserve **fewer** original leaves (content 4 @ budget 62 → 2 @ 63 on a bulky-first list).

**Root cause:** `_take`'s content is non-monotone in its own budget — a larger child budget can flip a value into a longer-but-emptier `_truncated` marker (e.g. `_take({"a": 1, "b": {...}}, 37) -> {"_truncated": "..."}`). Any search that maximizes the per-child budget inherits that regression.

## Fix

Replace the list path with `_fit_monotone`, a monotone analog of `_take` that keeps the same **prefix shape** (leading items in FULL detail, one truncated boundary item, an omitted-count marker) but makes every degree of freedom monotone:

- the full prefix only grows (a full item is full at every larger budget, and growing it replaces the old partial boundary with a full element);
- the boundary is filled by **recursing** (`_fit_monotone`), whose content is monotone in its own budget, so it never collapses to an emptier marker;
- the marker counts the content-free dropped tail.

The boundary's budget is derived by **exact frame arithmetic** (one recursion, no per-cap scan and no binary search — a candidate's serialized length is not monotone in its cap), so a huge or deeply nested leading item costs `O(depth)`, not `O(budget)` or `O(budget**depth)`. `_take` is left untouched (the dict-root enrich path still uses it).

## Verification

- All **105 existing** FieldExtract output-invariant tests pass unchanged (valid JSON, within budget, prefix+marker shape, item-granularity fill).
- New tests: a contiguous budget sweep over six list shapes (preserved-leaf count never regresses), the markerless-prefix floor, the huge-boundary perf bound, and JSON-escaped-string budgeting.
- Reviewed across **5 `codex review` passes**; four found edge issues (markerless collapse, nested-input blow-up, large-boundary re-serialization, escaped-string under-counting) — all folded in; the final pass is clean.
- Perf back in the `_take` ballpark (200-dict list @5000 ≈ 1 ms; 2000-element list @10000 ≈ 6 ms; 5000-int nested boundary @2000 ≈ 5 ms).
- Full suite `-m "not ollama"`: no new failures (2 pre-existing py3.13 stdio flakes).

## Out of scope

The `_compress_json` router in front of the final tier has a **separate, pre-existing** non-monotonicity — it prefers a fixed 5-item `indent=2` `_preview_list` whenever that pretty form fits, which carries less content than the budget-filling compact final tier. Tracked as a follow-up.

### DIFF
--- a/src/memtomem_stm/proxy/compression.py
+++ b/src/memtomem_stm/proxy/compression.py
@@ -1402,25 +1402,194 @@ class FieldExtractCompressor:
                     return candidate
             return '""'
 
-        # List or string root: ``_take`` is strict (<= budget) for containers, so
-        # a single call at the hard budget fills it — no search, no cliff.
-        candidate = dump(self._take(data, max_chars))
-        if len(candidate) <= max_chars:
-            return candidate
-
-        # Even the most compact ``_take`` form overflows. Floor to the smallest
-        # valid JSON within budget.
-        if isinstance(data, dict):
-            return "{}"
         if isinstance(data, list):
-            return "[]"
+            # ``_take`` content is NON-monotone in its budget: a larger child
+            # budget can flip a value into a longer-but-emptier omitted marker
+            # (e.g. ``_take({"a": 1, "b": {...}}, 37)`` -> ``{"_truncated": ...}``),
+            # so a search that maximizes the budget handed to ``_take`` can show
+            # LESS content at a LARGER ``max_chars``. ``_fit_monotone`` is the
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #406, opened 2026-05-31
fix(compression): monotone _compress_json router for FieldExtract

memtomem:
## What

The `_compress_json` router was content **non-monotone in the budget** — the separate, pre-existing issue scoped out of #405. On overflow it returned a FIXED 5-item `indent=2` head preview (`_preview_list` / `_extract_dict`) whenever that pretty form fit, and only fell through to the budget-filling compact final tier when it did not. The pretty preview carries **fewer** original leaves than the compact tier, so a **larger** `max_chars` could show **less** content:

| root | smaller budget | larger budget |
|---|---|---|
| `list(range(50))` | 60 → **9** leaves (compact) | 61 → **5** (pretty preview) |
| `[{"id","name"} ×100]` | 246 → **15** | 247 → **10** |
| nested dict | — | regressed at two budgets |

## Fix

Route every overflow through `_fit_extracted`. The router now:
- emits the **whole value** as lossless `indent=2` pretty JSON only when it fits (max content — can never regress as the budget grows);
- hands all overflow to the final tier, which for a **list root** fills the budget **monotonically** (`_fit_monotone`, #405).

This makes **list roots fully monotone end-to-end**. The **dict** final tier still inherits `_take`'s own non-monotonicity (the dict analog of #405) — tracked as a follow-up; this PR removes only the router-level cliff for dicts (verified via the pretty-is-lossless invariant).

The lossy head-preview machinery (`_preview_list`, `_extract_dict`, `_preview_dict`) is now unreachable and removed as a **consequent cleanup**; two comments/docstrings that named it are updated. Net **−128 / +151** (mostly deletions + tests).

## Behavior changes

- Overflowing dict/list output is the **compact final tier** rather than a pretty reduced preview (more content, fewer tokens; the `indent=2` form is kept only for the whole-value lossless case).
- The preview tier's **markdown code-fence-aware** string truncation (`_extract_dict`'s ` ``` ` handling) goes away with it; oversized string values now truncate plainly via `_take`. No test/consumer depended on it; flagged here for visibility.

## Verification

- New **E4** test section sweeps budgets **contiguously** through `_compress_json` for list and flat-dict roots (no leaf regression at any step, output valid + within budget), pins the canonical pre-fix boundaries (60/61, 246/247), and asserts pretty output is **always** the full lossless value.
- All **existing** FieldExtract/compression invariants pass unchanged (`126` in the FieldExtract file).
- Full suite `-m "not ollama"`: **2908 passed**, 2 pre-existing py3.13 local stdio flakes (fail identically on `main`, green on CI), no new failures.
- `ruff check src` + `ruff format --check src` clean.

## Out of scope

The **dict** final tier's `_take`-inherited non-monotonicity (the dict analog of #405's list fix) remains — surfaces as compact-output regressions on dict roots with nested collections. Tracked as a follow-up.

### DIFF
--- a/src/memtomem_stm/proxy/compression.py
+++ b/src/memtomem_stm/proxy/compression.py
@@ -1052,125 +1052,28 @@ class FieldExtractCompressor:
         return self._compress_text(text, max_chars)
 
     def _compress_json(self, data: object, max_chars: int) -> str:
-        if isinstance(data, dict):
-            preview: object = self._extract_dict(data, max_chars)
-        elif isinstance(data, list):
-            preview = self._preview_list(data)
-        else:
-            preview = data
+        # If the WHOLE value fits, emit it losslessly — pretty (indent=2) for
+        # containers, compact for scalars. ``json.dumps(data)`` always round-trips
+        # ``data`` exactly, so this tier carries the maximum possible content and
+        # can never regress as the budget grows.
         indent = 2 if isinstance(data, (dict, list)) else None
-        result = json.dumps(preview, ensure_ascii=False, indent=indent)
+        result = json.dumps(data, ensure_ascii=False, indent=indent)
         if len(result) <= max_chars:
             return result
-        # Final tier: emit VALID JSON within budget. The old
-        # ``result[:max_chars] + "\n... (truncated)"`` overshot the budget by the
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #407, opened 2026-05-31
fix(compression): monotone dict final tier for FieldExtract

memtomem:
## What

The dict final tier (`_fit_extracted`'s skeleton+enrich path) was content **non-monotone in the budget** — the dict analog of the list fix in #405, scoped out there and in #406. Completing it makes **list AND dict roots fully content-monotone end-to-end**.

The old enrich filled the budget with a per-key greedy `_take` search, non-monotone two ways:
1. `_take`'s own content flips to a longer-but-emptier `_truncated` marker as its per-child budget grows (the #405 root cause);
2. the per-key greedy let a growing budget hand an **earlier** key more budget, **displacing a later** key.

| root | regression (preserved leaves) |
|---|---|
| `{name, cfg:{…15}, tags:[…20]}` | cfg → marker-only form displaced `tags` |
| `{A:[…15 dicts], B:{…}}` | A stole budget from B (5 → 2 at one step) |
| wide nested dicts | regressed at multiple budgets |

## Fix

`_enrich_dict_monotone` applies `_fit_monotone`'s discipline to the dict root — a growing **FULL-value prefix** plus **ONE partial boundary**, every other key kept as its stub (tail = stubs, not a marker, so the "preserve key structure" contract holds). Each value is filled by `_fit_monotone`; **only one key is ever partial**, so a larger budget never displaces another key.

- **Priority:** scalars by size first (short ids/names/flags win a tight budget), then collections in document order, then **oversized scalars last** (a huge blob is low value).
- A `_content_leaves` guard keeps a key's stub rather than trading it for an emptier/marker-only enrichment.
- `_fit_monotone`'s dict branch now ranks **oversized scalar values last**, so a huge field collapses to a stub while its siblings survive — required for the recursive nested-dict case. Re-verified clean against all #405 list invariants.

## Behavior changes

- A **fully-omitted nested dict/list collapses to the compact `{N keys}` / `[N items]` stub** instead of a marker-only `{"_truncated": …}` dict. The stub is smaller AND carries a content leaf, so the fill never regresses from stub → larger-but-emptier marker as the budget grows (monotonicity). Siblings still survive. `test_dict_fit_reserves_space_for_omission_marker` moves to a budget where ≥1 key is retainable (the retained+marker path it intends); `test_dict_fully_omitted_nested_collapses_to_stub` pins the stub behavior.
- Nested dict key order may change (oversized-scalar keys rendered last).

## Verification

- New **E5** section sweeps budgets **contiguously** through `_fit_extracted` for dict roots (no leaf regression, valid + within budget at every step), pins the cross-key no-displacement boundary and oversized-scalar-ranked-last preservation.
- Broad adversarial sweep (13 dict + list shapes, contiguous budgets): **0 regressions**; fill stays tight (≥ ~90% on wide configs).
- All existing FieldExtract/compression invariants pass (incl. #405 list E3 + #406 router E4).
- Full suite `-m "not ollama"`: **2909 passed**, 2 pre-existing py3.13 local stdio flakes only (fail identically on `main`, green on CI).
- `ruff check src` + `ruff format --check src` clean; mypy error count unchanged (7, all pre-existing).

## Closes the series

This completes the FieldExtract final-tier monotonicity work: #405 (list final tier) → #406 (`_compress_json` router) → this (dict final tier). No known FieldExtract non-monotonicity remains.

### DIFF
--- a/src/memtomem_stm/proxy/compression.py
+++ b/src/memtomem_stm/proxy/compression.py
@@ -1212,13 +1212,24 @@ class FieldExtractCompressor:
         return value  # int / float / bool / None
 
     @staticmethod
+    def _is_shape_stub(s: str) -> bool:
+        """True for a placeholder emitted by ``_stub_value`` for a dropped value —
+        ``{N keys}`` / ``[N items]`` / ``""`` — none of which preserves any
+        ORIGINAL scalar content."""
+        return s == "" or (
+            (s.startswith("{") and s.endswith(" keys}"))
+            or (s.startswith("[") and s.endswith(" items]"))
+        )
+
+    @staticmethod
     def _content_leaves(obj: object) -> int:
         """Count preserved ORIGINAL scalar leaves — the metric the monotonicity
-        contract is stated in. Omitted-count markers carry no original content and
-        are excluded: a ``_truncated`` dict member and an ``... omitted`` in-array
-        string. A shape stub (``{N keys}`` / ``[N items]`` / ``""``) is a plain
-        scalar here and counts as one, so trading it for an emptier or marker-only
-        container is correctly seen as a loss."""
+        contract is stated in. PLACEHOLDERS carry no original content and count 0:
+        omitted-count markers (a ``_truncated`` dict member, an ``... omitted``
+        in-array string), shape stubs (``{N keys}`` / ``[N items]`` / ``""``), and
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #410, opened 2026-06-01
Handle AnyIO cancel scope shutdown cleanly

memtomem:
## Summary
- Treat the AnyIO cancel-scope RuntimeError seen during MCP stdio shutdown as a clean shutdown path
- Keep all other RuntimeError failures logged and re-raised through the existing main exception barrier
- Add regression coverage for the shutdown-only warning path

## Tests
- uv run pytest tests/test_server_tools.py -q
- uv run ruff format --check src/memtomem_stm/server.py tests/test_server_tools.py
- git diff --check


pandas-studio (approved):
LGTM

PR #412, opened 2026-06-02
fix(proxy): don't cache responses that embed a transient retrieval key

memtomem:
## Problem

A progressive first-chunk and a SELECTIVE/HYBRID chunk TOC each embed a key into a **process-local** pending store, but the proxy cache stored that pre-surfacing `compressed` payload under the cache TTL (3600s). That TTL is longer than the pending store's (progressive 1800s, selective 300s) and the in-memory store does not survive a restart. So a later cache hit could return a TOC / first-chunk whose `stm_proxy_read_more` / `stm_proxy_select_chunks` key was already dead — the agent loses the response tail with no way to recover it.

(Surfaced by a codex design review of `proxy/*` + `surfacing/*`; this is the High-severity finding of that review.)

## Fix

- **Don't cache transient-key responses.** A shared `response_carries_transient_key()` detector (in `cache.py`, the persistence layer) flags a response that embeds a transient retrieval key:
  - progressive first-chunks → `PROGRESSIVE_FOOTER_TOKEN`
  - SELECTIVE/HYBRID TOCs → the **pair** `"selection_key"` **and** `"ttl_seconds_remaining"` (requiring the pair avoids misclassifying an arbitrary upstream JSON that merely has a `selection_key` field; the pair is spacing-invariant, so it matches both the spaced selective dump and the compact `HybridCompressor._fit_toc_tail` dump — whose abbreviated hint drops `stm_proxy_select_chunks(`, which is why we key on fields, not the hint).
  The single `cache.set` site in `ProxyManager` skips the store (and logs at debug) when the detector fires. The next identical call re-runs the pipeline and mints a fresh, live key.
- **Purge legacy rows once.** `ProxyCache.initialize()` deletes any pre-existing rows whose `result` embeds a transient key, using `instr(...)` (case-sensitive literal match) so the SQL predicate mirrors the Python detector exactly — no LIKE wildcard/case-folding divergence.

A false positive only costs one un-cached response (re-fetched next call), never correctness. Progressive passthroughs (`content <= chunk_size`) and truncate fallbacks mint no key, carry no marker, and stay cacheable.

## Tests

- `tests/test_proxy_cache.py`: detector unit tests (progressive footer, real SELECTIVE TOC, real **HYBRID fitted-tail** TOC, truncate fallback, plain text, `selection_key`-field-without-TOC-shape, uppercase/case-sensitivity); legacy-purge tests (transient rows purged, non-TOC/case-variant rows kept).
- `tests/test_proxy_error_paths.py::TestTransientKeyResponsesNotCached`: end-to-end — a SELECTIVE TOC response is not cached (second identical call re-runs upstream), a plain response still is.
- Full suite `pytest -m "not ollama"`: 2929 passed (the 2 unrelated local hermeticity failures are addressed in a separate PR). ruff + mypy clean.

### DIFF
--- a/src/memtomem_stm/proxy/cache.py
+++ b/src/memtomem_stm/proxy/cache.py
@@ -50,6 +50,38 @@ def _make_key(server: str, tool: str, args: dict[str, Any]) -> str:
     return hashlib.sha256(raw.encode()).hexdigest()
 
 
+# Substrings that flag a response embedding a TRANSIENT retrieval key pointing
+# into a process-local pending store. Such payloads must never be cached/served:
+# the key outlives neither a process restart nor the (shorter) pending-store TTL,
+# so a cache hit would hand the agent a dead ``stm_proxy_read_more`` /
+# ``stm_proxy_select_chunks`` key with the response tail unrecoverable.
+#   - progressive first-chunks carry ``progressive.PROGRESSIVE_FOOTER_TOKEN``.
+#   - SELECTIVE / HYBRID chunk TOCs are JSON objects carrying BOTH a
+#     ``"selection_key"`` field and a ``"ttl_seconds_remaining"`` field
+#     (``SelectiveCompressor`` in compression.py). We require the PAIR — not
+#     ``"selection_key"`` alone — so an arbitrary upstream JSON that merely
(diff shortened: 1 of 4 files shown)


pandas-studio (approved):
LGTM

PR #413, opened 2026-06-02
fix(surfacing): expose per-memory id for batched feedback (EN-2/3)

memtomem:
## Summary

Surfaced memories gave an agent no way to rate (or invalidate) an **individual** memory. The injected bullet showed only `source` / `[bucket]` / preview, and the lone feedback example was the single-call `surfacing_id`+`rating` form — so `stm_surfacing_feedback(ratings=[{memory_id, rating}])`'s `memory_id`, and the engine's per-memory cache invalidation keyed on `str(chunk.id)`, were effectively unreachable from the rendered block (EN-2/3 from the dev-trio review).

This is PR B of the STM-review series (PR A = #412). Branched from `main`.

## Changes

- **Render the `memory_id` in each bullet** — `chunk.id` is emitted as a backticked token *before* the `[bucket]:` marker, so the preview parse and `preview_max_chars` cap are unaffected, and the token matches the `str(r.chunk.id)` key the engine filters invalidated memories on.
- **Batched feedback example** — a second `stm_surfacing_feedback(..., ratings=[{"memory_id": ..., "rating": ...}])` line is added after the single-call example (gated on `surfacing_id`; the single-call line stays first so the AST-scrape contract for the one-literal `rating=` example holds).
- **Preamble-pinned, whole-line truncation** — replaces the flat `[:max_chars]` slice. The feedback preamble (header + surfacing_id + both rating examples) is always kept; the body is dropped on whole-line boundaries so a `memory_id` token is never severed mid-string (a half-copied id silently no-ops invalidation). An orphan `**Working Memory:**` header that a whole-line cut could leave behind is trimmed.
- **Defensive id rendering** — `getattr(chunk, "id", None)`; a chunk without an id degrades to the legacy id-less bullet instead of raising in the hot injection path. (Every production chunk has an id — surrogate under compact, real `chunk_id` under structured — so this only guards a degenerate case.)
- **Documented the compact surrogate-id limitation** — under the default `result_format="compact"` the id is `sha256(content)[:16]`: it drives STM-side cache invalidation but not the LTM `increment_access` boost, and identical-content memories collide on one id. `structured` carries the real `chunk_id` end to end. (config docstring, `mcp_client.py` comment, `CHANGELOG.md`, `docs/surfacing.md` sample.)

### Behavior change

The injected bullet gains a `` `id` `` segment between the namespace badge and `[bucket]:`. Parsers keying on the exact `**source**{ns} [bucket]:` shape must account for it (the preview still follows `[bucket]: `).

## Tests

- Bullet id render + placement (not in the preview slice), batched-example present/absent, id-less degradation.
- Preamble-pin + whole-line truncation, plus the preamble-overrun (zero-bullet) and orphan scratch-header boundaries.
- Engine roundtrip: lifts the rendered id off the `apple memory` bullet and feeds it through `handle_feedback_batch` → next cache hit drops that memory (proves the rendered token equals the invalidation key).
- Updated 3 relevance-bucket assertions for the new bullet shape.

Gate: `ruff check`/`format` ✅, `pytest -m "not ollama"` ✅ (the 2 unrelated `test_double_start_no_leak` / `test_search_returns_empty_when_no_session` failures are pre-existing local-only — a live LTM server on the dev box — and pass in CI; tracked for PR E), `mypy` ✅.

A 13-agent adversarial review of the diff surfaced 6 findings (3 rejected); all 6 are folded into this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased]
 
+### Changed
+
+- **Surfaced memories expose a per-memory `memory_id` for feedback** (EN-2/3) — each injected bullet now renders its `chunk.id` as a backticked token, and the feedback preamble adds a batched `stm_surfacing_feedback(ratings=[{"memory_id": ..., "rating": ...}])` example alongside the single-call one, so an agent can rate (and invalidate) individual memories rather than the whole surfacing event. **Behavior change**: the injected bullet format gains a `` `id` `` segment between the namespace badge and the `[bucket]` label; parsers keying on the exact `**source**{ns} [bucket]:` shape must account for it (the preview still follows `[bucket]: `). Injection-size truncation now pins the whole feedback preamble and drops body bullets on whole-line boundaries so a `memory_id` token is never severed mid-string. Under the default `result_format="compact"` the rendered id is a content-derived surrogate (`sha256(content)[:16]`), which drives STM-side invalidation but not the LTM `increment_access` boost; set `result_format="structured"` for the real `chunk_id` end to end.
+
 ## [0.1.25] — 2026-06-02
 
 A small maintenance release: the standalone `mms` MCP stdio server now exits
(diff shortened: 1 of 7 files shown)


pandas-studio (approved):
LGTM

PR #414, opened 2026-06-02
fix(surfacing): pad short tool-call queries with the tool name (KR-4.2)

memtomem:
## Summary

`ContextExtractor.extract_query` appended the tool name to clear `min_query_tokens` **only when argument extraction produced nothing**. A short-but-non-empty extraction — e.g. `read_file {"path": "/etc/hosts"}` tokenizing to `"etc hosts"` (2 tokens, below the default floor of 3) — returned no query, so surfacing silently never fired for it, even though the tool-name token would have cleared the floor (KR-4.2 from the dev-trio review).

This is PR C of the STM-review series (PR A = #412, PR B = #413, both merged). Branched from `main`.

## Change

The fallback now fires whenever the **extracted tokens alone** fall below `min_query_tokens`, not only when `parts` is empty:

```python
if len(" ".join(parts).split()) < config.min_query_tokens:
    parts.append(tool.replace("_", " "))
```

- The empty-parts case is subsumed (`"".split()` → 0 tokens, and `min_query_tokens` is validated `gt=0`).
- The append runs **after** the deterministic `sorted(arguments.items())` scan, so the query string — the surfacing cache / cooldown / dedup key — stays stable.
- The unchanged final threshold check still returns `None` when the combined query is still too short even with the tool name (e.g. a one-token tool).

**Trade-off** (in the commit body): a marginal query now consumes a surfacing rate-limit / cooldown slot, and a large `min_query_tokens` widens how often the tool name dilutes the args — tune it there if noisy.

## Tests

New `TestShortNonEmptyQueryPadsWithToolName`:
- `/etc/hosts` → non-`None`, contains both the arg tokens and the appended tool-name tokens.
- Deterministic append order (`"etc hosts read file"`).
- Still `None` when args + tool name remain below the floor (`io` + `/etc/hosts`, floor 5).
- A strong query (≥ floor) is **not** diluted by an appended tool name.

Existing regressions stay green: `test_returns_none_for_short_query`, `test_falls_back_to_tool_name` / `_only_short_values` / `_all_underscore`, and the uuid/hex→`None` filters in `test_surfacing_query_extraction.py`. Note `test_short_query_rejected` and `test_no_query_records_skip` were already written to anticipate this padding behavior.

Gate: `ruff check`/`format` ✅, `pytest -m "not ollama"` ✅ (the 2 unrelated `test_double_start_no_leak` / `test_search_returns_empty_when_no_session` failures are pre-existing local-only — a live LTM server on the dev box — and pass in CI; tracked for PR E), `mypy` ✅. A 3-agent adversarial review of the diff (correctness / regression-scope / tests+plan) returned **0 findings**.


pandas-studio (approved):
LGTM

PR #417, opened 2026-06-02
fix(cli): drop dead top-level min_response_chars from the --lang ko preset

memtomem:
## What

`mms init --lang ko` wrote a **top-level** `min_response_chars: 230` into the proxy config, but `ProxyConfig` has **no such field** — it exists only on the nested `ExtractionConfig` (default `500`). With pydantic's default `extra="ignore"`, the key was **silently dropped on load**: one of the four advertised KO token-aware budgets never applied (a no-op, not a crash).

## Evidence

```python
cfg = ProxyConfig.model_validate({"min_response_chars": 230, "chars_per_token": 1.85, ...})
hasattr(cfg, "min_response_chars")        # False
cfg.extraction.min_response_chars         # 500  (the 230 was dropped)
"min_response_chars" in cfg.model_dump()  # False
```

The existing CLI test (`test_lang_ko_writes_proxy_and_per_server_fields`) only asserted the **written JSON** contained the key — it never loaded through `ProxyConfig`, so it passed despite the no-op (tested the wrong layer).

Intent was ambiguous: the preset grouped it with compression budgets and `docs/configuration.md` called it a "`min_response_chars`-style gate [that] skips compression", but **no top-level/compression `min_response_chars` gate exists** — the only ones gate **extraction** (`ExtractionConfig`, 500) and **surfacing** (`SurfacingConfig`, 5000). Even nested correctly under `extraction`, the value would be inert in the standalone `mms` server (no `index_engine` wired, #288). So the right fix is to **remove** the dead key, not relocate it.

## Change

- **`cli/proxy.py`** — remove `min_response_chars` from `_LANG_PRESETS["ko"]` and from the `--lang` help text. `chars_per_token`, `default_max_result_chars`, and per-server `max_result_tokens` are unaffected (those are real, effective fields).
- **`tests/cli/test_proxy_cli.py`** — harden the KO test: round-trip the written config through `ProxyConfig.load_from_file` and assert the *effective* fields, plus a `"min_response_chars" not in data` regression guard, so a future silently-dropped preset key fails loudly instead of passing a raw-JSON-only check.
- **`docs/configuration.md`** — drop the key from the `--lang ko` preset listing and JSON example; reword the inaccurate compression-gate sentence.
- **`CHANGELOG.md`** — `Fixed` entry.

## Provenance

Surfaced while documenting the *surfacing* `min_response_chars` (KR-4.1, #416). Independently confirmed by a codex review (NEEDS-FIX → remove + harden test + sync docs/help).

## Verification

- `TestInitLangPreset` + `tests/test_docs_sync.py`: 14 passed.
- Full suite `uv run pytest -m "not ollama"`: **2944 passed, 0 failed**.
- `ruff check src` clean; edited regions format-clean.


pandas-studio (approved):
LGTM

PR #419, opened 2026-06-03
feat(cli): add `mms stats` to read proxy/surfacing stats from disk

memtomem:
## Context

STM records rich observability to `~/.memtomem/proxy_metrics.db` and
`~/.memtomem/stm_feedback.db`, but there was **no CLI surface** to read it. The
only options were calling the MCP stats tools (a live client round-trip) or
hand-writing SQL against the DB files — in a recent session that meant manually
querying SQLite to report "41 calls, 495K→204K chars, 58.7% saved". This adds a
one-liner: `mms stats`.

## What it does

```
$ mms stats
Config : ~/.memtomem/stm_proxy.json (ok)
Enabled: yes
Servers: 3

Proxy / Compression
==============================
  calls: 41  (errors: 3)
  chars: 495,495 -> 204,443  (saved 291,052, 58.7%)

  SERVER       TOOL                          CALLS       ORIG       COMP  SAVED%
  context7     query-docs                       13     48,346     44,747    7.4%
  langfuse-doc searchLangfuseDocs                6    286,057     61,389   78.5%
  ...

Surfacing
==============================
  surfaced events: 0  (distinct tools: 0)
  feedback ratings: 0
```

Flags: `--json` (scripting / future `memtomem web` contract), `--tool` (filter), `--config`.

## Design notes

- **Read-only, by a separate process.** The CLI can't see the MCP server's
  in-memory `TokenTracker`, so it aggregates from the persistent stores. The
  read path follows the existing `inspect_feedback_db` precedent: a `?mode=ro`
  open with **no** `tune_connection` (avoids a `PRAGMA journal_mode=WAL` write)
  that never runs `CREATE TABLE`/migrations — a *read* command must not mutate
  the user's DB. New `read_compression_summary` / `read_surfacing_summary`
  module helpers keep the SQL/schema knowledge in the store modules; the CLI
  only orchestrates + formats.
- **Degrades, doesn't crash.** Missing file, a foreign DB, or a pre-migration
  schema (no `is_error` column) → `available`/`schema_outdated` flags instead of
  a traceback.
- **Config precedence.** Mirrors the server's `env > file > defaults`
  (`collect_proxy_env_overrides` + `ProxyConfig.load_from_file`) and reports
  `config_status` (ok/missing/invalid); surfacing path via
  `STMConfig().surfacing.feedback_db_path` like `mms health`.
- **No query-text leak.** Surfacing output omits the `recent` query previews
  `FeedbackStore.get_stats` can surface.

## Scope

[...]

PR #420, opened 2026-06-03
fix(cli): `mms stats` must mirror the server's env-enabled config bypass

memtomem:
## Context

Follow-up to #419, surfaced by a dev-trio **codex** review of the merged `mms
stats` code (verdict NEEDS-FIX, one Major).

`mms stats` always called `ProxyConfig.load_from_file(...)` to locate
`proxy_metrics.db`. But the server bootstrap (`server.app_lifespan`) **skips the
JSON file entirely** when `MEMTOMEM_STM_PROXY__ENABLED` is set and uses the
env-only proxy config:

```python
# server.py
if not os.environ.get("MEMTOMEM_STM_PROXY__ENABLED"):
    file_cfg = ProxyConfig.load_from_file(config.proxy.config_path, ...)
    if file_cfg is not None:
        config.proxy = file_cfg
```

So in env-enabled mode a file-level `metrics.db_path` is ignored by the server
but was still honored by `mms stats` → the CLI could report a **different (often
empty) `proxy_metrics.db`** than the one the server actually writes to.

## Fix

Mirror the server's branch:

- `MEMTOMEM_STM_PROXY__ENABLED` set → use `STMConfig().proxy` (env-only), bypass
  the JSON file, report `config_status: "env"` (and a clear "JSON file bypassed"
  line in the human output).
- otherwise → unchanged `env > file > defaults` via `load_from_file`
  (ok/missing/invalid).

The surfacing path now reuses the same `STMConfig()` instance instead of a
second construction.

## Tests

New `test_env_enabled_bypasses_file_metrics_path`: with
`MEMTOMEM_STM_PROXY__ENABLED=1` and a config file pointing `metrics.db_path` at a
**seeded** DB, asserts the CLI reports `config_status: "env"` and `total_calls:
0` (i.e. it reads the env/default path, not the bypassed file's seeded DB).

## Verification

- `uv run pytest -m "not ollama" tests/cli/` → **408 passed** (+ the new case);
  stats suites green.
- `ruff check src` + `ruff format --check src` clean; `mypy src/.../cli/proxy.py`
  clean (the type-narrowing was tidied so `proxy_cfg` is never `… | None`).
- Manual smoke: `MEMTOMEM_STM_PROXY__ENABLED=1 mms stats` now prints
  "MEMTOMEM_STM_PROXY__ENABLED set — env config (JSON file bypassed)".

### DIFF
--- a/src/memtomem_stm/cli/proxy.py
+++ b/src/memtomem_stm/cli/proxy.py
@@ -681,25 +681,40 @@ def stats(config_path: str, *, tool_filter: str | None = None, as_json: bool = F
     path = Path(config_path)
     resolved = path.expanduser().resolve()
 
-    # Mirror the server's env > file > defaults precedence so the resolved DB
-    # paths honor MEMTOMEM_STM_* overrides (server.py app_lifespan).
-    proxy_cfg = ProxyConfig.load_from_file(path, env_overrides=collect_proxy_env_overrides())
-    if proxy_cfg is None:
-        # Parse/validation error — fall back to defaults purely to locate the
-        # DB paths to probe, but flag the config as invalid.
-        config_status = "invalid"
-        proxy_cfg = ProxyConfig()
-    elif not resolved.exists():
-        config_status = "missing"
-    else:
-        config_status = "ok"
-
-    # Surfacing config lives outside the proxy JSON file; read it (env-aware)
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #422, opened 2026-06-04
feat(surfacing): per-upstream surfacing toggle (surfacing_enabled + `mms surfacing`)

pandas-studio:
## Why

Surfacing scope was controllable only **globally** (`surfacing.enabled`) or **per-tool** (`context_tools`). The one per-server lever — `exclude_tools` glob, which already matches `server__tool` — lives in the **top-level env config** (`MEMTOMEM_STM_SURFACING__EXCLUDE_TOOLS`). That means every MCP client must carry it in its own `env` block, and the setting **drifts across clients**: Claude Code / Codex / Antigravity all launch the same `mms` and share one `~/.memtomem/stm_proxy.json`, but **not** one env. For an upstream whose calls never match the user's LTM (e.g. third-party doc servers), surfacing also runs a **wasted LTM search on every call**.

This adds a per-upstream toggle that lives in the **shared** proxy config: set once, hot-reloaded, visible in `mms status`, consistent across every client.

## What

- **`UpstreamServerConfig.surfacing_enabled: bool = True`** — persisted in `stm_proxy.json`.
- **`mms surfacing <server> [on|off]`** — toggle it (omit the state to print current); `mms status` now renders `surfacing=on/off` per server.
- **Enforced in `ProxyManager`** on both surfacing entry points (`_apply_surfacing` and `_apply_surfacing_on_progressive`), reading the **hot-reloaded** config via `self._config.upstream_servers` — *not* the `SurfacingEngine` relevance gate, which is built once at startup from the top-level `SurfacingConfig` and never sees per-upstream config. When disabled, surfacing is skipped **before the LTM search** (saving the round-trip) and recorded as a new **healthy** `upstream_disabled` reason in `stm_surfacing_stats`. Unknown servers fail open (surfacing stays best-effort).

## Why this enforcement point

Two independent reviews converged on it: the `SurfacingEngine` can't honor a `stm_proxy.json` field because it's instantiated once at startup; the `ProxyManager` reads the proxy config live, so it's the only place a hot-reloadable per-upstream flag can take effect.

## Docs

- `docs/surfacing.md`: new **Scoping surfacing per upstream** section (per-upstream toggle + server-qualified `exclude_tools` glob + env recipe), enriched `exclude_tools` table row, `upstream_disabled` skip-reason legend, and a **fix** to the surfacing sequence diagram — the circuit breaker is checked *before* query extraction and the gate, not after.
- `docs/cli.md`: `mms surfacing` command reference.
- `CHANGELOG.md`: Unreleased → Added.

## Testing

- New `tests/test_per_upstream_surfacing.py` (config default/round-trip; `_surfacing_enabled_for` incl. fail-open; ProxyManager short-circuit + `record_skip` on both normal and progressive paths; enabled path still surfaces).
- New `TestSurfacingCommand` in `tests/cli/test_proxy_cli.py` (on/off/show, persistence, unknown-server error, invalid-state rejection, `mms status` rendering).
- Full CI-filtered suite: **2973 passed, 3 skipped**. `ruff check`/`format` clean, `mypy` clean. CLI smoke-tested end to end.

## Notes for reviewers

- `exclude_tools` (env, glob, tool-grained / cross-server) and `surfacing_enabled` (shared config, per-upstream, persistent) are complementary; the docs steer between them.
- No behavior change at the default (`surfacing_enabled=True` everywhere).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased]
 
+### Added
+
+- **Per-upstream surfacing toggle** — `UpstreamServerConfig` gains a `surfacing_enabled` flag (default `true`), and `mms surfacing <server> [on|off]` toggles it in `stm_proxy.json` (`mms status` renders it per server). Disabling an upstream short-circuits proactive surfacing for every tool on that server *before* the LTM search runs. It is enforced in `ProxyManager`, which reads the hot-reloaded proxy config — not the `SurfacingEngine` relevance gate, which is built once at startup from the top-level `SurfacingConfig` and never sees per-upstream config — and is counted as a new healthy `upstream_disabled` reason in `stm_surfacing_stats`. Because the flag lives in the shared proxy config rather than per-client `env`, every MCP client proxying through the same `mms` sees one consistent scope, unlike the existing `MEMTOMEM_STM_SURFACING__EXCLUDE_TOOLS` glob (which also matches `server__tool` but must be carried in each client's env). `docs/surfacing.md` now documents both the per-upstream toggle and the server-qualified `exclude_tools` glob, and corrects the circuit-breaker ordering in the surfacing sequence diagram.
+
 ## [0.1.26] — 2026-06-03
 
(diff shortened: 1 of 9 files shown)


memtomem (approved):
LGTM

PR #426, opened 2026-06-10
fix(proxy): PROGRESSIVE-strategy calls crash with UnboundLocalError at the metrics record

memtomem:
## Summary

Any tool configured with `compression: progressive` crashed on every call: the Stage-2 PROGRESSIVE branch in `_call_tool_inner` never assigned `metrics_strategy`, so the `CallMetrics` record at the end of the pipeline raised `UnboundLocalError` — after the upstream call had already succeeded. The agent got an internal error instead of a response; the strategy was unusable end to end.

## Why nothing caught it

- The F6 progressive tests exercise `_apply_surfacing_on_progressive` in isolation, never `call_tool` with the strategy configured.
- The ratio-guard tests drive the progressive *fallback* tier, which lives in the compression branch where `metrics_strategy` is assigned.
- #412's transient-key integration tests cover the SELECTIVE path plus per-marker detection units; a PROGRESSIVE-strategy integration test would have hit this crash.

Found while writing regression tests for the cache/`read_more` interaction #412 fixed — the test could not even drive the branch.

## Fix

One line: mirror `metrics_strategy = effective_compression.value` at the point the PROGRESSIVE branch sets its other metrics fields, matching the assignment at the top of the compression branch.

## Behavior change

- Tools configured with PROGRESSIVE now return responses instead of erroring (the bugfix itself).
- Their metrics rows now record `compression_strategy = "progressive"` — a value that previously never appeared via this branch, so dashboards/queries grouping by strategy will start seeing it.

## Tests

- `test_progressive_strategy_call_completes_and_records_metric` — end-to-end PROGRESSIVE call returns a keyed first chunk and records `"progressive"` in the metrics row (fails with `UnboundLocalError` pre-fix).
- Extended #412's `TestTransientKeyResponsesNotCached` to the now-reachable progressive path: the first chunk is not cached and a second identical call re-runs upstream; a repeat call after progressive-store eviction mints a fresh, live `read_more` key (the original data-loss shape); single-chunk passthrough carries no key and stays cacheable.
- Full suite: 2987 passed, 3 skipped (`-m "not ollama"`).

### DIFF
--- a/src/memtomem_stm/proxy/manager.py
+++ b/src/memtomem_stm/proxy/manager.py
@@ -1589,6 +1589,11 @@ class ProxyManager:
                 )
             _compress_ms = 0.0
             compressed_chars_for_metrics = len(cleaned)
+            # The metrics record at the bottom reads ``metrics_strategy`` on
+            # every path; mirror the assignment at the top of the compression
+            # branch (pre-fix this branch left it unbound and every
+            # PROGRESSIVE-strategy call died with UnboundLocalError).
+            metrics_strategy = effective_compression.value
             # F6: surface on progressive when ``injection_mode`` is append/section.
             # ``prepend`` stays skipped (offset-shift); helper returns ``ok=None``
             # in that case and logs a one-time WARNING.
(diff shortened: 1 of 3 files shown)


pandas-studio (approved):
LGTM

PR #398, opened 2026-05-31
feat(surfacing): support network LTM MCP transport

memtomem:
## Summary
- add surfacing LTM MCP transport config for stdio, SSE, and streamable HTTP
- route McpClientSearchAdapter and mms health through the selected transport, including URL and headers for network LTM services
- document the new environment variables and cover config, adapter, and health behavior with tests

Fixes #297

## Verification
- uv run pytest tests/test_config_constraints.py tests/test_mcp_client_reconnect.py tests/test_surfacing_engine.py::TestSurfacingLtmOutcomeDispatch tests/cli/test_proxy_cli.py::TestHealth tests/test_remote_ltm_integration.py -q
- uv run ruff check src/memtomem_stm/surfacing/config.py src/memtomem_stm/surfacing/mcp_client.py src/memtomem_stm/surfacing/engine.py src/memtomem_stm/cli/proxy.py tests/test_config_constraints.py tests/test_mcp_client_reconnect.py tests/cli/test_proxy_cli.py
- git diff --check

PR #427, opened 2026-06-10
fix(surfacing): classify httpx errors as transport errors so network LTM reconnects

memtomem:
## Summary

After the first transient network error, surfacing over the #398 `sse` / `streamable_http` LTM transports is permanently dead until process restart.

`_TRANSPORT_ERRORS = (OSError, ConnectionError, EOFError, BrokenPipeError, asyncio.TimeoutError)` covers only stdio failure modes. The mcp network clients raise `httpx` errors (`ConnectError`, `ReadTimeout`, `RemoteProtocolError`, `ConnectTimeout`, ...) whose MRO — `httpx.* → TransportError → RequestError → HTTPError → Exception` — has no `OSError` ancestor (verified empirically against this repo's venv). So a network blip falls through `except self._TRANSPORT_ERRORS` into the generic `except Exception` in `search()` / `increment_access()` / `scratch_list()`: outcome `"call_error"`, no `_reconnect()`, no `_needs_reconnect`. The dead `ClientSession` is reused forever. stdio self-heals only because pipe failures *are* `OSError`/`BrokenPipeError`.

## Fix

Add `httpx.TransportError` to `_TRANSPORT_ERRORS`:

- `TransportError` is the exact transient-failure base (timeouts, network errors, protocol violations). `HTTPStatusError` / `DecodingError` are deliberately excluded — the server answered, the transport is healthy, and reconnecting would mask real application errors. Pinned by a no-reconnect test.
- httpx is imported plainly: it is a declared direct dependency (`httpx>=0.28`) and this module already imports the mcp sse/streamable_http clients unconditionally, which themselves require httpx. (An earlier revision wrapped the import in a dead try/except guard — codex R1 caught that the trimmed-env claim was unsatisfiable; dropped in `5d771ea`.)
- All three call sites share the class-level tuple, so they heal uniformly.

## Why nothing caught it

The reconnect/heal suite constructed every adapter with the default stdio config and injected `ConnectionError`/`OSError` — types that happen to be in the tuple. The network transports' only coverage was transport *selection* (`_open_transport` returns the right client), never error classification.

## Tests

New `TestNetworkTransportErrorsReconnect` in `tests/test_mcp_client_reconnect.py`:

- `test_httpx_error_triggers_reconnect_and_retry` — parametrized over `{ConnectError, ReadTimeout, RemoteProtocolError} × {sse, streamable_http}`: reconnect fires once, the retry's results are returned, outcome `"ok"`.
- `test_httpx_error_with_failed_reconnect_is_transport_error` — outcome is `"transport_error"` (reconnect path), not `"call_error"` (the pre-fix misclassification).
- `test_httpx_status_error_does_not_reconnect` — HTTP 500 stays application-level: outcome `"call_error"`, no reconnect.

Pre-fix these fail 7/8 (only the status-error guard passes); post-fix 8/8. Full suite: 2991 passed, 3 skipped (`-m "not ollama"`). `ruff` and `mypy` clean (with the plain import the tuple is a literal of exception types, so no annotation is needed).

### DIFF
--- a/src/memtomem_stm/surfacing/mcp_client.py
+++ b/src/memtomem_stm/surfacing/mcp_client.py
@@ -12,6 +12,7 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, Literal, cast
 
+import httpx
 from mcp import ClientSession
 from mcp.client.sse import sse_client
 from mcp.client.stdio import StdioServerParameters, stdio_client
@@ -20,16 +21,6 @@ from mcp.types import TextContent
 
 from memtomem_stm.surfacing.config import SurfacingConfig
 from memtomem_stm.utils.numeric import safe_float
-
-# httpx ships as a dependency of the mcp SDK (its sse/streamable_http
-# clients are built on it), but guard the import so a trimmed stdio-only
-# environment still loads this module.
-try:
-    import httpx
-
-    _HTTPX_TRANSPORT_ERRORS: tuple[type[Exception], ...] = (httpx.TransportError,)
-except ImportError:  # pragma: no cover — httpx comes with mcp
-    _HTTPX_TRANSPORT_ERRORS = ()
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #428, opened 2026-06-10
fix(server): recognize ExceptionGroup-wrapped cancel-scope errors on shutdown

memtomem:
## Summary

#410 (`6b906be`) taught `main()` to swallow the AnyIO cancel-scope RuntimeError on stdio EOF, but the guard is `except RuntimeError` — and `mcp.run()` executes the server inside `anyio.create_task_group()`, whose strict mode (anyio ≥ 4, this repo pins 4.13) wraps anything escaping the group in an `ExceptionGroup`. Verified empirically against this repo's venv: an error raised under `anyio.run()`'s task group surfaces as `ExceptionGroup`, not `RuntimeError`.

So the production shape sails past the #410 guard into the generic barrier: ERROR log + re-raise + non-zero exit on every clean shutdown — exactly the noise #410 set out to remove. The #410 test passes only because it injects a *bare* `RuntimeError` as `mcp.run`'s `side_effect`, bypassing the real task-group wrapping.

## Fix

New `_is_clean_cancel_scope_shutdown(exc)`:

- walks `ExceptionGroup`s recursively (`exc.exceptions`), and
- classifies the tree as a clean shutdown **only when every leaf** is a cancel-scope RuntimeError (per the existing `_ANYIO_CANCEL_SCOPE_SHUTDOWN_MESSAGES` markers).

A group mixing in any real failure still hits the barrier, logs at ERROR, and re-raises. `main()` now catches `(RuntimeError, ExceptionGroup)`; the bare-`RuntimeError` path is unchanged. Plain `BaseExceptionGroup`s containing non-`Exception` leaves (e.g. `KeyboardInterrupt`) keep propagating untouched, as before — `ExceptionGroup` is the `Exception`-derived variant, and a group whose leaves are all RuntimeErrors is always an `ExceptionGroup`.

## Tests

- `test_exception_group_wrapped_cancel_scope_error_exits_cleanly` — single-wrap and nested-group shapes (the message anyio actually uses, `"unhandled errors in a TaskGroup"`): clean exit, WARNING not ERROR. **Pre-fix this group escaped `main()` entirely** (uncaught, non-zero exit).
- `test_exception_group_with_real_failure_hits_barrier` — `[cancel-scope RuntimeError, ValueError]` group is logged and re-raised (also passes pre-fix via the generic barrier; pins no-regression on the strict side).
- Full suite: 2993 passed, 3 skipped (`-m "not ollama"`). `ruff` and `mypy` clean on the touched file.

## Known related gap (out of scope)

The daemon teardown path (`daemon/server.py` `_teardown`) can exit the LTM transport's cancel scope from a different task than entered it and currently swallows that error via `_quiet()`, which can leak the warm LTM child on shutdown. Fixing it is a contested design choice (eager adapter start in the serve task vs. a dedicated worker task vs. guard + force-kill) tracked in the 2026-06-10 implementation review backlog — deliberately not bundled here.

### DIFF
--- a/src/memtomem_stm/server.py
+++ b/src/memtomem_stm/server.py
@@ -1426,6 +1426,22 @@ def _is_anyio_cancel_scope_shutdown_error(exc: RuntimeError) -> bool:
     return any(marker in message for marker in _ANYIO_CANCEL_SCOPE_SHUTDOWN_MESSAGES)
 
 
+def _is_clean_cancel_scope_shutdown(exc: BaseException) -> bool:
+    """True when *exc* consists only of AnyIO cancel-scope shutdown errors.
+
+    ``mcp.run()`` executes the server inside ``anyio.create_task_group()``,
+    and anyio >= 4 (strict task groups) wraps anything escaping a task group
+    in an ``ExceptionGroup`` — on stdio EOF the cancel-scope RuntimeError
+    reaches ``main()`` in that wrapped shape, not bare. Walk groups
+    recursively and treat the tree as a clean shutdown only when EVERY leaf
+    is the cancel-scope error; any other leaf is a real failure that must
+    hit the exception barrier.
+    """
+    if isinstance(exc, BaseExceptionGroup):
+        return all(_is_clean_cancel_scope_shutdown(sub) for sub in exc.exceptions)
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #404, opened 2026-05-31
Fix surfacing feedback demotion

memtomem:
## Summary
- add STM-side shadow demotion for memories with repeated negative surfacing feedback
- count distinct negative surfacing events per memory, including legacy blanket negative ratings
- document the demotion knobs and cover the behavior with regression tests

Closes #372

## Tests
- uv run pytest tests/test_feedback.py tests/test_surfacing_engine.py tests/test_config_constraints.py -q
- uv run ruff check src/memtomem_stm/surfacing/config.py src/memtomem_stm/surfacing/feedback_store.py src/memtomem_stm/surfacing/engine.py tests/test_feedback.py tests/test_surfacing_engine.py tests/test_config_constraints.py
- git diff --check

PR #429, opened 2026-06-10
fix(surfacing): classify no_results_demoted so it survives the stats healthy/fault split

memtomem:
## Summary

PR #404 introduced the `no_results_demoted` skip reason to distinguish "score filter passed but feedback demotion killed every candidate" from `no_results_dedup` / `no_results_score`. But the label was recorded as a bare string in `engine.py` and never added to the `SkipReason` Literal, `HEALTHY_SKIP_REASONS`, or `FAULT_SKIP_REASONS`. `_split_skip_reasons_by_category` (server.py) renders only reasons present in one of the two frozensets and silently drops everything else, so the count never reached `stm_surfacing_stats` output — the operator-facing signal #404 existed to provide. `docs/surfacing.md` already documents the label, so the documented operator workflow was broken.

The CI guard didn't catch it: `test_categorization_is_exhaustive_and_disjoint` inspects only `get_args(SkipReason)`, and the bare string never joined the Literal.

## Fix

- Add `no_results_demoted` to the `SkipReason` Literal — this also puts it under the exhaustiveness test's protection like every other reason.
- Classify it in `HEALTHY_SKIP_REASONS`: an expected threshold outcome (operator tuning knob: `feedback_demotion_negative_threshold`), not a degraded-LTM fault.

## Behavior change

`stm_surfacing_stats` output now includes `no_results_demoted` counts under the "Healthy skips" subsection. Previously the label was recorded internally but invisible in the rendered stats (and in any per-tool totals derived from the split views). No config or API surface changes.

## Tests

- `test_no_results_demoted_records_skip` (engine): drives the all-demoted path with a real `FeedbackTracker` + `SurfacingObservability` wired and pins `snapshot()["skip_reasons"]["read_file"] == {"no_results_demoted": 1}`, mirroring the sibling `test_no_results_dedup_records_skip`.
- `test_no_results_demoted_renders_under_healthy_subsection` (render path): pins that the label survives `_split_skip_reasons_by_category` into the Healthy skips subsection. **Fails before the fix** — the rendered output is empty because the only recorded reason is dropped.

`uv run pytest -m "not ollama"`: 2999 passed, 3 skipped. `ruff check src` + `ruff format --check src` clean. mypy: no errors in touched files (8 pre-existing in compression.py untouched).

### DIFF
--- a/src/memtomem_stm/surfacing/observability.py
+++ b/src/memtomem_stm/surfacing/observability.py
@@ -41,6 +41,11 @@ SkipReason = Literal[
     "gate_cooldown",
     "no_results_score",
     "no_results_dedup",
+    # #404: durable negative-feedback demotion excluded every candidate
+    # that passed the score filter — distinct from ``no_results_score`` /
+    # ``no_results_dedup`` so an operator can tell the demotion filter
+    # (not the threshold or session-dedup) is suppressing surfacing.
+    "no_results_demoted",
     "no_results_invalidated",
     "no_results_empty_cache",
     # #295: LTM adapter outcome typing — distinguish unreachable from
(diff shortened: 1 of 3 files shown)


pandas-studio (approved):
LGTM

PR #430, opened 2026-06-10
fix(mms): treat non-dict mcpServers in host configs as no candidates instead of crashing

memtomem:
## Summary

`scan_claude_code`, `scan_cursor`, and `scan_claude_desktop` guard `mcpServers` with `(config.get("mcpServers") or {})`, which only replaces falsy values. A structurally-valid JSON whose `mcpServers` key is a list, string, or number is truthy and goes straight to `.items()` — an unhandled `AttributeError` that crashes `mms import`, `mms host status`, `mms host scan`, and `mms host sync`. Any other tool writing `~/.claude.json`, `<cwd>/.mcp.json`, `~/.cursor/mcp.json`, or the Claude Desktop config with a different schema (or a simple hand-edit) triggers it.

The module docstring promises unreadable configs scan as "no candidates (silent)" — wrong-typed `mcpServers` should behave like a missing file, not a traceback. `scan_codex` already has exactly this guard (`if not isinstance(servers, dict): return []`); the JSON scanners were the inconsistency.

## Fix

Extract a `_mcp_servers(config) -> dict` helper (returns the value if it's a dict, else `{}`) and use it at all five `mcpServers` call sites — user scope, per-project scope, and `.mcp.json` in `scan_claude_code`, plus `scan_cursor` and `scan_claude_desktop` — so the guard cannot drift per-site.

## Tests

- `TestNonDictMcpServers` (scanner level): parametrized list/str/int at every site; **all fail before the fix** with `AttributeError: 'list' object has no attribute 'items'` (and str/int equivalents).
- `test_bad_user_scope_does_not_block_other_scopes`: a wrong-typed user scope skips only that scope — the scan still picks up `.mcp.json` candidates.
- `test_status_non_dict_mcpservers_exit_0` (CLI level): pins the exit-0 contract for `mms host status` against a list-typed `mcpServers`, extending the existing invalid-JSON edge case that `_read_json_safely` already covered.

`uv run pytest -m "not ollama"`: 3014 passed, 3 skipped. `ruff check src` + `ruff format --check src` clean. mypy: no errors in the touched file.

### DIFF
--- a/src/memtomem_stm/mms/import_hosts.py
+++ b/src/memtomem_stm/mms/import_hosts.py
@@ -224,13 +224,26 @@ def _wrap(name: str, raw: dict[str, Any], source_label: str) -> ImportCandidate 
     )
 
 
+def _mcp_servers(config: dict[str, Any]) -> dict[str, Any]:
+    """Return ``config["mcpServers"]`` if it is a dict, else ``{}``.
+
+    A structurally-valid JSON whose ``mcpServers`` is a list/str/number is
+    truthy, so an ``or {}`` guard would pass it straight to ``.items()`` and
+    crash the scan with an AttributeError. Wrong-typed entries follow the
+    module contract for unreadable configs — no candidates, silent — same as
+    ``scan_codex``'s ``isinstance`` guard on ``mcp_servers``.
+    """
+    servers = config.get("mcpServers")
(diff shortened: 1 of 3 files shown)


pandas-studio (approved):
LGTM

PR #431, opened 2026-06-10
fix(cli): give the upstream probe an end-to-end deadline so mms health honors --timeout

memtomem:
## Summary

`_probe_one` — the upstream-server probe behind `mms health` (per-server `--timeout`) and `mms add --validate` — only wrapped `session.initialize()` in `asyncio.wait_for`. The transport connect (`async with ctx`) and `session.list_tools()` were unbounded, and for `sse`/`streamable_http` transports no `timeout=`/`sse_read_timeout=` was passed to the SDK client at all. Consequences:

- An SSE/HTTP upstream whose endpoint hangs on TCP connect blocks `mms health --timeout N` indefinitely.
- Any upstream that completes initialize but stalls on `tools/list` does the same.
- `_safe_probe` only catches `asyncio.TimeoutError`, which was never raised for the unbounded phases — so the user never even got the `timeout (Ns)` classification.

This is exactly the failure mode `_probe_ltm_mcp_server` (#398) was written to avoid — that function uses a shared end-to-end deadline wrapping transport enter, `initialize()`, AND `list_tools()` — but the older upstream probe never received the same treatment.

## Fix

Mirror the sibling's deadline pattern in `_probe_one`: compute the deadline once, pass the remaining budget as `timeout=`/`sse_read_timeout=` to `sse_client`/`streamablehttp_client`, and wrap `stack.enter_async_context(ctx)`, `initialize()`, and `list_tools()` each in `asyncio.wait_for(..., timeout=remaining())`.

## Behavior change

`mms health` / `mms add --validate` now resolve to DISCONNECTED with `timeout (Ns)` within the `--timeout` bound for connect-phase and list_tools-phase stalls, instead of hanging indefinitely. Successful-probe behavior is unchanged.

## Tests

All invoke `_probe_one`/`_probe_servers` directly (CliRunner's stderr lacks `fileno` for the real subprocess path) and **fail before the fix** — each hangs ~8s on the fake transport instead of timing out at 0.1s:

- `test_upstream_probe_timeout_bounds_transport_enter` (parametrized sse + streamable_http): hanging `__aenter__` → `TimeoutError` within bound; also pins that the SDK client receives `timeout=`/`sse_read_timeout=` ≈ the budget (mirroring `test_sse_ltm_probe_timeout_bounds_transport_enter`).
- `test_upstream_probe_timeout_bounds_list_tools`: initialize succeeds, `tools/list` stalls → bounded.
- `test_safe_probe_classifies_deadline_timeout`: the deadline timeout surfaces through `_probe_servers` as `connected: False` + `error: "timeout (0.1s)"`.

`uv run pytest -m "not ollama"`: 3001 passed, 3 skipped. `ruff check src` + `ruff format --check src` clean. mypy: no errors in the touched file.

### DIFF
--- a/src/memtomem_stm/cli/proxy.py
+++ b/src/memtomem_stm/cli/proxy.py
@@ -2317,11 +2317,29 @@ def prune(
 
 
 async def _probe_one(cfg: dict[str, Any], timeout: float) -> dict[str, Any]:
-    """Probe a single upstream server: connect, initialize, list tools."""
+    """Probe a single upstream server: connect, initialize, list tools.
+
+    *timeout* is an **end-to-end** budget shared across transport connect +
+    ``initialize()`` + ``list_tools()`` — each ``wait_for`` gets
+    ``deadline - now`` so a stall in any phase can't push the probe past the
+    ``mms health --timeout`` / ``mms add --validate --timeout`` ceiling.
+    Mirrors ``_probe_ltm_mcp_server``'s deadline pattern; previously only
+    ``initialize()`` was bounded, so a network upstream hanging on TCP
+    connect (or any upstream stalling on ``tools/list``) blocked the probe
+    indefinitely and ``_safe_probe``'s timeout classification never fired.
+    """
     from mcp import ClientSession
     from mcp.client.sse import sse_client
     from mcp.client.stdio import StdioServerParameters, stdio_client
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #432, opened 2026-06-10
fix(config): load the JSON file even when MEMTOMEM_STM_PROXY__ENABLED is set, and propagate file consumer_model to surfacing

memtomem:
## Summary

Two silent config-propagation bugs (2026-06-10 review, medium):

1. **`MEMTOMEM_STM_PROXY__ENABLED` dropped the whole JSON file.** `app_lifespan` only called `load_from_file` when the env var was NOT set. Enabling the proxy via env (a documented method) skipped the file load entirely — `upstream_servers` (a file-only field env vars cannot practically populate) came up empty, so the proxy started enabled but proxied nothing. The env-overlay mechanism (`collect_proxy_env_overrides` + `_deep_merge`) was built precisely to layer env over the file, and it already includes `enabled` — so env still wins without the guard.

2. **File-set `consumer_model` never reached surfacing.** `STMConfig.model_post_init` copies `proxy.consumer_model` → `surfacing.consumer_model`, but it runs once at construction — before `app_lifespan` swaps `config.proxy` to the file config. A `consumer_model` set only in `stm_proxy.json` (the documented place) reached `config.proxy` but never surfacing, so the model-aware budgets (`effective_max_injection_chars` / `effective_max_results`) silently used un-scaled defaults.

## Fix

Both fixes live in a new `_apply_proxy_file_config(config, proxy_env_overrides)` helper called from `app_lifespan`: load the file unconditionally with env deep-merged on top, then re-apply the `consumer_model` propagation after the swap (same `if proxy set and surfacing unset` guard as `model_post_init`).

**CLI mirror (#420) updated in the same change:** `mms stats` mirrored the server's env bypass (`config_status: "env"`, "JSON file bypassed" output) to keep probing the same metrics db as the server. With the server now honoring the file, the mirror does too — the `"env"` status and bypass text are removed, and the file-level `metrics.db_path` is honored under env-enabled mode. Updating one side without the other would make them contradict. The `collect_proxy_env_overrides` docstring loses its stale "except `ENABLED`" clause.

## Behavior change

- With `MEMTOMEM_STM_PROXY__ENABLED` set **and** a JSON config file present, the server now loads the file (upstreams, metrics paths, compression settings, …) and overlays env vars on top, instead of running env-only with zero upstreams. Set env vars still win field-by-field; `enabled` from env still wins over the file.
- `mms stats` no longer reports `config_status: "env"` / "JSON file bypassed" — it reports the file status (`ok`/`missing`/`invalid`) and honors a file-level `metrics.db_path`, matching the new server behavior.
- A `consumer_model` set only in the JSON file now activates surfacing's model-aware injection/result scaling (previously inert).

## Tests

- `test_file_config_loaded_even_when_env_enabled` (lifespan level): pins that `load_from_file` is called with the env overlay even when the env var is set — **fails before the fix** (the guard skips the call).
- `TestApplyProxyFileConfig` (helper contract): env-enabled keeps file upstreams while env wins on `enabled`; file `consumer_model` reaches `effective_max_injection_chars()` (3000 → 5000 for a >200K-context model); an explicitly-set surfacing `consumer_model` is not clobbered.
- `test_env_enabled_still_honors_file_metrics_path` (CLI): rewritten from the #420 bypass test; asserts `config_status: "ok"` and that the file's seeded metrics db is probed — **fails before the fix** (`"env"`, db ignored).

`uv run pytest -m "not ollama"`: 3001 passed, 3 skipped. `ruff check src` + `ruff format --check src` clean. mypy: no errors in touched files.

### DIFF
--- a/src/memtomem_stm/cli/proxy.py
+++ b/src/memtomem_stm/cli/proxy.py
@@ -685,13 +685,13 @@ def stats(config_path: str, *, tool_filter: str | None = None, as_json: bool = F
     path = Path(config_path)
     resolved = path.expanduser().resolve()
 
-    # Resolve config + DB paths exactly as the server does (server.app_lifespan).
-    # ``STMConfig()`` pulls proxy + surfacing from ``MEMTOMEM_STM_*`` env. When
-    # the proxy is enabled via env, the server uses that env-only config and
-    # SKIPS the JSON file entirely (server.py), so honoring a file-level
-    # ``metrics.db_path`` here would point ``mms stats`` at a different
-    # ``proxy_metrics.db`` than the one the server actually writes to.
-    env_enabled = bool(os.environ.get("MEMTOMEM_STM_PROXY__ENABLED"))
+    # Resolve config + DB paths exactly as the server does (server.app_lifespan):
+    # JSON file loaded unconditionally with env vars deep-merged on top, so a
+    # file-level ``metrics.db_path`` points ``mms stats`` at the same
+    # ``proxy_metrics.db`` the server writes to even when the proxy is enabled
+    # via ``MEMTOMEM_STM_PROXY__ENABLED``. (The server used to skip the file
+    # in env-enabled mode and this command mirrored the bypass; both honor the
+    # file now.) ``STMConfig()`` still supplies the surfacing feedback path.
     try:
         stm_config: STMConfig | None = STMConfig()
     except Exception:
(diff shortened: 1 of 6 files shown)


memtomem:
Codex R1 flagged a regression in the missing-file env-only path: rebuilding `config.proxy` from the raw-string env overlay fails validation for JSON-encoded complex values (e.g. `MEMTOMEM_STM_PROXY__UPSTREAM_SERVERS='{"gh": ...}'`) that pydantic-settings accepts, collapsing a working env-only setup to defaults. Fixed in `c5c4575`: when the file is missing, the swap is skipped entirely — `STMConfig()`'s pydantic-settings parse is strictly more capable than the overlay rebuild, so the rebuild could only lose information. `mms stats` mirrors the same rule on its `missing` branch. Regression tests added at the helper and CLI level.

memtomem:
R2 finding addressed in `8d96c78`: the missing-file decision now lives inside the single `load_from_file` call (`missing_ok=False` → `None` → no swap) instead of a separate `exists()` pre-check, so deletion can't race the two checks apart. Every disappearance mode (missing at check, deleted before read) converges on `None`. `mms stats` uses the same mode; its post-hoc `exists()` only selects the status label for read-only display. Hot-reload loader keeps the default `missing_ok=True` behavior unchanged.

pandas-studio (approved):
LGTM

PR #433, opened 2026-06-10
fix(surfacing): re-apply durable feedback demotion on the cache-hit path

memtomem:
## Summary

#404's negative-feedback demotion filter runs only on the cache-MISS path, before the cache write. `_render_cached` filters cache hits solely against `_invalidated_ids` — an in-memory set populated only by feedback submitted through *this* engine during the TTL window. A memory that crosses the durable demotion threshold via feedback recorded in a **different process** (the proxy and daemon share `stm_feedback.db`) is re-surfaced by every cache hit until the entry TTL-expires — contradicting the miss-path comment's promise that a rejected memory "does not keep reappearing from a cached query". (2026-06-10 review; reported both as the medium L89 and low L116 findings — same bug.)

## Fix

Intersect the cached list with `_feedback_demoted_ids(...)` in `_render_cached`, symmetric with the miss path:

- The helper already gates itself on `feedback_tracker` / `record_feedback_events` / `feedback_demotion_enabled`, so the dedup-only daemon path and no-tracker engines skip the DB read entirely — no new config knob needed.
- When demotion (not invalidation) empties a hit, the same `no_results_demoted` label the miss path records is used, so `stm_surfacing_stats` shows which filter suppressed the cached result. The miss-path comment now documents the hit-path symmetry.

## Tests

Both drive the real `FeedbackTracker` across a miss→hit sequence, with threshold-crossing feedback recorded **directly on the store** (as another process would — `_invalidated_ids` stays empty), and **fail before the fix** with the demoted memory resurfacing from cache:

- `test_cache_hit_filters_durably_demoted_memory`: all-demoted hit → content absent, `cache == {miss: 1, hit: 1}`, skip label `no_results_demoted`.
- `test_cache_hit_keeps_non_demoted_memories`: partial demotion → offending memory absent, surviving memory still renders as a normal `surfaced_cache_hit`.

`uv run pytest -m "not ollama"`: 2999 passed, 3 skipped. `ruff check src` + `ruff format --check src` clean.

Note: mypy (advisory) flags the bare `no_results_demoted` literal at both record sites — the miss-path one has existed since #404; both resolve when the `SkipReason` Literal member lands (#429).

### DIFF
--- a/src/memtomem_stm/surfacing/engine.py
+++ b/src/memtomem_stm/surfacing/engine.py
@@ -561,8 +561,13 @@ class SurfacingEngine:
 
         Filters out memory IDs in ``_invalidated_ids`` — memories the agent
         already rated ``not_relevant`` or ``already_known`` within the cache
-        TTL window. A filtered-empty result pass-throughs like the natural
-        empty case.
+        TTL window — and re-applies the durable feedback-demotion filter,
+        symmetric with the miss path: a memory that crossed the
+        negative-feedback threshold after this entry was cached (e.g.
+        feedback recorded by another process sharing ``stm_feedback.db``,
+        which never reaches this engine's in-memory sets) must not keep
+        resurfacing from a warm entry for the rest of the TTL. A
+        filtered-empty result pass-throughs like the natural empty case.
         """
         # Track whether the entry started non-empty so we can distinguish a
         # ``no_results_invalidated`` outcome (had results, all rejected) from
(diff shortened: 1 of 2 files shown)


memtomem:
Codex R1's Major (`no_results_demoted` unclassified → dropped from `stm_surfacing_stats`) is exactly what **#429** fixes — SkipReason Literal member + `HEALTHY_SKIP_REASONS` classification + render-path test (`test_no_results_demoted_renders_under_healthy_subsection`). That PR is already codex-SHIP-approved. Duplicating the classification here would conflict with #429 on the same lines of `observability.py`, so this stays split per the one-focused-change convention. **Merge order: #429 before (or together with) #433** — the miss path has had the identical gap since #404 either way, so #433 standalone makes nothing worse than main; #429 closes the rendering gap for both record sites at once.

pandas-studio (approved):
LGTM

PR #434, opened 2026-06-10
fix(compression): bound over-length LLM summaries to max_chars on the success path

memtomem:
## Summary

Every sync compressor tier provably bounds its output to `max_chars` (the 2026-06-10 review verified this empirically across 300+ random JSON shapes — zero over-budget). `LLMCompressor.compress()` is the sole exception: the success path returns the model's content string verbatim. The system prompt only *asks* the model to honor `max_chars` (`system_prompt.format(max_chars=...)`), and LLMs routinely overshoot length constraints. Nothing downstream clamps it — the manager's ratio guard fires only on UNDER-retention — so `LLM_SUMMARY` can silently return a payload **larger** than the budget the whole pipeline exists to enforce. Every failure path (timeout / circuit-breaker / privacy / error) already bounds output via `TruncateCompressor`, which highlights the success path as the one gap.

## Fix

When the model's output exceeds `max_chars`, truncate the **summary** (not the original text — the LLM's distillation retains more signal per char) via the same `TruncateCompressor` the failure paths use, log a warning, and set `last_fallback = "llm_overlength"` so the overshoot is observable through the existing fallback channel (`manager` reads `compressor.last_fallback` for metrics). The circuit breaker is **not** failed — the API responded fine; overshoot is a model-quality event, not an endpoint failure.

## Tests

- `test_overlength_llm_output_is_bounded`: mocked `_call_api` returning a 1000-char string at `max_chars=200` → result ≤ 200 + `llm_overlength` marker. **Fails before the fix** (over-budget payload leaks through).
- `test_overlength_does_not_count_as_breaker_failure`: 4 consecutive overshoots leave the breaker closed.
- `test_within_budget_llm_output_returned_verbatim`: compliant summaries are untouched, `last_fallback` stays `None`.

`uv run pytest -m "not ollama"`: 3000 passed, 3 skipped. `ruff check src` + `ruff format --check src` clean.

### DIFF
--- a/docs/compression.md
+++ b/docs/compression.md
@@ -262,7 +262,7 @@ Routes through an external LLM for intelligent summarization:
 }
 ```
 
-Providers: `openai`, `anthropic`, `ollama`. `llm_timeout_seconds` (default 60.0) bounds the per-call LLM wait. On timeout, privacy-pattern hit, circuit-breaker open, or any other API failure, compression falls back to `TruncateCompressor` and records the strategy as `llm_summary→{timeout,privacy,circuit_breaker,llm_error}_fallback` in `proxy_metrics` for observability.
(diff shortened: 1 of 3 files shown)


memtomem:
Codex minor addressed in `fb5134a`: `docs/compression.md` fallback-value enumeration now includes `llm_overlength` with its semantics (model overshot the length instruction; endpoint fine, breaker unaffected).

pandas-studio (approved):
LGTM

PR #435, opened 2026-06-10
fix(progressive): read_more continues with the originating tool's chunk_size and hint preference

memtomem:
## Summary

`read_more` hardcodes `chunk_size = limit or 4000` and `include_hint=True`, while `_apply_progressive` builds the **first** chunk from the per-tool `ProgressiveConfig` (`cfg.chunk_size` / `cfg.include_structure_hint`). `ProgressiveResponse` never persisted those values, so `read_more` couldn't recover them. For a tool configured with `progressive.chunk_size=2000, include_structure_hint=False`: the first chunk is 2000 chars with no `[Remaining: ...]` hint, but every follow-up reverts to 4000 chars **and** emits hints — operators who tuned the chunk size down to fit a smaller context window silently get 2x-larger continuation reads. (2026-06-10 review, medium; empirically reproduced there.)

## Fix

Persist `chunk_size` and `include_structure_hint` on `ProgressiveResponse` — set from `cfg` in `_apply_progressive`, round-tripped through `ProgressiveStoreAdapter`'s `__meta__` JSON — and read them back in `read_more` (`chunk_size = limit or resp.chunk_size`, `include_hint=resp.include_structure_hint`). An explicit `limit` from the agent still wins; only the no-limit default changes. The new dataclass fields default to `4000` / `True` (the historical hardcoded behavior), so rows persisted before this change deserialize to exactly the previous semantics.

## Tests

- `test_follow_up_uses_configured_chunk_size`: **fails before the fix** (`4000 != 2000`).
- `test_follow_up_respects_hint_suppression`: **fails before the fix** (hint re-appears on the follow-up).
- `test_explicit_limit_still_overrides`: agent-passed `limit` keeps winning (passes pre and post — invariant pin).
- `test_store_adapter_round_trips_chunking_fields` + `test_legacy_meta_rows_default_to_previous_behavior`: adapter round-trip and legacy-row compatibility.

`uv run pytest -m "not ollama"`: 3002 passed, 3 skipped. `ruff check src` + `ruff format --check src` clean. mypy: no errors in touched files.

### DIFF
--- a/src/memtomem_stm/proxy/manager.py
+++ b/src/memtomem_stm/proxy/manager.py
@@ -1044,6 +1044,8 @@ class ProxyManager:
             server=server,
             tool=tool,
             trace_id=trace_id,
+            chunk_size=cfg.chunk_size,
+            include_structure_hint=cfg.include_structure_hint,
         )
         store.put(key, resp)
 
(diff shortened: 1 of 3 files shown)


pandas-studio (approved):
LGTM

PR #437, opened 2026-06-10
fix(daemon): degrade on corrupted handshake fields in mms daemon stop/status, and cover the ops CLI with tests

memtomem:
## Summary

Two paired findings from the 2026-06-10 review (both medium):

1. **Corrupted handshake → traceback instead of degrading.** `read_handshake()` guarantees only that the parsed JSON is a dict — its docstring (and the discovery module's) explicitly anticipate hand-edited/corrupted files, but field types are not validated. `stop_cmd` does a bare `int(raw.get("pid", -1))`: a handshake whose `pid` is a non-numeric string or `null` raises an uncaught `ValueError`/`TypeError` and Click prints a traceback — crashing **before** the tail branch whose whole purpose is cleaning up the stale/bad handshake. `status_cmd` has the identical exposure at its stale-branch `int()` and at `float(hs.get("created_at", ...))`. (`client.shutdown()` returns `False` — never raises — for an unreachable daemon, so control reliably reaches the crashing lines.)
2. **Zero CliRunner coverage for `mms daemon stop` / `status` / `restart`** — the commands carrying the unguarded casts plus the SIGTERM-fallback and handshake-cleanup branches.

## Fix

Tolerant `_as_int` / `_as_float` coercion helpers applied at the three sites. `try/except (TypeError, ValueError)` coercion — not `isinstance` checks — so every previously-working input (numeric strings, float pids) coerces exactly as before; only the previously-crashing inputs now degrade:

- `stop`: corrupted `pid` → falls through to the handshake-unlink cleanup, `exit 0`, "cleaned stale handshake".
- `status`: corrupted stale `pid` → `state=stale, pid_alive=false`; garbage `created_at` → `uptime 0.0`.

## Tests

New CLI suite in `tests/daemon/test_internals.py` (10 tests):

- **stop**: graceful ack / no daemon + no handshake / **corrupted-pid degrade-and-clean (fails pre-fix: `TypeError`)** / stale-handshake SIGTERM (recorded `os.kill`, POSIX-only).
- **status**: running `--json` shape / **corrupted `created_at` (fails pre-fix: `ValueError`)** / **stale corrupted `pid` (fails pre-fix: `TypeError`)** / stopped.
- **restart**: stop-then-start composition / wedged-teardown "still shutting down" message (lifetime lock held; 5s wait fast-forwarded via a fake module clock).

`uv run pytest -m "not ollama"`: 3007 passed, 3 skipped. `ruff check src` + `ruff format --check src` clean. mypy: no errors in the touched file.

## Not in this PR

The third Group E finding — the daemon `_teardown` cross-task cancel-scope exit for a STARTED LTM adapter (the #428 out-of-scope note) — is a contested design choice (eager-warm vs worker-task vs guard+force-kill) and is left for a follow-up decision.

### DIFF
--- a/src/memtomem_stm/cli/daemon_cmd.py
+++ b/src/memtomem_stm/cli/daemon_cmd.py
@@ -59,18 +59,28 @@ def _as_int(value: Any, default: int = -1) -> int:
     turned a garbage ``pid`` into an uncaught ValueError/TypeError traceback
     in the very commands whose tail branches exist to clean up bad
     handshakes. Degrade to *default* instead.
+
+    ``bool`` is rejected explicitly: JSON ``true`` would coerce to pid 1
+    (launchd/init) and could steer the SIGTERM fallback at the wrong
+    process. ``OverflowError`` covers JSON ``Infinity`` — Python's
+    ``json.loads`` accepts it by default, and ``int(float("inf"))`` raises
+    it rather than ValueError.
     """
+    if isinstance(value, bool):
+        return default
     try:
         return int(value)
-    except (TypeError, ValueError):
+    except (TypeError, ValueError, OverflowError):
         return default
 
 
 def _as_float(value: Any, default: float) -> float:
     """Tolerant float twin of :func:`_as_int` (e.g. ``created_at``)."""
+    if isinstance(value, bool):
+        return default
     try:
         return float(value)
-    except (TypeError, ValueError):
+    except (TypeError, ValueError, OverflowError):
         return default
 
 
--- a/src/memtomem_stm/cli/daemon_cmd.py
+++ b/src/memtomem_stm/cli/daemon_cmd.py
@@ -50,6 +50,30 @@ def _load_config() -> STMConfig:
     return STMConfig()
 
 
+def _as_int(value: Any, default: int = -1) -> int:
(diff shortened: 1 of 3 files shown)


memtomem:
Codex R1 Major addressed in `643addc`: `_as_int`/`_as_float` now reject `bool` explicitly (JSON `true` → pid 1 could steer the SIGTERM fallback at launchd/init) and catch `OverflowError` (Python's `json.loads` accepts the `Infinity` extension by default, and `int(float('inf'))` raises OverflowError, not ValueError). The corrupted-pid stop test is parametrized over `str`/`null`/`Infinity`/`-Infinity`/`true` and asserts no kill is attempted.

memtomem:
Codex R2 Major addressed in `e6c9e65`: `is_pid_alive` gains an `OverflowError` arm returning `False` — a pid beyond pid_t range (huge hand-edited integer) cannot name a live process, and `OverflowError` is not an `OSError` subclass so it escaped every existing arm. Real syscall behavior pinned in `test_is_pid_alive` (`10**1000 → False`); the CLI corrupted-pid parametrization gains a `huge-int` case with an `os.kill` fake that faithfully mimics the pid_t overflow.

pandas-studio (approved):
LGTM

PR #438, opened 2026-06-10
fix(daemon): sweep for a leaked LTM child when adapter teardown unwinds cross-task

memtomem:
## Background

E-3 from the 2026-06-10 implementation review (the one remaining item after #429–#437): `DaemonServer._teardown()` exits the LTM adapter's anyio cancel scopes from the serve task, while they were entered in whichever connection-handler task first lazy-started the adapter. anyio raises its cross-task cancel-scope `RuntimeError` partway through the stdio transport's unwind; `_quiet()` swallowed it at DEBUG, so whether the warm LTM child was actually reaped went unverified.

The new end-to-end test confirms the error **does** fire on every real warm teardown (real engine wiring against `tests/_fake_memtomem_server.py`). Today's mcp terminates the child before the failing scope exit — the child does not leak with the current pin — but that ordering is an implementation detail of mcp's stdio shutdown `finally`, and a reorder silently re-opens the leak with no signal at default log levels.

This is option (c) from the review: guard + force-kill, no change to the daemon's resource model (no eager warm-up, no dedicated LTM worker task).

## What changed

- `_teardown()` stops the adapter via a new `_stop_adapter()`:
  - The known cancel-scope shape (bare `RuntimeError` or the anyio ≥ 4 group-wrapped form) is logged at WARNING instead of being swallowed at DEBUG.
  - After **any** failed stop, the daemon sweeps its direct children (`pgrep -P`, POSIX best-effort, no-op on Windows): a child present both before and after a failed stop is the leaked LTM child — the daemon process spawns nothing else. Leaked children get SIGTERM to their own process group (the stdio child is spawned `start_new_session=True`, so the group kill also reaches `uv run`-style grandchildren), escalating to SIGKILL after the same 2 s grace mcp's own shutdown sequence uses. A clean stop performs no sweep.
- The cancel-scope detection helpers (`_is_anyio_cancel_scope_shutdown_error` / `_is_clean_cancel_scope_shutdown` and the message tuple) move from `server.py` to a new `utils/anyio_shutdown.py` so the daemon can reuse them without importing the whole proxy stack. `server.py`'s stdio-EOF exception barrier imports them from there; nothing outside `server.py` referenced the old underscore names.

## Behavior change

- Daemon shutdown now emits a WARNING (`cross-task cancel-scope`) on every warm-LTM teardown under the current mcp pin — previously this path was silent outside DEBUG. This is intentional observability for a real (if currently benign) error.
- On a failed adapter stop, the daemon may now signal surviving direct child processes. In the dedicated `mms daemon run` process the only possible direct child is the LTM stdio child; the sweep is intersection-scoped (before ∩ after) and skipped entirely when stop succeeds.

## Tests

- Unit: leaked-child kill on the bare and group-wrapped cancel-scope shapes (real `sleep` child in its own session, asserts `-SIGTERM` exit and both WARNING lines).
- Unit: generic stop failure sweeps exactly the before ∩ after intersection; clean stop performs no sweep and no second probe.
- Unit: `_direct_child_pids()` observes a real spawned child and stops listing it after reap.
- E2E regression: real `_build_engine` wiring + fake memtomem stdio server; a surface call warms a real LTM child in a connection-handler task, daemon shutdown must leave no live LTM child (pinned regardless of which unwind path reaps it).

Full suite: 3062 passed, 3 skipped (`-m "not ollama"`). `ruff check` + `ruff format --check src` clean; mypy introduces no new errors in touched files.

### DIFF
--- a/src/memtomem_stm/daemon/server.py
+++ b/src/memtomem_stm/daemon/server.py
@@ -33,12 +33,15 @@ import logging
 import os
 import secrets
 import signal
+import subprocess
+import sys
 import time
 from typing import Any
 
 from memtomem_stm.cli.hook_cmd import run_surfacing_hook
 from memtomem_stm.config import STMConfig
 from memtomem_stm.daemon import discovery, locking
+from memtomem_stm.utils.anyio_shutdown import is_clean_cancel_scope_shutdown
 from memtomem_stm.daemon.protocol import (
     MAX_MESSAGE_BYTES,
     OP_PING,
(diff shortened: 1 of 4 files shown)


pandas-studio (approved):
LGTM

PR #439, opened 2026-06-10
fix(proxy): null the extractor on stop and sanitize _context_query on the miss path

memtomem:
## Background

Two low findings from the 2026-06-10 implementation review (report L55, L59 — first wave of the low-severity backlog, re-verified against current main before fixing).

## What changed

- **`stop()` now nulls `_extractor` after closing it.** It closed the extractor but left the attribute pointing at the closed instance, unlike `_llm_compressor` two lines above which is set to `None`. `_get_extractor()` rebuilds only on `None`, so a stop→start cycle returned the closed extractor and the first post-restart extraction hit `FactExtractor`'s `assert self._client is not None` (extraction.py sets `_client = None` in `close()`). A restarted manager could never extract again.

- **The miss path now coerces a non-str `_context_query` to `None` at the extraction point.** The cache-hit path already guards with `isinstance(context_query, str)`; the miss path threaded the raw value into compression, surfacing, progressive surfacing, auto-index, and extract-and-store. An agent passing `_context_query` as a JSON object/number could raise inside a scorer on a miss but be silently dropped on a hit — behavior depending solely on cache state. Both paths now sanitize identically; `_context_query` is still stripped from upstream arguments.

## Tests

- `test_stop_nulls_extractor_so_restart_rebuilds` — pins `_extractor is None` after `stop()` (fails pre-fix).
- `TestContextQuerySanitization::test_non_str_context_query_coerced_to_none_on_miss_path` — drives `call_tool` end-to-end with a dict `_context_query`, asserts downstream surfacing receives `None` and upstream args are stripped (fails pre-fix).

Affected suites: 55 passed (`test_proxy_manager_lifecycle.py` + `test_proxy_manager_pipeline.py`); ruff clean.

### DIFF
--- a/src/memtomem_stm/proxy/manager.py
+++ b/src/memtomem_stm/proxy/manager.py
@@ -421,6 +421,10 @@ class ProxyManager:
             self._llm_compressor = None
         if self._extractor is not None:
             await self._extractor.close()
+            # Null it like _llm_compressor above: _get_extractor() rebuilds on
+            # None, so a stop->start cycle gets a fresh httpx client instead of
+            # the closed instance (whose extract() asserts _client is not None).
+            self._extractor = None
         for conn in self._connections.values():
             if conn.stack is not None:
                 try:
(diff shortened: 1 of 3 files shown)


pandas-studio (approved):
LGTM

PR #440, opened 2026-06-11
fix(proxy): keep InMemoryPendingStore eviction order in sync across put/touch/delete

memtomem:
## Background

Low findings from the 2026-06-10 implementation review (report L297, L301), re-verified against current main.

`InMemoryPendingStore.evict_oldest` pops from `_order` (insertion order), but three mutations let `_order` drift from recency:

- `put()` appended unconditionally — re-putting an existing key left a duplicate `_order` entry, so eviction popped the re-put key first and **dropped the fresh selection**.
- `touch()` refreshed `created_at` but never moved the key — a just-touched entry was evicted first.
- `delete()` never removed its `_order` entry — a stale entry plus a later re-put reintroduced the duplicate through the delete path (consequent to the `put` fix; on main alone it only wasted an eviction iteration).

The SQLite backend already behaves correctly on all three (`INSERT OR REPLACE` rewrites `created_at`; `evict_oldest` orders by it), so this aligns the in-memory backend with its documented drop-in sibling.

## What changed

A stated invariant — `_order` holds exactly the keys of `_data`, once each, in recency order — and the three mutations now maintain it: `put` moves an existing key to the back, `touch` moves the key to the back, `delete` removes the key's `_order` entry.

## Tests

Three regression tests, each failing pre-fix (verified by stashing the src change):
- `test_reput_refreshes_recency_for_eviction`
- `test_touch_moves_key_to_back_of_eviction_order`
- `test_delete_then_reput_does_not_leave_stale_order_entry`

`test_pending_store.py` 21 passed, `test_progressive.py` green; ruff clean.

### DIFF
--- a/src/memtomem_stm/proxy/pending_store.py
+++ b/src/memtomem_stm/proxy/pending_store.py
@@ -37,8 +37,16 @@ class InMemoryPendingStore:
         self._order: deque[str] = deque()
         self._lock = threading.Lock()
 
+    # Invariant: ``_order`` holds exactly the keys of ``_data``, once each,
+    # in recency order (oldest left) — matching the SQLite backend, whose
+    # evict_oldest keeps the most recent ``created_at`` rows. Every mutation
+    # below maintains it; a duplicate or stale ``_order`` entry makes
+    # evict_oldest silently drop a fresh entry.
+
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #442, opened 2026-06-11
fix(compression): forward the TOC budget, guard scorer length, and survive all-empty BM25 sections

memtomem:
## Background

Low findings L131, L320, L324(+L328 test) from the 2026-06-10 implementation review, re-verified against current main. L324 had been mis-triaged as a false positive by a verification agent — re-running the probe showed `BM25Scorer().score_sections('redis', [('!!!','@@@')])` → `ZeroDivisionError`, so it ships as a real fix.

## What changed

- **`compress_full_toc` forwards `max_chars`** to `_store_and_build_toc`, engaging the existing preview-shrink search (was accepted-and-ignored; method currently has no in-repo callers — contract fix only, flagged for possible future removal).
- **Centralized wrong-length scorer validation** (`_scores_or_none`): codex round 1 found the initial `_json_key_truncate`-only guard left four sibling `score_sections` consumers exposed to a custom scorer returning the wrong number of scores — selective TOC ordering (IndexError in sort key), `_enrich_by_relevance` (IndexError), SchemaPruning (`zip()` prefix taken as real signal), Skeleton (one-element budgets rendered only the first heading). All five consumers now route through the helper and fall back to their query-less path; byte-identical for well-behaved scorers.
- **BM25 `avgdl` floors to 1.0 when it computes to 0** — all-empty tokenization leaves `doc_lens` non-empty but all zeros, passing the existing `if doc_lens` guard and dividing by zero in the BM25 denominator.

## Review

Codex R1 NEEDS-FIX (major: guard only one of five consumers) → **R2 SHIP** (verified byte-identity at all five sites, Skeleton negative-only fallback intact, no consumers outside compression.py).

## Tests

`TestWrongLengthScorerDegradation` — one regression per consumer (four failing pre-fix; the already-guarded site pins the helper refactor), plus `test_all_empty_sections_no_crash` and `test_compress_full_toc_honors_max_chars` (both failing pre-fix). Affected suites 275 passed; ruff clean.

### DIFF
--- a/src/memtomem_stm/proxy/compression.py
+++ b/src/memtomem_stm/proxy/compression.py
@@ -143,6 +143,24 @@ class NoopCompressor:
         return text
 
 
+def _scores_or_none(
+    scorer: RelevanceScorer, query: str, sections: list[tuple[str, str]]
+) -> list[float] | None:
+    """Score *sections*, returning ``None`` when there is no usable signal.
+
+    No-signal covers both an all-non-positive score vector AND a scorer
+    returning the wrong number of scores (custom ``RelevanceScorer`` drift).
+    Every consumer indexes, sorts, or ``zip()``s the scores against its
+    sections, so a wrong-length list would raise IndexError, silently
+    truncate, or mis-allocate budgets — all consumers must obtain scores
+    through this helper and fall back to their query-less path on ``None``.
+    """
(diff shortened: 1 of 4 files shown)


pandas-studio (approved):
LGTM

PR #443, opened 2026-06-11
fix(hook): skip compression below the sentinel budget, offload spawn, fix _orchestrate docs

memtomem:
## Background

Three low findings from the 2026-06-10 implementation review (report L181, L185, L189), re-verified against current main — all in `cli/hook_cmd.py`.

## What changed

- **Tiny-budget skip (L181):** `HookCompressionConfig.max_chars` only validates `gt=0`, so it can be configured below the ~18-char sentinel prefix. `maybe_compress_builtin` floored the body budget at 1 and prepended the sentinel unconditionally — `max_chars=1` produced a ~18-char replacement, ~18× the configured cap. It now returns `None` (leave output unchanged) when the budget cannot hold the sentinel; the now-dead `max(1, ...)` floor is removed.
- **Spawn offload (L185):** on the daemon-unreachable branch, `request_spawn` (flock probe + `subprocess.Popen` fork/exec with a `close_fds` scan) ran directly on the event loop inside the `wait_for`-bounded hook call. Now `await asyncio.to_thread(request_spawn, config)` — same fire-and-forget semantics, off the loop.
- **Docstring (L189):** `_orchestrate` claimed compression shrinks output "before it could be sent whole to the daemon's 4 MiB frame" — actually compression clones the payload and never alters what surfacing receives; frame protection comes solely from `_bounded_surfacing_payload` / `_SAFE_DAEMON_BUDGET`. Reworded to attribute the mechanisms correctly.

## Behavior change

With `max_chars` configured at or below the sentinel length (pathological config), Bash output is now left unchanged instead of being replaced by a sentinel-only stub. Valid configs are unaffected.

## Tests

- `test_compress_skipped_when_budget_below_sentinel` (fails pre-fix)
- `test_hook_run_hook_autospawn_runs_off_event_loop` — asserts the patched `request_spawn` runs on a worker thread (fails pre-fix)

`test_hook_cmd.py` + `test_server.py` 82 passed; ruff clean.

### DIFF
--- a/src/memtomem_stm/cli/hook_cmd.py
+++ b/src/memtomem_stm/cli/hook_cmd.py
@@ -231,7 +231,13 @@ def maybe_compress_builtin(
         # at/near ``max_chars`` (TruncateCompressor's own suffix may still add a
         # small, bounded overage — the budget is a target, not a hard cap).
         prefix = f"{_COMPRESS_SENTINEL}\n"
-        body_budget = max(1, cfg.max_chars - len(prefix))
+        if cfg.max_chars <= len(prefix):
+            # A budget that cannot even hold the sentinel cannot produce a
+            # meaningful replacement — prepending anyway would EXPAND the
+            # configured cap many-fold (max_chars=1 → ~18 chars of sentinel).
+            # Leave the output unchanged instead.
+            return None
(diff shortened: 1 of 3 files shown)


pandas-studio (approved):
LGTM

PR #444, opened 2026-06-11
fix(daemon): cap start respawns, bound the response drain, honor sub-second idle timeouts

memtomem:
## Background

Low findings L154, L162, L166 (+ L258 docs-drift) from the 2026-06-10 implementation review, re-verified against current main.

## What changed

- **Spawn cap (L154):** `mms daemon start` retried `request_spawn` every 0.3 s for 10 s. A child that crashes during startup (invalid LTM command, unreadable DB, bad config) releases the lifetime lock immediately, so every iteration found the lock free and spawned another detached, DEVNULL'd child — ~33 crash-looping processes per `start`. Actual spawns (`request_spawn` returning `True`) are now capped at `_START_MAX_SPAWNS = 3`; the rest of the window keeps ping-polling. The mid-shutdown handoff the loop was built for needs at most one respawn, so the existing retry-until-lock-frees behavior is preserved (its test still passes — declined probes don't count toward the cap).
- **Bounded drain (L162):** `_handle_conn` bounded only the inbound read; `writer.drain()` was unbounded, so a half-open peer that stopped reading pinned the handler task and its buffered response forever. Now `asyncio.wait_for(..., _WRITE_TIMEOUT_SECONDS=30.0)`, symmetric to the read bound; timeout falls through to the existing close path.
- **Idle poll floor (L166):** `_idle_watch` floored its poll at 1.0 s, so `idle_timeout=0.2` shut down ~1 s+ late. Floor lowered to 0.05 s and documented; default (multi-second) timeouts poll exactly as before.
- **Docs (L258):** `config_fingerprint`'s exclusion list now states that *all* of `config.hook` except `record_feedback_events` is excluded — explicitly naming `auto_spawn` and the `compression` sub-block — and the exclusion test gains a compression-toggle pin.

## Tests

- `test_start_caps_spawn_attempts_for_crash_looping_child` — fake clock compresses the 10 s window; asserts exactly 3 spawns instead of ~33 (fails pre-fix).
- `test_handler_write_timeout_drops_stuck_consumer` — stuck-drain writer; handler completes and closes (hangs pre-fix, fails via outer wait_for).
- `test_sub_second_idle_timeout_shuts_down_promptly` — `idle_timeout=0.2` exits in < 0.9 s (pre-fix lower bound was the 1.0 s floor).
- Compression-toggle fingerprint invariance pin.

`tests/daemon/` 63 passed; ruff clean.

### DIFF
--- a/src/memtomem_stm/cli/daemon_cmd.py
+++ b/src/memtomem_stm/cli/daemon_cmd.py
@@ -128,6 +128,13 @@ def run_cmd(detached: bool, foreground: bool) -> None:
     raise SystemExit(run_daemon(config))
 
 
+# Cap on detached children `mms daemon start` may launch in one invocation.
+# A crashed child releases the lifetime lock, making "lock free" ambiguous
+# (mid-shutdown handoff vs crash loop); 3 covers the handoff with margin while
+# a persistently-failing config produces a handful of children, not dozens.
+_START_MAX_SPAWNS = 3
+
+
 @daemon_group.command(name="start")
 def start_cmd() -> None:
     """Spawn the daemon detached if one isn't already running for this config.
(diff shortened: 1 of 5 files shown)


pandas-studio (approved):
LGTM

PR #445, opened 2026-06-11
fix(cli): config-file handling consistency in the mms proxy CLI

memtomem:
## Background

Low findings L204, L208, L212, L216 from the 2026-06-10 implementation review, re-verified against current main — all in `cli/proxy.py`. Note: a verification agent had marked L208/L212 as fixed by #432 with mismatched evidence; direct re-verification showed both still open (#432 touched server config loading, not these CLI paths).

## What changed

- **Atomic `.mcp.json` (L204):** `_write_mcp_json_for_stm` used plain `write_text` while `_save` and `_desktop_json_remove_entry` go through `atomic_write_text` precisely because a concurrent reader (Claude Code at project load) can hit a truncated file. Now atomic.
- **`init --config <custom>` divergence note (L208):** the registered MCP client entry carries no `--config` reference, so the registered server boots reading the DEFAULT config — silently ignoring the file the user just built. The non-default-path hint block now warns about this and names the `MEMTOMEM_STM_PROXY__CONFIG_PATH` env override (verified live: `server.py` loads from `config.proxy.config_path`, which pydantic-settings binds from that env).
- **`remove` missing-config guard (L212):** `prune`/`register` pre-check `resolved.exists()`; `remove` called `_load` directly, so a wrong `--config` path reported `server 'X' not found` instead of `config not found`. Same guard, same wording.
- **Prune reporting helper (L216):** the 18-line "Removed from source client(s) / could not remove N" block duplicated verbatim between `_handle_source_prune` and the `prune` command is now `_report_prune_results(pruned, failed) -> bool`; the command path exits non-zero on `True`. Byte-identical output.

## Tests

- `test_write_mcp_json_failure_leaves_prior_contents` — patched `os.replace` failure leaves the prior file intact, no temp litter (fails pre-fix).
- `test_init_custom_config_warns_registered_entry_reads_default` (fails pre-fix).
- `test_remove_missing_config_reports_config_not_found` (fails pre-fix).
- Helper extraction covered by the existing prune suites.

`tests/cli/test_proxy_cli.py` 224 passed; ruff clean (the file's pre-existing unformatted spots are untouched — CI format gate covers `src` only).

### DIFF
--- a/src/memtomem_stm/cli/proxy.py
+++ b/src/memtomem_stm/cli/proxy.py
@@ -332,8 +332,16 @@ def _write_mcp_json_for_stm(target_dir: Path, server_cmd: str, server_args: list
         servers["memtomem-stm"] = entry
     # Atomic like every sibling config writer (_save, _desktop_json_remove_entry):
     # Claude Code re-reads .mcp.json at project load, and a crash/disk-full
-    # mid-write must leave the prior contents, not truncated JSON.
-    atomic_write_text(mcp_path, json.dumps(existing, indent=2) + "\n")
+    # mid-write must leave the prior contents, not truncated JSON. Unlike
+    # those secret-bearing configs, .mcp.json is a shared project file: keep
+    # an existing file's mode, default new ones to 0o644 — without an
+    # explicit mode the helper's mkstemp temp (0600) survives the rename and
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #446, opened 2026-06-11
fix(mms): reject unknown names in project enable, refuse ambiguous show, drop dead guard

memtomem:
## Background

Low findings L235, L239, L243 from the 2026-06-10 implementation review, re-verified against current main. (L235's earlier agent-triage claimed it fixed by #430 with mismatched evidence; direct verification showed `show_cmd` still took `matches[0]` unguarded.)

## What changed

- **`mms project show NAME` ambiguity (L235):** the projects index is keyed by canonical path, so two projects can legitimately share a name. `enable`/`disable` (via `_resolve_project_for_mutation`) raise "ambiguous (N matches)" for that state, but `show` silently displayed `matches[0]`. It now raises the mirrored error and points at running `show` from inside the target project root.
- **`mms project enable` validation (L239):** any requested name was appended to the enabled list with no registry check — `mms project enable filesytem` (typo) succeeded and persisted a name that resolves to nothing at proxy time. Unknown names now raise a `ClickException` listing them plus the registered set; nothing is persisted. `disable` intentionally stays registry-agnostic (its contract: works against an empty registry).
- **Dead guard (L243):** `_value_looks_secret`'s `not value or` disjunct was unreachable after the `len(value) < 32` floor; dropped, whitespace check unchanged.

## Behavior change

`mms project enable` now exits non-zero on unregistered names instead of silently persisting them. Scripts enabling names before registering them would need to reorder — registering first was already the documented flow (RFC §7.1).

## Tests

- `test_show_duplicate_name_is_ambiguous` (fails pre-fix)
- `test_enable_unknown_name_rejected_with_hint` — asserts the error, the hint, and that nothing was persisted (fails pre-fix)

`test_mms_project.py` + `tests/mms/` + `test_mms_host.py` 284 passed; ruff clean.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -477,7 +477,7 @@ Options:
 
 `enable` adds MCP names to the project's `[mcp].enabled` list; `disable` removes them. Both require either a marker at cwd (or up the tree) or an explicit `--project NAME`. `enable` additionally requires a non-empty `~/.mms/registry.toml` — populate it with `mms import` first. `disable` works regardless of registry state since it only mutates project state.
 
-The MCP-name validity check (does this name actually exist in the registry?) is intentionally deferred to sync time (W2) per RFC §7.1, so `enable` accepts any name as long as the registry is non-empty.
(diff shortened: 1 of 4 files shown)


pandas-studio (approved):
LGTM

PR #447, opened 2026-06-11
fix(server): warn at startup when surfacing is enabled but the proxy is disabled

memtomem:
## Background

Low finding L74 from the 2026-06-10 implementation review, re-verified against current main. The surfacing-engine init block is nested inside `if config.proxy.enabled:` in `app_lifespan`, so `surfacing.enabled=true` with the proxy disabled (the default control-only mode) is silently skipped. The only startup log said nothing about surfacing; the mismatch was observable solely via `stm_proxy_health`.

## What changed

The proxy-disabled branch now emits a WARNING when surfacing is configured-enabled, mirroring the #288 `auto_index`/`extraction` "config is enabled but inert" pattern: names what is inert, the consequence (no memories surfaced), and both remedies (enable the proxy / set `surfacing.enabled=false`).

## Tests

`test_proxy_disabled_warns_when_surfacing_configured_on` — covers both directions: the warning fires with surfacing configured on (fails pre-fix) and the plain proxy-disabled default stays warning-free. `test_server_tools.py` 75 passed; ruff clean.

### DIFF
--- a/src/memtomem_stm/server.py
+++ b/src/memtomem_stm/server.py
@@ -244,15 +244,19 @@ async def app_lifespan(server: FastMCP) -> AsyncIterator[STMContext]:
                 logger.warning("Langfuse init failed, continuing without tracing", exc_info=True)
         else:
             logger.info("Proxy disabled (enabled=false) — only STM control tools available")
-            if config.surfacing.enabled:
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #448, opened 2026-06-11
fix(proxy): degrade on third-party API drift in tool registration and tracing

memtomem:
## Background

Low findings L51 and L70 from the 2026-06-10 implementation review, re-verified against current main (L51 had been mis-triaged as fixed by #412 with mismatched evidence; the unguarded assignments are still at `_fastmcp_compat.py:107-109` on main). Both are "third-party boundary must fail open" gaps. This completes the bucket-1+2 low drive.

## What changed

- **`register_proxy_tool` (L51):** `server.add_tool` and `_tool_manager._tools.get` were already wrapped in version-drift guards, but the `.parameters`/`.fn_metadata` assignments were bare. pydantic v2 raises on setattr to a non-field, and `server.py` loops `register_proxy_tool` with no per-tool guard — one renamed fastmcp field would abort registration of every remaining proxied tool and fail server startup. Now the same try/except + WARNING posture: the tool stays registered with a default schema.
- **`traced()` (L70):** called `client.start_as_current_observation` unguarded while the module contract promises tracing never breaks the traced call (the proxy hot path wraps upstream calls in `with traced(...)` at 7 sites, surfacing at 2). The report's suggested constructor-only guard was marked *fix-incomplete* — Langfuse CMs do their real work in `__enter__` — so the fix adds a `_SafeObservation` delegating wrapper: SDK enter/exit failures degrade to untraced (WARNING), a never-entered inner CM gets no `__exit__`, and traced-body exceptions still propagate.

## Review

Codex R1 NEEDS-FIX (test sampling-rate pollution at source; non-atomic schema override — fn_metadata now written FIRST so a partial failure never advertises a schema the original model rejects; per-call warning spam — once-then-DEBUG with lock) → R2/R3 test-global hygiene → **R4 SHIP**.

## Tests

- `test_register_proxy_tool_degrades_when_schema_fields_renamed` — frozen-tool object raising on setattr; no exception, warning logged, tool stays registered (fails pre-fix).
- `test_traced_degrades_when_client_method_raises` / `..._when_observation_enter_raises` (both fail pre-fix) and `test_traced_body_exception_still_propagates` (semantics pin).
- New tracing tests save/restore `_sampling_rate`, which an earlier init test pollutes with a MagicMock.

`test_fastmcp_compat.py` + `test_stm_remaining.py` 48 passed; `test_server_tools.py` green; ruff clean.

### DIFF
--- a/src/memtomem_stm/observability/tracing.py
+++ b/src/memtomem_stm/observability/tracing.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 import logging
 import os
 import random
+import threading
 from contextlib import nullcontext
 from typing import Any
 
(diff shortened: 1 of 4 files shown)


pandas-studio (approved):
LGTM

PR #449, opened 2026-06-11
refactor(compression): drop _take's unreachable container machinery, unify the truncate fallback

memtomem:
Two cleanups from the 2026-06-10 implementation review (low tier, report items L135 + L139).

## What

**1. `_take` container machinery deleted as unreachable (~100 lines)**

The monotone wave rerouted every container root away from `_take`'s greedy
prefix-fill: dict roots go through `_enrich_dict_monotone` (#407), list roots
and nested containers through `_fit_monotone` (#405). The remaining `_take`
call sites in `_fit_extracted` pass only strings; the defensive tail of
`_fit_extracted` is reachable only by an EMPTY dict (every other shape is
handled earlier) and now floors to `{}` directly — byte-identical to what the
old path produced. `_take` keeps its live leaf contract: string prefix
truncation, scalar stub, fits-whole containers.

Comments that described the deleted search as live behavior are rewritten —
notably the "dict-enrich search relies on `_take`" note, which predated #407
(`_enrich_dict_monotone` never calls `_take`), and the router parenthetical
claiming the dict final tier "still inherits `_take`'s non-monotonicity".

**2. `_plain_truncate` helper**

`TruncateCompressor().compress(...)` was constructed at 13 sites across
`SelectiveCompressor` (1), `LLMCompressor` (6) and `HybridCompressor` (6).
All now route through one module-level helper, so a future change to the
degrade path (shared scorer, clamping) is a one-line edit. Behavior-preserving:
LLM sites keep not forwarding `context_query`; Selective/Hybrid sites keep
forwarding it.

## Tests

- The two direct unit tests of the deleted container machinery are removed;
  their fits-whole contract is folded into
  `test_take_keeps_full_container_when_it_fits` (now with an exact-fit case).
- Full suite: 3101 passed (3103 on main minus the 2 deleted tests). No
  behavior change.

Codex review: SHIP (R1, no findings).

### DIFF
--- a/src/memtomem_stm/proxy/compression.py
+++ b/src/memtomem_stm/proxy/compression.py
@@ -745,6 +745,14 @@ class TruncateCompressor:
         return cut + sentinel + footer
 
 
+def _plain_truncate(text: str, *, max_chars: int, context_query: str | None = None) -> str:
+    """Run a default ``TruncateCompressor`` over ``text``: the single
+    construction site for the plain-truncation degrade path every richer
+    compressor falls back to (and for the ``truncate`` tail mode), so a future
+    change to that path is a one-line edit instead of a many-site sweep."""
+    return TruncateCompressor().compress(text, max_chars=max_chars, context_query=context_query)
+
+
 @dataclass
 class PendingSelection:
     """Stores original chunks while waiting for section selection."""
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #450, opened 2026-06-11
fix(compression): refill the freed budget after _json_key_truncate drops keys

memtomem:
From the 2026-06-10 implementation review (low tier, report item L127).

## Problem

`TruncateCompressor._json_key_truncate` sized each per-key part ONCE against
the FULL key set, then its assembler only dropped whole trailing keys to fit.
When one key's value was large, everything after the drop point was discarded
and the freed budget was never re-spent — the output froze far below
`max_chars`. Reproduced before fixing: for
`{"a": {...}, "b": {"big": "B"*3000}, "c": {...}}` the output was the same 66
chars (key `a` + marker) at every budget from 80 through 1500 (18x waste).

## Fix

- After the whole-key drop loop finds the fitting prefix, the BOUNDARY key
  (first dropped) is grown to the largest `_truncate_json_value` form that
  still fits, found by binary search.
- `_truncate_json_value` now applies the #395 no-expand rule
  (`value[:max(0, budget-3)] + "..."`): the old `value[:budget] + "..."`
  rendered budget+3 chars, so just below a string's full length the truncated
  form was LONGER than the full value — a non-monotone search domain that let
  the binary search discard the fitting full-value region (codex R1 Major,
  reproduced: preserved content shrank 60 → 58 chars from budget 147 → 148).
- The omitted-count marker key is now collision-safe against a real top-level
  `_truncated` key, mirroring the convention everywhere else in the file.

Only one key is ever refilled: a full-form boundary would re-create the
`keep+1` candidate the drop loop already rejected with a smaller-or-equal
boundary, so the boundary always stays partial.

## Verification

Contiguous sweeps over curated fixtures plus 150 random config shapes: output
is always valid JSON, within `max(2, budget)`, and preserved ORIGINAL content
is non-decreasing in the budget. Adjudicated + accepted residual (documented
in the test scope note): on arbitrary shapes the proportional allocator's
integer rounding can trade a few chars between two KEPT keys as the budget
grows (net −2 observed, no keys dropped) — allocation noise inherent to
proportional/BM25-weighted shares, not a cliff.

New tests: coarse freeze regression, boundary refill, contiguous
content-monotonicity sweeps (incl. the ellipsis-crossing shape from codex R1),
marker collision safety. Full suite: 3112 passed.

Codex review: R1 NEEDS-FIX (ellipsis render expansion breaking the search
domain — fixed) → R2 SHIP (exhaustive small-shape search vs brute force).

### DIFF
--- a/src/memtomem_stm/proxy/compression.py
+++ b/src/memtomem_stm/proxy/compression.py
@@ -352,12 +352,12 @@ class TruncateCompressor:
             # max_chars=80 through 1500 on a one-huge-key payload). Grow the
             # boundary to the largest ``_truncate_json_value`` form that still
             # fits: the rendered length is monotone in the value budget, so
-            # binary-search it. Only ONE key is refilled: a full-form boundary
-            # would re-create the ``keep + 1`` candidate this loop already
-            # rejected, so (apart from the few-char window where a truncated
-            # string render exceeds its full form by up to 3 chars) the
-            # boundary stays partial and no room is left for the keys after
-            # it.
+            # binary-search it (sound because ``_truncate_json_value`` never
+            # expands a value, so the rendered length is monotone in the value
+            # budget). Only ONE key is refilled: a full-form boundary would
+            # re-create the ``keep + 1`` candidate this loop already rejected
+            # with a smaller-or-equal boundary, so the boundary always stays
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #451, opened 2026-06-11
fix(cleaning): bound the injection scan honestly — full scan to 20k, head+tail beyond

memtomem:
From the 2026-06-10 implementation review (low tier, report item L316).

## Problem

`_check_injection` silently sliced `text[:10_000]` before scanning, so an
upstream response that placed a manipulation pattern ("ignore previous
instructions", `<system>`) after a large benign prefix produced no warning —
and the truncation was undocumented, letting readers assume full-response
coverage. Detection-only path: the miss costs a log line, not an exploit.

## Fix

- A response up to **20,000 chars is scanned in full** (one window — no seam
  a pattern can straddle).
- A larger response gets the **first and last 10,000 chars** (appending the
  payload after a large benign body is the cheap way around a head-only
  window); an injection buried in its middle stays unlogged as the documented
  cost bound.
- The docstring now states the bound explicitly.

Codex R1 caught a real seam in the first cut (adjacent head/tail windows at
exactly 20k split a straddling pattern so neither sample matched) — fixed by
the full-scan-to-20k rule above, same total cost bound. R2: SHIP.

## Tests

Pins added for: tail coverage after a >10k benign prefix, the
boundary-straddle at exactly 20k, and the documented >20k middle gap.
Full suite: 3106 passed.

### DIFF
--- a/src/memtomem_stm/proxy/cleaning.py
+++ b/src/memtomem_stm/proxy/cleaning.py
@@ -61,15 +61,17 @@ class DefaultContentCleaner:
         """Log a warning if the text contains likely prompt injection patterns.
 
         Detection-only: the output is never altered or blocked, so a miss
-        costs a log line, not an exploit. The scan is cost-bounded to the
+        costs a log line, not an exploit. The scan cost is bounded to 20,000
+        chars: a response up to that size is scanned in FULL (one window, so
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

PR #452, opened 2026-06-11
fix(config): validate hybrid ordering, rescue zero model budgets, name env overrides in load warnings

memtomem:
## Summary

Four config-validation findings from the 2026-06-10 implementation review (lows L262/L266/L270/L274), all in `ProxyConfig` load/validation:

- **L270 — `HybridConfig` cross-field validation**: `min_head_chars > head_chars` previously loaded fine but made `HybridCompressor`'s head-budget guard fire on every call, silently degrading the operator's chosen hybrid strategy to plain truncation. Now rejected at load (`_check_ordering`), mirroring `UpstreamServerConfig`'s dependent numeric-field validators.
- **L274 — zero model budget falls back to the static default**: `context_budget_ratio` is validated `ge=0.0`, so `0` is legal — but it produced a 0-char model budget that flowed into every per-server `max_chars`. A degenerate budget now means "model scaling effectively off", not "no output": fall back to `default_max_result_chars` (itself `gt=0`). The previous behavior was pinned by a characterization test (`test_zero_ratio`), which this PR deliberately flips.
- **L262 — env-override hints in load-failure warnings**: a malformed `MEMTOMEM_STM_PROXY__*` value fails validation of the MERGED config, and the warning blamed the file — the operator debugs a file that is fine. The warning now appends the implicated env var name(s). Attribution is two-staged: a differential `(loc, type, msg)` pre-filter against a file-alone validation (so file-caused errors are never pinned on an innocent env sibling), then naming derived from the env overlay's leaves (never synthesized from the error location). Full contract in the `_env_override_hint` docstring. Fallback semantics are unchanged — this only improves the warning text.
- **L266 — eager `api_key` validation documented as deliberate**: every `llm` block is validated at load, even when nothing selects the `llm_summary` strategy or the block sits under a disabled `extraction`. Reviewed and kept — deferring the check would move a genuinely-enabled compressor's failure from startup to the first tool call. Now stated in a comment at the validator and a callout in `docs/compression.md`.

## Behavior change

- Configs with `hybrid.min_head_chars > hybrid.head_chars` now **fail validation** (previously loaded and silently no-opped to truncation).
- `context_budget_ratio: 0` (or any ratio that floors the model budget to 0 chars) now yields `default_max_result_chars` instead of a 0-char budget.
- Non-object config roots (e.g. a top-level JSON array) are rejected **before** the env merge — previously `[]` slipped through `_deep_merge` (`dict([])` is `{}`) and could validate cleanly under env overrides, silently accepting an invalid file.

## Review

7 codex rounds. R1–R6 each caught a real corner of the env-hint attribution (loc-synthesized names → leaf-derived; missing-field in env-created subtrees → file provenance; `[]` root; `loc=()` root validators → differential validation; file-broken-subtree false positive → generalized `(loc, type)` pre-filter; same-key masking between model validators → key extended with the message). R7: **SHIP**, with one acknowledged non-blocking nit — when the file and an env override break the same leaf with the identical pydantic message, the hint is skipped; refining that comparison risks reintroducing the innocent-sibling misattribution R5 fixed.

## Test plan

- `tests/test_env_overrides.py` — attribution matrix: leaf / subtree / root locations, env-created subtrees, file-broken-sibling pre-filter, `[]` root, `loc=()` root validators, same-location different-message disambiguation
- `tests/test_config_constraints.py` — hybrid ordering rejection + zero-budget fallback (flipped `test_zero_ratio` pin)
- Full suite: 3122 passed, 3 skipped (`-m "not ollama"`); `ruff check` + `ruff format --check` clean

### DIFF
--- a/docs/compression.md
+++ b/docs/compression.md
@@ -264,6 +264,8 @@ Routes through an external LLM for intelligent summarization:
 
 Providers: `openai`, `anthropic`, `ollama`. `llm_timeout_seconds` (default 60.0) bounds the per-call LLM wait. On timeout, privacy-pattern hit, circuit-breaker open, or any other API failure, compression falls back to `TruncateCompressor` and records the strategy as `llm_summary→{timeout,privacy,circuit_breaker,llm_error}_fallback` in `proxy_metrics` for observability. A successful call whose summary still exceeds `max_chars` is clamped by the same truncation and recorded as `llm_summary→llm_overlength_fallback` — the model overshot the length instruction; the endpoint is fine, so the circuit breaker is unaffected.
 
+> **`api_key` is validated eagerly.** Every `llm` block in the config is validated when the config loads — for `provider: openai` / `anthropic`, a missing `api_key` (and missing `OPENAI_API_KEY` / `ANTHROPIC_API_KEY` env var) fails the load even if nothing selects the `llm_summary` strategy or the block sits under a disabled `extraction`. This keeps the startup fail-fast for configs that DO use the LLM. Don't leave placeholder `llm` blocks you aren't using — remove them, or point them at `provider: ollama` (no key required).
+
 Sensitive content (API keys, passwords, PII) is auto-detected and **never** sent to external LLMs — falls back to local truncation.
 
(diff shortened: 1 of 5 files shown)


pandas-studio (approved):
LGTM

PR #457, opened 2026-06-11
fix(mms): serialize sync --apply / import --apply behind a cross-process write lock

memtomem:
## Summary

`mms host sync --apply` and `mms import --apply` both load `~/.mms/registry.toml` + `import_state.toml`, decide, and save both files back. The individual writes are atomic (`atomic_write_text`), but the read-modify-write span is not: two concurrent `--apply` runs interleave, the loser's load goes stale, and its save silently drops the winner's mutation (2026-06-10 implementation review, L231).

- **`state.write_lock()`** — advisory `fcntl.flock` on `~/.mms/.lock` (0600), non-blocking poll with a 10s default timeout (`WRITE_LOCK_TIMEOUT_SECONDS`, monkeypatch-friendly). Timeout raises `WriteLockTimeout`; `flock` auto-releases on process exit, so a crashed holder cannot wedge later runs and no stale-lock cleanup exists or is needed.
- **`cli/_write_lock.py: with_write_lock`** — decorator between the `@click.option` stack and the callback, so the lock spans the ENTIRE command: load → classification → the TTY confirmation prompt → both saves. Releasing around the prompt would re-open the stale-load window the lock exists to close, so the prompt deliberately stays inside. `WriteLockTimeout` surfaces as a clean `ClickException` (exit 1), not a traceback.
- **`--plan` never locks** (`enabled=not is_plan`): plan runs only read, and the mutating side's atomic writes keep readers consistent.
- **Windows**: no `fcntl` → documented no-op (single-user CLI; the lock is best-effort hardening there). The `import fcntl` is `try/except ImportError`-guarded, satisfying `test_no_posix_only_imports`.

The decorator lives in its own module rather than joining the `mms_import` → `mms_host` cross-import: the promotion trigger documented at `_classify_against_registry` fires when the next *domain* helper joins that pair, and this is CLI glue, not sync/import domain logic.

## Behavior change

A second concurrent `--apply` now fails after 10s with `Error: timed out ... waiting for the mms write lock` instead of silently dropping the first run's mutation. `--plan` behavior is unchanged.

## Review

codex R1: **SHIP**, no findings.

## Test plan

`tests/mms/test_write_lock.py` (11 new): contention via a second fd (flock attaches to the open file description, so same-process fds contend like cross-process holders), timeout error content, waiter acquires after release, release/reacquire, `enabled=False` no-ops, held-lock CLI exits cleanly for both commands, `--plan` skips the lock under a held lock, post-wrap mutation still works, `--help` docstrings preserved. Full suite: 3114 passed, 3 skipped; ruff + format clean.

### DIFF
--- a/src/memtomem_stm/cli/_write_lock.py
+++ b/src/memtomem_stm/cli/_write_lock.py
@@ -1,0 +1,49 @@
+"""Write-lock decorator shared by the mutating mms CLI commands.
+
+Lives in its own module rather than joining the ``mms_import`` →
+``mms_host`` cross-import: the promotion trigger documented at
+``_classify_against_registry`` fires when the *next domain helper* joins
+that pair, and this is CLI glue around
+:func:`memtomem_stm.mms.state.write_lock`, not sync/import domain logic.
+"""
+
+from __future__ import annotations
+
+import functools
+from collections.abc import Callable
+from typing import Any, TypeVar
+
+import click
+
+from memtomem_stm.mms import state
+
+F = TypeVar("F", bound=Callable[..., Any])
+
+
+def with_write_lock(f: F) -> F:
+    """Run a Click command callback under the mms cross-process write lock.
+
+    Applied between the ``@click.option`` stack and the callback so the lock
+    spans the ENTIRE command — load, classification, the TTY confirmation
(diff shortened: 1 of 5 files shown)


pandas-studio (approved):
LGTM

PR #458, opened 2026-06-11
fix(stores): chmod feedback/progressive/pending SQLite files to 0600

### LINKED ISSUE #455: Feedback/progressive/pending SQLite files are created world-readable (no 0600 chmod)
## Background

`ProxyCache` and `MetricsStore` chmod their DB files to `0600` right after connect (`cache.py:92-99`, `metrics_store.py:169-176` at `f54b87b`). The other local stores only `mkdir` the parent with `mode=0o700` and never chmod the DB file itself:

- `FeedbackStore` (`surfacing/feedback_store.py:246-252`)
- `CompressionFeedbackStore` (`proxy/compression_feedback_store.py:92-98`)
- `ProgressiveReadsStore` (`proxy/progressive_reads_store.py:64-70`)
- `SQLitePendingStore` (`proxy/pending_store.py:98-100`)

`mkdir(mode=0o700)` only applies when the directory is created, so a pre-existing permissive `~/.memtomem` is left as-is.

## Observed

On a real machine with a pre-existing `~/.memtomem` (`drwxr-xr-x`): `stm_feedback.db` is `-rw-r--r--`, and its `-wal` / `-shm` sidecars are 644 as well. The DB contains surfacing queries (hashed only when they match secret patterns), memory IDs, ratings, compression feedback text, and progressive read telemetry — readable by other local users when both the directory and file modes are permissive.

## Proposed fix

Centralize private SQLite initialization in one helper (6+ call sites, matching this repo's helper-extraction precedent) that applies `chmod(0o600)` to the DB file **and any existing `-wal`/`-shm` sidecars**. SQLite creates sidecars by copying the DB file's mode, so correcting the DB file at open time must also correct sidecars left over from earlier runs. Consider additionally warning when `~/.memtomem` itself is group/world-accessible.

Found by an internal security audit (2026-06-11); file modes confirmed on a real install.

memtomem:
Closes #455.

## Summary

- Extract `utils/sqlite_private.ensure_private_db_files()`: chmods the SQLite DB file **and any existing `-wal`/`-shm` sidecars** to `0600`, best-effort (`OSError` swallowed per path, mirroring the previous inline chmod).
- Wire it into all six SQLite-backed stores, right after `sqlite3.connect`:
  - `ProxyCache`, `MetricsStore` — replace their inline chmod (which covered only the DB file, not sidecars),
  - `FeedbackStore`, `CompressionFeedbackStore`, `ProgressiveReadsStore`, `SQLitePendingStore` — previously no chmod at all.
- Tests (`tests/test_sqlite_private.py`): helper unit tests (sidecar correction, missing files, unrelated files untouched) plus a 6-store parametrized pin — fresh DB is `0600`, and a pre-existing `0644` DB with leftover `0644` sidecars is corrected by `initialize()`.

## Why sidecars matter

SQLite creates `-wal`/`-shm` by copying the main DB file's mode, so a `0644` DB keeps producing `0644` sidecars until the DB itself is corrected — and sidecars left behind by earlier permissive runs are never touched by a DB-file-only chmod. Confirmed on a real install: `stm_feedback.db`, `stm_feedback.db-wal`, and `stm_feedback.db-shm` were all `-rw-r--r--`.

## Behavior change

On startup, every store now corrects the modes of its existing DB file and sidecars to `0600`. Installs with world-readable `stm_feedback.db` (surfacing queries, ratings, feedback text, read telemetry) get corrected at the next start. No schema or API change; failure to chmod never fails store initialization.

## Out of scope

The "warn when `~/.memtomem` itself is group/world-accessible" idea from #455 is intentionally dropped rather than deferred: with all DB files at `0600`, a permissive parent directory discloses file names only, not contents — and the directory is shared with the LTM server, so a per-store warning would nag on every start for marginal benefit. Reopen if file-name disclosure is considered sensitive.

## Verification

- `uv run pytest -m "not ollama"` — 3148 passed, 3 skipped (includes 16 new tests)

### DIFF
--- a/src/memtomem_stm/proxy/cache.py
+++ b/src/memtomem_stm/proxy/cache.py
@@ -12,6 +12,7 @@ from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
 
+from memtomem_stm.utils.sqlite_private import ensure_private_db_files
 from memtomem_stm.utils.sqlite_tuning import tune_connection
 
 logger = logging.getLogger(__name__)
@@ -93,10 +94,7 @@ class ProxyCache:
         self._db_path.parent.mkdir(parents=True, exist_ok=True, mode=0o700)
         db = sqlite3.connect(str(self._db_path), check_same_thread=False, timeout=5.0)
         try:
-            try:
-                self._db_path.chmod(0o600)
-            except OSError:
-                pass
+            ensure_private_db_files(self._db_path)
             tune_connection(db)
             db.execute(_CREATE_TABLE)
             db.execute(_CREATE_INDEX)
--- a/src/memtomem_stm/proxy/compression_feedback_store.py
+++ b/src/memtomem_stm/proxy/compression_feedback_store.py
@@ -19,6 +19,7 @@ import threading
 import time
 from pathlib import Path
 
+from memtomem_stm.utils.sqlite_private import ensure_private_db_files
 from memtomem_stm.utils.sqlite_tuning import tune_connection
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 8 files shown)


pandas-studio (approved):
LGTM

PR #459, opened 2026-06-11
docs(proxy): state the sync-sqlite-on-event-loop invariant at every store write path

memtomem:
## Summary

All five sqlite stores (proxy `MetricsStore`, `ProxyCache`, `ProgressiveReadsStore`, `CompressionFeedbackStore`, surfacing `FeedbackStore`) execute synchronous writes directly on the asyncio event loop. That is a deliberate trade for the current local single-MCP-client deployment — but nothing in the code said so, and a future multi-client change could inherit the blocking writes silently (2026-06-10 implementation review, L293; decided document-only).

Each site now states:
- the real blast radius — while a write runs, **every runnable coroutine stalls** (other in-flight proxied calls included), not just the call that produced it;
- why it is accepted — local single-client deployment, low call volume, cheap local WAL writes vs the upstream calls they account for;
- the reopen trigger — multi-client serving (or materially higher concurrency) → move persistence off-loop (`asyncio.to_thread`);
- per-store lock readiness for that move — `MetricsStore`/`ProxyCache` lock all connection access (lock-ready); the other three lock **writes only**, so a thread move also needs their read/stat paths locked (the `MetricsStore.__init__` reader/writer convention) or per-thread connections.

Comment/docstring-only; no behavior change.

## Review

codex R1 NEEDS-FIX: initial wording overstated the blast radius ("delays only the call being served") and overclaimed `to_thread` readiness for the three write-only-locked stores — both reworded per store after verifying lock coverage. R2: **SHIP**, no findings ("rewording is accurate per store").

## Test plan

Doc-only — full suite 3132 passed, 3 skipped; ruff check + format clean.

### DIFF
--- a/src/memtomem_stm/proxy/cache.py
+++ b/src/memtomem_stm/proxy/cache.py
@@ -83,6 +83,17 @@ def response_carries_transient_key(text: str) -> bool:
 
 
 class ProxyCache:
+    """SQLite-backed cross-restart cache for proxied tool results.
+
+    Every method does synchronous sqlite I/O on the caller's thread — in
+    the proxy that is the asyncio event loop. Deliberate under the proxy's
+    single-MCP-client invariant: a blocking ``get``/``set`` delays only the
+    one in-flight call it serves, and is far cheaper than the upstream
+    call a hit avoids. Multi-client serving is the reopen trigger — then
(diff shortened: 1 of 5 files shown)


pandas-studio (approved):
LGTM

Issue #453, opened 2026-06-11
Sensitive responses are persisted to the response cache and auto-index despite the SECURITY.md exclusion guarantee


memtomem:
## Background

SECURITY.md ("Sensitive content auto-detection") states that responses containing secret-like patterns "are detected and excluded from the response cache and from being indexed into LTM." #289 wired the privacy scan into the LLM compression route only: it prevents the **outbound provider call** (`manager.py:731-740`, `compression.py:2247-2262` at `f54b87b`) and falls back to truncation on a hit. The (possibly truncated) sensitive text then flows on into persistence unchecked — the scan is routing-only, nothing re-checks the value at persistence time.

## Concrete gap

- **Cache write path** checks only for transient retrieval keys, never sensitive content (`manager.py:2046-2072`); `ProxyCache.set()` stores the result verbatim (`cache.py:152-175`). The cache is enabled by default (`CacheConfig.enabled = True`), so this leg is live in every default deployment — a secret-bearing upstream response lands in `~/.memtomem/proxy_cache.db`.
- **Auto-index** writes the full response text *and the raw call arguments* into a Markdown file before indexing, with no `contains_sensitive_content()` gate (`memory_ops.py:128-157`).
- **Extracted facts** persist through the same index path with no gate (outbound side of extraction is tracked separately).

## Reachability

Auto-index and extraction are gated on `index_engine` (`manager.py:1886`, `manager.py:1980`), which the standalone `mms` server does not wire (#288) — those two legs are currently inert in the shipped entry points and apply to embedding hosts. The **cache leg is live by default everywhere**.

## Proposed fix

A common pre-persistence guard shared by the cache store, auto-index, and fact persistence:

- scan the exact value being persisted (the compressed text for the cache; the full Markdown content *including rendered arguments* for auto-index),
- on a hit: skip persistence, record an observable skip reason (metrics/log), and return the tool response to the agent unchanged,
- regression tests: cache skip and auto-index skip on the default privacy patterns.

## Why one issue, possibly two PRs

The cache guard (live, default-on) can ship first; the auto-index/extraction-persistence guard touches the currently-inert path and can follow. Both implement the same SECURITY.md sentence, so they share this issue.

Found by an internal security audit (2026-06-11). Line references are at `main` `f54b87b`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

memtomem:
All legs of this issue are fixed on `main` (`dbfc6f9`):

[...]

PR #460, opened 2026-06-11
fix(cache): never persist secret-looking responses to the response cache

memtomem:
Part of #453 (cache leg — the default-on path; auto-index/extraction legs follow separately).

## Summary

- `ProxyCache.set()` drops the row when `contains_sensitive_content(result)` matches (debug-logged with server/tool, never the content; the response to the agent is unaffected). Enforced at the store chokepoint so no caller can bypass it — the transient-key check stays in the manager because it is a pipeline-freshness concern, while this is a store-confidentiality invariant.
- `ProxyCache.get()` mirrors the gate on the read path: a sensitive row that never went through `set()` (written by an older still-running pre-gate process, or an external SQL writer after the startup purge ran) is refused and physically evicted. The sensitivity check runs **before** the expiry check so an already-expired secret row is deleted rather than left resting on disk. Eviction is best-effort: if a concurrent writer holds the file lock, `get()` warns and serves a miss instead of aborting the caller's request; the row is retried on the next `get()` and swept at the next startup.
- `ProxyCache.initialize()` purges matching rows written by pre-gate releases, mirroring the transient-key legacy purge. Purge, write gate, and read guard all share `contains_sensitive_content`, so they cannot diverge.
- Tests (11): parametrized gate pins (api_key/password/sk-token/AKIA/private-key header), clean-row contrast, an explicit email-PII pin (see below), legacy-purge round-trip, post-startup raw-SQL insert refused + evicted, expired-sensitive eviction, and a blocked-eviction scenario under a held `BEGIN IMMEDIATE` write lock.

## Behavior change

- Secret-looking upstream responses are no longer cached — each repeat identical call re-runs the pipeline. A false positive costs one un-cached response, never correctness.
- The gate scans the full `DEFAULT_PATTERNS` (credentials + PII): per the #461 storage-vs-action split, the cache is a STORAGE consumer — an email is fine to show an external summarizer but not fine to persist. **Email-bearing responses are deliberately uncacheable**, pinned by a dedicated test so any future pattern change is a conscious decision.
- On startup, rows matching the patterns are deleted from existing `proxy_cache.db` files. The scan runs every startup (like the marker purge); cost is bounded by `max_entries` (default 10k rows) and is a no-op steady-state.

## Verification

- `uv run pytest -m "not ollama"` — full suite green at each round (latest local run: 3143 passed, 3 skipped)

### DIFF
--- a/src/memtomem_stm/proxy/cache.py
+++ b/src/memtomem_stm/proxy/cache.py
@@ -131,8 +131,9 @@ class ProxyCache:
             # scan and the gate share ``contains_sensitive_content`` so they
             # can never diverge. Runs on every startup (matching the marker
             # purge) — after the first pass it finds nothing, because ``set()``
-            # refuses new matching rows; the rescan also covers rows written by
-            # older processes still running pre-gate code against this file.
+            # refuses new matching rows. Rows written LATER by an older
+            # still-running pre-gate process are outside this purge's reach;
+            # the read-side guard in ``get()`` refuses and evicts those.
             stale_keys = [
                 key
                 for key, result in db.execute("SELECT cache_key, result FROM proxy_cache")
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

Issue #454, opened 2026-06-11
FactExtractor sends raw response text to remote LLM providers without the privacy scan


memtomem:
## Background

LLM compression honors `privacy_scan_enabled` (default `True`) and skips the provider call when the response matches secret patterns (`manager.py:731-740`, `compression.py:2247-2262` at `f54b87b`). Fact extraction reuses the same `LLMCompressorConfig` (`ExtractionConfig.llm`) but never consults `privacy_scan_enabled` — `contains_sensitive_content` has no call site in `extraction.py`.

## Concrete gap

`FactExtractor._extract_llm()` calls `_call_api(text)` directly (`extraction.py:223-224`); `_openai()` / `_anthropic()` send the raw text as the user message (`extraction.py:272-281`, `extraction.py:302-310`). With extraction configured against OpenAI, Anthropic, or a remote custom `base_url`, a secret-bearing upstream response crosses the network/provider trust boundary — more severe than local persistence.

## Reachability / severity

Extraction is gated on `index_engine` (`manager.py:1980`), which the standalone `mms` server does not wire (#288) — the path is **inert in the shipped entry points today** and applies to embedding hosts that wire an index engine. Filed now because it silently diverges from the compression-path contract and becomes live the moment extraction is wired.

## Proposed fix

Run the same privacy scan before `_call_api()`; on a hit, fall back to heuristic extraction (or no extraction) with an observable fallback reason, mirroring `LLMCompressor.compress` (`compression.py:2253-2262`). Add a test proving `_call_api` is not invoked for secret-bearing text.

Note: heuristic-fallback output still persists into LTM — the persistence side is covered by the pre-persistence guard in #453.

Found by an internal security audit (2026-06-11). Line references are at `main` `f54b87b`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #462, opened 2026-06-11
fix(memory-ops): never persist secret-looking content via auto-index or extraction

memtomem:
Part of #453 (auto-index + extraction-persistence legs; the cache leg is #460).

## Summary

- `index_content_matches_privacy()` — a shared predicate scanning every caller-controlled field that reaches the persisted markdown (server/tool names, rendered arguments, agent intent, response text; the static scaffolding cannot match). Used by both in-function gates **and** the manager's background pre-check so they can never diverge.
- `auto_index_response()` gates on the predicate before anything touches disk; on a hit it records `privacy_skip`, logs at info (server/tool only, never content), and returns the agent summary unchanged.
- **Background auto-index pre-checks in the manager** before composing the `[Indexing…] · scheduled` footer: a secret-bearing response gets the un-footered response, no task is scheduled, and the same attempt + `privacy_skip` counters are recorded — the footer never promises an indexing run the task's gate would decline.
- `extract_and_store()` gates twice: the input scan runs before `extractor.extract()` (the possibly-remote extraction LLM never sees the content), and **every candidate fact markdown is built and scanned before any disk/dedup/index interaction** — the extractor can reassemble sensitive content that evaded the input scan. One hit terminally skips the whole batch (no partial writes), including the `index_engine=None` file-write path.
- `_render_args()` shared by the markdown builders and the gates; `privacy_skip` joins the `IndexOutcome` label set (per call, terminal, fires from both paths) — `stm_index_stats` renders labels dynamically, so only docstrings change there.

## Pattern-set choice (post-#461)

Persistence gates scan the full `DEFAULT_PATTERNS` (credentials + PII), per the #461 storage-vs-action split: an email is fine to show an external summarizer but not fine to store. Pinned by a dedicated email test. The action-side gate for the extraction LLM call itself (credentials-only, mirroring compression routing) is #454 and lands separately.

## Reachability

Both legs are gated on `index_engine` (#288) and therefore inert in the standalone `mms` server today — this implements the documented SECURITY.md contract for embedding hosts and any future index-engine wiring.

## Behavior change

For deployments with an index engine wired: secret-bearing responses (or calls whose arguments/intent carry secrets) no longer produce LTM chunks or fact files; background mode returns the response without the scheduled footer instead of scheduling a doomed task; operators see `privacy_skip` in `stm_index_stats`. No schema or API change — `AutoIndexOutcome`/`ExtractOutcome` shapes are unchanged (`ok=True`, zero counts on skip).

## Verification

- `uv run pytest -m "not ollama"` — 3166 passed, 3 skipped (11 new tests)

### DIFF
--- a/src/memtomem_stm/proxy/index_observability.py
+++ b/src/memtomem_stm/proxy/index_observability.py
@@ -49,9 +49,14 @@ from typing import Literal
 # - ``error``: per error event. Outer ``extractor.extract`` failure counts
 #   once (per call); per-fact ``index_file`` failure counts per fact;
 #   auto_index ``index_file`` failure counts once per call.
+# - ``privacy_skip``: per call, terminal — the content matched a privacy
+#   pattern before anything was written (#453), so no file and no
+#   ``index_file`` call exist for this attempt. Fires from both paths:
+#   auto_index scans the exact markdown it would persist; extract scans
+#   the response text + rendered arguments before the extractor runs.
 #
-# Both write paths share this 4-label outcome set deliberately. auto_index
-# only fires ``stored`` and ``error``; ``dedup_skip`` and
+# Both write paths share this 5-label outcome set deliberately. auto_index
+# only fires ``stored``, ``error``, and ``privacy_skip``; ``dedup_skip`` and
 # ``extracted_zero_facts`` are extract-specific. Operators read the raw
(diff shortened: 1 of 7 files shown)


pandas-studio (approved):
LGTM

PR #463, opened 2026-06-11
fix(extraction): scan for credentials before the extraction LLM call

### LINKED ISSUE #454: FactExtractor sends raw response text to remote LLM providers without the privacy scan

memtomem:
Closes #454.

## Summary

`FactExtractor` now mirrors the LLM-compression action gate: when `privacy_scan_enabled` (default on, via the shared `LLMCompressorConfig`) and the response text matches `CREDENTIAL_PATTERNS`, the provider call is skipped with an info log and heuristic extraction runs locally instead. The gate sits in `extract()` **before the `max_input_chars` slice** and before the strategy dispatch, so it scans the full pre-truncation text and covers both LLM-bound strategies (LLM and HYBRID); the provider still receives only the truncated text. As in `LLMCompressor.compress`, the gate precedes the circuit-breaker check.

## Review R1 fixes

- **Blocker — slice-before-scan bypass**: the first revision scanned inside `_extract_llm()`, i.e. after `extract()` had truncated to `max_input_chars`. A credential straddling the boundary shipped a near-complete secret prefix that no longer matched any pattern (`max_input_chars=30`, `"x"*10 + " AKIA" + "A"*16` → the truncated text holds AKIA + 15 of the key's 16 chars, unmatched). The gate now scans the unsliced text; the regression test pins the boundary-split shape itself (sliced text must NOT match, full text must) so a future pattern change cannot quietly turn the test vacuous.
- **Major — persistence claim ahead of its base**: the comment citing memory_ops persistence gating predated #462 on this branch's old base. Rebased onto main with #462 merged; the comment now cites what #462 actually scans (extraction input + every candidate fact file before disk).

## Pattern-set choice (post-#461)

Credentials only — this is an ACTION gate ("may this text be sent to an external provider?"), so it uses `CREDENTIAL_PATTERNS`, exactly like compression routing after #461. An email alone does **not** block the provider call (pinned by test); persistence of extracted output is the STORAGE side and is gated separately in memory_ops (#462, full `DEFAULT_PATTERNS`).

## Reachability

Inert in the standalone `mms` server today (#288, no `index_engine` wired) — and in the wired manager path, the #462 input gate already blocks secret-bearing text before the extractor runs. This PR hardens the `FactExtractor` API boundary itself: embedding hosts and direct library users get the same provider-trust-boundary guarantee the compression route has, instead of silently diverging from it.

## Tests

- credential-bearing text → `_call_api` never awaited, heuristic facts still produced locally
- credential split at the `max_input_chars` boundary → gate still fires (scan-before-slice regression)
- email-only text → `_call_api` awaited (credentials-only pin)
- `privacy_scan_enabled=False` → `_call_api` awaited (operator opt-out honored)
- HYBRID strategy → gate covers it

## Verification

- `uv run pytest -m "not ollama"` — 3198 passed, 3 skipped (5 new tests)
- `uv run mypy src/memtomem_stm/proxy/extraction.py` — clean

### DIFF
--- a/src/memtomem_stm/proxy/extraction.py
+++ b/src/memtomem_stm/proxy/extraction.py
@@ -208,6 +208,27 @@ class FactExtractor:
             return []
         if strategy == ExtractionStrategy.HEURISTIC:
             return _extract_heuristic(truncated, max_facts=self._cfg.max_facts)
+
+        # #454: same action gate as the LLM compression route (#289, credential
+        # set per #461) — a credential-bearing response must not cross the
+        # provider trust boundary. The scan runs on the UNSLICED text: slicing
+        # to max_input_chars can split a credential at the boundary so the
+        # truncated text no longer matches a pattern while a near-complete
+        # secret prefix still ships. Sitting before the strategy dispatch it
+        # covers both LLM-bound strategies (LLM and HYBRID). Heuristic
+        # extraction runs locally instead; persistence of its output is
+        # separately gated in memory_ops (#462 scans the extraction input and
+        # every candidate fact file before disk). Operators disable per-config
+        # when the provider is local/trusted (``privacy_scan_enabled=False``).
+        if self._llm_cfg.privacy_scan_enabled and contains_sensitive_content(
+            text, CREDENTIAL_PATTERNS
+        ):
+            logger.info(
+                "Sensitive content detected, skipping LLM extraction (provider=%s)",
+                self._llm_cfg.provider.value,
+            )
+            return _extract_heuristic(truncated, max_facts=self._cfg.max_facts)
+
         if strategy == ExtractionStrategy.HYBRID:
             return await self._extract_hybrid(truncated, server=server, tool=tool)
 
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

memtomem:
Internal review round 1 — two findings, fix planned (next session):

1. **Blocker — slice-before-scan bypass.** `extract()` truncates to `max_input_chars` before the gate scans, so a credential split exactly at the slice boundary evades `CREDENTIAL_PATTERNS` while a near-complete prefix still ships to the provider. Fix: scan the unsliced input before truncation + a boundary-split regression test.
2. **Major — sequencing.** The comment claiming heuristic-fallback persistence "is separately gated in memory_ops" is only true once #462 merges (this branch is based on plain main). Plan: merge #458/#460/#462 first, rebase this branch, then fix (1) and re-review.

PR #464, opened 2026-06-11
fix(memory-ops): slug server/tool in persisted filenames, private dir and file modes

### LINKED ISSUE #456: auto_index_response trusts the raw server name in file paths and creates memory_dir with default mode
## Background

`auto_index_response()` sanitizes `/` in the tool name but interpolates the `server` name raw into the file name (`memory_ops.py:120-126` at `f54b87b`):

```python
safe_tool = tool.replace("/", "_")
fname = f"{server}__{safe_tool}__{ts}.md"
```

`memory_dir.mkdir(parents=True, exist_ok=True)` also uses the default mode (no `mode=0o700`), unlike the SQLite store parents.

## Impact

Low. Server keys come from trusted config/import flows, auto-index is opt-in, and the path is inert in the standalone `mms` server (#288, `index_engine` gate). This is hardening against hand-edited configs with path separators (or `..`) in server keys making the auto-index write target unpredictable.

## Proposed fix

Slug both `server` and `tool` before path construction, and create the auto-index directory with an explicit private mode. Can ship together with the auto-index leg of #453 (same file) or standalone.

Found by an internal security audit (2026-06-11).

memtomem:
Closes #456.

## Summary

Two hardening fixes for the markdown persistence paths in `memory_ops.py`, completing the security-audit drive (#453–#456):

1. **Filename slug for `server` and `tool`.** `auto_index_response` sanitized `/` in the tool name but interpolated the server key raw into `f"{server}__{safe_tool}__{ts}.md"`; the fact-file name in `extract_and_store` had the same shape. A shared `_path_slug()` now keeps `[A-Za-z0-9._-]` and replaces everything else with `_` for both fields at both sites — with separators gone, dots cannot climb, and the composed name never collapses to `.` or `..`, so the write target always stays inside `memory_dir`.
2. **Private modes.** Both `memory_dir.mkdir(...)` calls get `mode=0o700` (matching every SQLite store parent after #458), and both `atomic_write_text(...)` calls get `mode=0o600`.

## Why the file mode too (not just the dir, as the issue proposed)

Same reasoning `utils/sqlite_private.py` documents for the DB files: `mkdir(mode=0o700)` only applies when the directory is **created** — a `memory_dir` that already exists permissive (created by an older version at the umask default) would leave fresh markdown files world-readable. The persisted content is the user's tool-response/conversation data; `atomic_write_text` already supports `mode=` (chmod on the temp before the rename, so the file is never observable permissive). One extra argument per call site closes the gap completely.

## Behavior change

- New auto-index/fact markdown files are `0o600`; `memory_dir` is created `0o700`. Pre-existing files and dirs are not re-chmod'd (creation-time policy, like #458's mkdir half — the md-file population is unbounded, and unlike SQLite sidecars there is no single fixed path to correct at open).
- Filenames change **only** for server/tool names containing characters outside `[A-Za-z0-9._-]` (previously only `/` in `tool` was replaced). For well-formed names the filenames are byte-identical. Nothing parses these filenames back (verified: no consumer splits on `__`).

## Reachability

Low severity as triaged in #456: server keys come from trusted config/import flows, auto-index is opt-in, and the path is inert in the standalone `mms` server (#288, no `index_engine`). This hardens against hand-edited configs.

## Tests

- server `"../../tmp/evil"` → file lands **inside** `memory_dir` with the slugged name, both for auto-index and fact files
- fresh `memory_dir` has no group/other bits (umask-proof assert); written files are exactly `0o600` — POSIX-only via the repo's `_skip_on_windows` idiom (NTFS doesn't enforce mode bits); the traversal tests stay cross-platform
- existing `tool` slash test unchanged (slug is a superset of the old replace)

## Verification

- `uv run pytest -m "not ollama"` — 3197 passed, 3 skipped (4 new tests)
- `uv run mypy src/memtomem_stm/proxy/memory_ops.py` — clean

### DIFF
--- a/src/memtomem_stm/proxy/memory_ops.py
+++ b/src/memtomem_stm/proxy/memory_ops.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+import re
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any
@@ -30,6 +31,23 @@ def _render_args(arguments: dict[str, Any]) -> str:
     value and the persisted value can never diverge.
     """
     return ", ".join(f"{k}={v!r}" for k, v in arguments.items()) if arguments else "(none)"
+
+
+_PATH_SLUG_RE = re.compile(r"[^A-Za-z0-9._-]")
+
+
+def _path_slug(name: str) -> str:
+    """Collapse a server/tool name into one filesystem-safe path component.
+
+    Server keys and tool names come from trusted config / import flows, but
+    a hand-edited config can carry path separators (or a bare ``..``) that
+    make the write target under ``memory_dir`` unpredictable (#456). The
+    slug keeps ``[A-Za-z0-9._-]`` and replaces everything else with ``_``:
+    with separators gone, dots cannot climb directories, and the composed
+    ``{server}__{tool}__{ts}`` filename never collapses to ``.`` or ``..``.
+    Empty input becomes ``_`` so the filename keeps its field structure.
+    """
+    return _PATH_SLUG_RE.sub("_", name) or "_"
 
 
 def index_content_matches_privacy(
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):
LGTM

Issue #469, opened 2026-06-11
Learning-to-rank over eligible candidates only (gated on offline eval)


memtomem:
**Stage 5 of the tool-selection track. GATE: #468-style offline eval baseline exists and heuristic tuning has plateaued — do not start before.**

Train a feature-based ranker (LTR) on logged outcomes:
- Input features: relevance score + toolgraph `rank_features` facts (drift/unmapped/unbacked/classification/risk_score) + execution stats.
- **Scope guard (toolgraph ADR-0005)**: the model ranks ONLY the post-hard-filter eligible set. `NOT_GRANTED`, `violation`, drifted are rejected outside and before the model; no score can resurrect them.
- Offline evaluation first (same replay harness), shadow mode before live.

**Acceptance**: beats tuned heuristic on replay metrics with confidence; shadow-mode parity check; rollback = flip `ranker_version`.

memtomem:
## Readiness audit (2026-08-18) — what the gate still needs, measured against `main` (95d02e4)

Re-checked this issue's prerequisites against the code rather than against the plan text, so the next person to pick it up does not have to re-derive them. The infrastructure is further along than the issue body implies; the gap is narrower and more specific.

### Already in place

- **`ranker_version` cohort stamping is fully plumbed** on *both* halves of the selection/execution pair (`src/memtomem_stm/proxy/selection_log.py:88-95`), with three real versions defined at `src/memtomem_stm/proxy/tool_relevance.py:62-80` (`v1-bm25-tool-relevance`, `v2-bm25-risk-penalty`, `v3-bm25-graph-risk-penalty`). The rollback flag named in Acceptance already exists as a telemetry field.
- **The offline eval harness exists** (#468, delivered via #698): `src/memtomem_stm/proxy/selection_eval.py`, driven by `mms selection replay`, with a labelled corpus (`src/memtomem_stm/data/selection_eval_v1.json`) and a `review_risk_penalty` × `graph_risk_scale` grid sweep reporting hit@k / MRR / nDCG@5 / forbidden-exposure counts.
- **Execution outcomes are joinable**: `log_execution` (`selection_log.py:259-286`) records `ok`, `latency_ms`, `error_type`, `cache_hit`, left-outer-joined by `selection_id`. A weak implicit relevance label — which advertised candidate the client model actually called — is available today.
[...]

Issue #468, opened 2026-06-11
Offline eval: replay selection logs, tune heuristic weights (gated on telemetry)


memtomem:
**Stage 4 of the tool-selection track. GATE: do not start until #467 telemetry has accumulated real selection logs.**

Replay harness over the JSONL selection log:
- Metrics: task success rate, unsafe-candidate reject rate, unnecessary-rejection rate, latency, retry count, user-correction rate, policy-violation count.
- Compare ranker variants offline (`ranker_version` A/B on the same replayed contexts).
- Output: tuned weights for the rule-based score table — still human-readable, still deterministic.

This is the stage that distinguishes "improvement" from "luck" before any learning is attempted (report: 지속 개선 2단계).

**Acceptance**: fixed test set + replay produces stable metric reports; a weight change shows its metric delta before rollout.

Issue #466, opened 2026-06-11
Relevance / task-match candidate ranking over proxied tool descriptions


memtomem:
**Stage 2 of the tool-selection track.** Recall stage: rank candidate tools by task relevance BEFORE the toolgraph hard filter prunes them (filter is eligibility, not relevance — per memtomem/toolgraph ADR-0005 this scoring lives harness-side, never in toolgraph).

- Reuse mem2mem's existing hybrid search (embedding + keyword) over proxied tool names/descriptions/schemas — no new search infra.
- Output: top-N candidates + scores, handed to the selector adapter (Stage 1 issue) for filtering, then merged: `final = relevance_rank(eligible_only)`.
- Rule-based combination first (e.g. relevance score + toolgraph `risk_score` penalty table from the report) — no learned weights yet.

**Independent of** toolgraph timelines; can ship before Stage 1 as plain ranked exposure.

**Acceptance**: deterministic, reproducible ranking given same inputs; rank features logged per selection (feeds the telemetry issue).

memtomem:
Two things to pin down before this stage starts:

1. **Which "existing hybrid search" to reuse.** What's in-repo today: `proxy/relevance.py` has `BM25Scorer` + `EmbeddingScorer`, but they're wired for memory-section ranking inside selective/hybrid compression — pointing them at tool names/descriptions/schemas is new (small) wiring, not zero work. The other reading — reusing LTM `mem_search` — would mean indexing tool metadata into the LTM store, a much bigger change that couples tool selection to LTM. Proposal: reuse the in-repo scorers; keep LTM out of it.

2. **When does ranking happen?** MCP tool advertisement is per-session, but task relevance is per-request — at exposure/registration time there is no task yet. Options: (a) dynamic re-advertisement via `tools/list_changed`, driven by recent call context (the same signal source surfacing uses); (b) a selection meta-tool the client calls; (c) v0: compute and log the ranking per call as telemetry input only (feeds #467/#468) without changing exposure at all. (c) is the cheapest deterministic start and doesn't bet on client `list_changed` support.

Neither question blocks #467, so starting there.

memtomem:
**Design decision (maintainer-approved, after an independent design review):**

**Q1 — RESOLVED: reuse the in-repo scorers; BM25-only in v0; LTM stays out.**
[...]

Issue #465, opened 2026-06-11
Selector adapter: toolgraph hard filter at proxy tool-exposure time


memtomem:
**Stage 1 of the tool-selection track** (memtomem/toolgraph `docs/adr/0005-selector-surface-boundary.md`, `docs/context-engineering-tool-selection-report.md` — PR memtomem/toolgraph#7).

The STM proxy is the tool-exposure choke point: it decides which upstream tools the agent sees. Integrate toolgraph's deterministic selector surface there:

- On proxy tool registration / exposure, call `eligible_tools(agent, candidates, profile)` and expose only eligible tools; attach reject reasons to logs.
- Per-agent profile config (`strict` default for production, `review`/`explore` opt-in).
- Feature lookups via `rank_features` batch call — never per-candidate `check_access` on the hot path.
- Cache graph-derived features keyed by `graph_generation` (memtomem/toolgraph#13); invalidate on generation change.
- Hard boundary: a learned/heuristic ranker may never override a hard reject (NOT_GRANTED / violation / drifted).

**Depends on**: memtomem/toolgraph#15 (selector surface), memtomem/toolgraph#13 (graph_generation). A prototype against today's `check_access`/`unsafe_callable_tools` MCP tools is fine behind a feature flag.

**Acceptance**: toolgraph's regression criteria pass through the proxy (ambiguous bare names never auto-exposed; violation rows hidden under `strict`; drifted tools excluded in all online profiles); this closes toolgraph#15's first-consumer gate.

memtomem:
Dependency status check (2026-06-11): toolgraph#15 and toolgraph#13 are both open / not started, and ADR-0005 + the report currently live in the still-open toolgraph PR #7 — so full integration is blocked upstream for now.

STM-side prep is cheap, though: the proxy already has a per-tool exposure gate (the `hidden` override applied in `get_proxy_tools()`, `proxy/manager.py` + `ToolOverrideConfig`), which is the natural seam for `eligible_tools`. When this stage starts, the feature-flagged prototype against today's `check_access` / `unsafe_callable_tools` MCP tools is the route we'll take, as the issue suggests — MCP-only integration also keeps the "no Python-level dependency on core" invariant intact.

Sequencing per the track: #467 (telemetry) goes first — it's dependency-free and unlocks toolgraph#16.

memtomem:
[...]

PR #471, opened 2026-06-11
feat(proxy): opt-in selection/execution telemetry JSONL sink (#467)

### LINKED ISSUE #467: Selection + execution telemetry: in-path logging with redaction
**Stage 3 of the tool-selection track** — the proxy is in the call path, so it can record what an advisory analyzer (toolgraph) never can: what was actually called and how it went.

Append-only JSONL sink (schema per the report's Execution-telemetry section):
- selection event: `candidate_tools`, `selected_tool`, per-candidate graph features, `reject_reasons`, `ranker_version`, `graph_generation`
- execution outcome: success/error, latency_ms, retry_count, cost
- feedback: user_corrected, operator_override
- **Redaction policy**: no raw prompts, no secret-bearing resource URIs, no auth headers; task text as hash/redacted summary only.

**Why load-bearing**: this unlocks memtomem/toolgraph#16 (telemetry design gate) and is the prerequisite for every learning stage. Without it, offline eval / LTR / bandit are unfounded.

**Acceptance**: replayable log (stable schema + versioned), redaction tests, sampling/rotation config.

memtomem:
Closes #467 (Stage 3 of the tool-selection track).

## What

An append-only JSONL telemetry sink recording, per proxied call, a `selection` event (which tool the client model called, out of which advertised candidate set) and a paired `execution` event (ok/error, proxy-side latency). New module `proxy/selection_log.py`, config block `proxy.selection_telemetry`, wired in `ProxyManager.call_tool` and the server lifespan. **Off by default** — it is a new disk write path, so the operator opts in explicitly.

## Why now

The proxy is the only component in the call path; nothing persisted selection outcomes until now. Offline eval (#468) and the learning stages (#469/#470) are gated on these logs existing and accumulating, and the STM-native hard filter (#465) needs a place for its reject reasons to land. This also opens toolgraph#16's design gate (a real consumer producing selection logs).

## Schema v1 (stable + versioned — acceptance)

- One JSON object per line, keys sorted; every record carries `schema_version` (exact per-event key sets are **pinned by tests** — any shape change fails `test_selection_and_execution_key_sets_pinned` and forces a version-bump decision) and `ranker_version` (`"v0-passthrough"`: no in-proxy ranking exists yet, so this marks the unranked baseline for replay comparisons).
- `selection_id` joins the pair; `trace_id` joins `proxy_metrics.db` for typed error categories and stage timings instead of duplicating them into the log.
- Reserved-null fields (`reject_reasons`, `candidate_features`, `graph_generation`, `retry_count`, `cost`, `cache_hit`) keep the v1 shape stable when #465 / toolgraph#13/#15 populate them.
- `feedback` event (`user_corrected` / `operator_override`) is schema-pinned with an emitter-less `log_feedback` API — nothing in the proxy produces that signal today; emitters arrive with their signal sources.

## Redaction (acceptance)

Structural, not filter-based: no field ever carries raw arguments, results, prompts, resource URIs, or error message strings. Arguments appear only as canonical-JSON sha256 + char count; error detail beyond the exception class name stays in `proxy_metrics.db` behind the `trace_id` join (same reasoning as #462's error-text handling). Backstop: every serialized line is screened with `contains_sensitive_content` (the `privacy.py` storage-gating rule) and dropped + counted on match — the never-persist posture of #460/#462. Files are created `0600` under a `0700` parent (#458); rotation preserves the mode.

## Sampling / rotation (acceptance)

`sample_rate` applies to the selection+execution pair atomically at selection time (no orphan halves from sampling; redaction-backstop drops are per-record by design — documented as a left-outer join contract). Size-based rotation at `max_bytes` keeping `max_backups` files; `0` truncates.

## Behavior change

- New opt-in write path: with `selection_telemetry.enabled = true`, the proxy appends 2 JSONL records per proxied call to `~/.memtomem/stm_selection_log.jsonl` (0600). With the default `enabled = false` there is **no behavior change** — `ProxyManager` takes an optional `selection_log` that defaults to `None` and the call path is telemetry-free.
- The write is a synchronous file append on the event loop, same accepted contract (and same documented reopen trigger) as `MetricsStore.record`.
- Telemetry failures never propagate: write errors are counted and logged, the proxied call proceeds untouched (covered by tests, including append-to-a-directory).

## Tests

- 22 new tests in `tests/test_selection_log.py`: exact key-set pinning per event type, pair `selection_id`/`trace_id` joining, reserved-field nullness, canonical-hash redaction (secret-bearing args unfindable on disk), sensitive-line backstop drop + counter, deterministic seeded-rng sampling, rotation bounds + zero-backup truncation, 0600/0700 modes (POSIX-only, Windows-skipped per repo convention), write-failure isolation, and ProxyManager wire-in with **counter snapshots** (`events_written` / `events_sampled_out` / `write_errors`) on success, upstream-error, sampled-out, and telemetry-disabled paths.
- `tests/test_server_tools.py` lifespan mocks gain `selection_telemetry.enabled = False`; `tests/test_docs_sync.py` allowlists the new `ProxyManager` kwarg.
- Full suite: 3224 passed, 3 skipped (`-m "not ollama"`); `ruff check` + `ruff format --check src` clean.

## Follow-ups (not in this PR)

- Surfacing telemetry counters via `stm_proxy_stats` (operator visibility).
- `cache_hit` population (needs a hit/miss signal at the `call_tool` boundary).
- #465 hard filter writing real `reject_reasons`; #466 ranking stamping a real `ranker_version`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,7 @@ A second tier of management lets you decide *which MCP servers a given project s
 | [Compression](https://github.com/memtomem/memtomem-stm/blob/main/docs/compression.md) | All 10 strategies — pick the right one for your content |
 | [Caching](https://github.com/memtomem/memtomem-stm/blob/main/docs/caching.md) | Skip repeated work with response caching |
 | [Configuration](https://github.com/memtomem/memtomem-stm/blob/main/docs/configuration.md) | Tune settings without touching code |
+| [Selection telemetry](https://github.com/memtomem/memtomem-stm/blob/main/docs/selection-telemetry.md) | Opt-in JSONL log of tool selection + execution outcomes |
 | [CLI](https://github.com/memtomem/memtomem-stm/blob/main/docs/cli.md) | CLI commands, host sync, hooks, daemon, and MCP tools |
 
 STM advertises four model-facing MCP tools by default. Eight observability and
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -298,9 +298,24 @@ Full example with all options:
   "progressive_reads": {
     "enabled": true,
     "db_path": "~/.memtomem/stm_feedback.db"
+  },
+  "selection_telemetry": {
+    "enabled": false,
+    "path": "~/.memtomem/stm_selection_log.jsonl",
+    "sample_rate": 1.0,
+    "max_bytes": 50000000,
+    "max_backups": 3
   }
 }
 ```
+
+`selection_telemetry` (off by default) appends one `selection` + one
+`execution` JSONL record per proxied call — which tool the client picked out
+of the advertised candidate set, and how the call went. The schema, the
+redaction policy (no raw arguments, results, or error text ever reach the
+log), and the replay-join contract are documented in
+[selection-telemetry.md](selection-telemetry.md). The flag is read at
+startup, like `metrics.enabled` — toggling it requires a restart.
 
 ### Token-equivalent budgets (CJK / non-Latin workloads)
 
--- a/docs/selection-telemetry.md
+++ b/docs/selection-telemetry.md
@@ -23,11 +23,14 @@ Off by default — it is a new disk write path, so the operator opts in:
 ```
 
 The flag is read at startup (like `metrics.enabled`); toggling it requires a
-restart. Files are created `0600` with a `0700` parent directory. When the
-log reaches `max_bytes` it rotates (`log → log.1 → … → log.N`, oldest
-dropped; `max_backups: 0` truncates instead). `sample_rate` keeps the given
-fraction of calls and applies to the selection+execution pair atomically, so
-sampling never produces orphan halves.
+restart. Files are created `0600`; the parent directory is created `0700`
+(a pre-existing parent keeps its mode, matching the other STM stores — the
+file mode alone guards the content). When the log reaches `max_bytes` it
+rotates (`log → log.1 → … → log.N`, oldest dropped; `max_backups: 0`
+truncates instead). `sample_rate` keeps the given fraction of calls and
+applies to the selection+execution pair atomically — and a selection whose
+write failed also skips its execution event — so neither sampling nor write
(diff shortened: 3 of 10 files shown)


memtomem:
**Review gate: 2 Codex rounds → SHIP.**

- **R1 — NEEDS-FIX (3 findings).** (1) *Major, accepted & fixed in `bb1b92a`*: a failed selection write still returned a `selection_id`, so a transiently failing disk could persist an orphan `execution` record, contradicting the documented "write failed → skip paired execution" contract. `_append` now distinguishes write failure (`False`) from an intentional redaction drop (`True`, preserving the left-outer contract); `log_selection` returns `None` on failure and the manager skips the pair atomically — pinned by a wire-in counter test (`write_errors == 1`, where `2` would mean an execution write was attempted). (2) *Minor, adjudicated — docs aligned instead of chmod*: every sibling store (`MetricsStore`, `memory_ops` #456/#464) applies `0o700` at `mkdir` time only and leaves a pre-existing parent's mode untouched; the real bug was this PR's doc overclaim, now corrected. Force-tightening a shared pre-existing `~/.memtomem` is a cross-store policy decision for its own PR. (3) *Nit, rebutted as measurement artifact*: the >100-column report counted UTF-8 bytes (`─` = 3 bytes); in characters — what `ruff` enforces — zero lines exceed 100 in either new file.
- **R2 — SHIP.** Reviewer independently re-ran the targeted tests (22 passed), docs-sync + lifespan tests (14 passed), `ruff check`/`format` (clean), a character-based line-length check (0 over), and verified the sibling-dir adjudication against `metrics_store.py:170` / `memory_ops.py:218`. Cleared: `_append` return asymmetry (only `log_selection` consumes it), redaction-`True` semantics ("consumed", not "persisted"), lock coverage of the new return paths, and the orphan-regression counter pin.

Full suite after fixes: 3224 passed, 3 skipped. Awaiting merge approval.

PR #473, opened 2026-06-11
feat(proxy): STM-native tool-eligibility hard filter at exposure time (#465)

memtomem:
## Summary

Implements the re-scoped, STM-native stage of #465: a deterministic hard filter at the proxy's tool-exposure choke point (`get_proxy_tools()`), built entirely from signals the proxy already owns — no toolgraph dependency. **Issue #465 stays open after this merge** as the umbrella for the optional toolgraph `eligible_tools` provider stage (gated on toolgraph#15/#13 landing).

New module `proxy/tool_eligibility.py` absorbs what used to be three scattered exposure guards (per-tool `hidden` skip inline in `get_proxy_tools`, connect-time duplicate-name skip, connect-time 64-char-overflow skip) into one pure, auditable pass, and adds profile- and signal-based rules. `_connect_server` now keeps every discovered tool and only logs the #261 prefix guidance — the filter is the single exclusion point, so every verdict reaches telemetry on the normal startup path and connect/reconnect get identical treatment.

## Rule vocabulary (one reason per withheld name)

| code | rule | profiles |
|---|---|---|
| `duplicate_name` | composed name carried by >1 discovered tool — the **entire ambiguous group** is withheld (ambiguity pre-pass, runs before every other rule; upstream calls route by raw name, so same-named occurrences are one callable entity wearing several metadata claims — a "clean" copy must not launder a flagged sibling) | all |
| `config_hidden` | per-tool `hidden: true` | all |
| `profile_excluded` | `expose_in_profiles` (per-upstream / per-tool, tool wins) excludes the active profile | all |
| `name_overflow` | composed client name > 64-char MCP limit (#261) | all |
| `sensitive_metadata` | credential pattern (CREDENTIAL_PATTERNS only, not PII) in raw description / advertised description / **full** raw schema (unbounded scan — security gate, no downstream backstop on the advertisement path) | strict rejects; review demotes |
| `unhealthy` | upstream-attributable error rate (transport/timeout/protocol/upstream_error only) ≥ 0.95 over ≥ 5 calls in a 24h `proxy_metrics.db` window | strict rejects; review demotes |

Profile ladder (`exposure.profile`, default `strict`): **strict** = signal rules hard-reject; **review** = signal-flagged tools stay advertised but carry `review_risk_penalty` in #466 ranking telemetry (observe what strict would hide); **explore** = signal rules off. Config/structural rules apply in every profile.

**Telemetry never contradicts the advertisement:** `reject_reasons` keys are disjoint from `candidate_tools` structurally — the ambiguity pre-pass withholds every multi-occurrence name outright, so each surviving name has exactly one candidate and an entry always means "this name was withheld entirely".

## Track integration

- **#467**: the advertisement's reject map rides into every selection event's `reject_reasons` (schema v1 key set unchanged — the field was reserved for exactly this; pinned by tests).
- **#466**: review-profile penalties feed `ToolRelevanceRanker` — `final_score = relevance_score * (1 - risk_penalty)`, ordering follows `final_score`, and calls shaped by a nonzero penalty stamp `v2-bm25-risk-penalty` on both pair halves (all-zero map keeps the v1 stamp, since the math degenerates to v1 exactly).
- **Invariant**: ranking can never resurrect a hard reject — it runs over the filter's eligible output; pinned end-to-end by a wire-in test (rejected tool absent from `candidate_tools` and `ranked_candidates`, present only in `reject_reasons`).

## Design decisions

- **Health is evaluated once per `start()`** from the persisted metrics store and cached: MCP clients aren't guaranteed to re-list tools, and mid-session exposure drift would make telemetry lie about the candidate set the client saw. A health-hidden tool re-qualifies at the next startup once its failures age out of the window (startup-grained half-open probing).
- **Proxy-internal failures never count against a tool's health** (`internal_error`/`lock_timeout`/`programming` excluded) — hiding a tool because our pipeline broke would punish the wrong party.
- **Sensitive-metadata scan runs over the RAW description/schema** (truncation only removes text) plus the advertised description (covers operator `description_override`), with no length cap.
- **Registration-race safety without the connect-time guard**: the FastMCP registration loop consumes the filter's output, which is unique by construction.
- **Operators can tell withheld from missing**: `get_upstream_health` reports `tools` (discovered) and `advertised_tools`; `stm_proxy_health` renders "N tools discovered, M advertised"; rejects log once per advertisement change (WARNING for structural/signal, DEBUG for the operator's own config choices).

## Behavior change

Default profile is `strict`, so two classes of tools that were advertised before can now be withheld **without any config change**:

1. **Consistently failing tools** — ≥ 95% upstream-attributable errors over ≥ 5 calls within the last 24h (deliberately conservative; requires `metrics.enabled`, which is the default). These tools were failing nearly every call anyway; hiding them is the feature.
2. **Tools with credential-looking metadata** — e.g. a description containing `api_key=`-style text or a provider-prefixed token. Suspicious by definition; advertising one pastes the match into the client context on every `tools/list`.

Additionally: same-named duplicate groups are now withheld entirely (previously first-listed copy won — see rule table for why that was unsafe); `hidden`-tool rejects (previously silent) now appear in `reject_reasons` telemetry when `selection_telemetry` is enabled; `conn.tools` / health tool counts now reflect the full discovered catalogue (exclusion happens at exposure, not discovery). Escape hatches: switch `exposure.profile` to `review`/`explore`, tune health thresholds, or scope tools with `expose_in_profiles`.

## Review

Codex 4 rounds: R1 NEEDS-FIX (Major: duplicate-name telemetry could contradict the advertised set → occurrence/name fold) → R2 NEEDS-FIX (Majors: connect-time skips bypassed the filter on the normal startup path → single choke point; 20k schema-scan cap → unbounded) → R3 NEEDS-FIX (Major: first-eligible-wins let a clean copy launder a flagged same-named sibling → ambiguity pre-pass withholds the whole group, which also deleted the R1 fold as unnecessary; Minor: discovered-vs-advertised counts) → **R4 SHIP** (no findings).

## Tests

- `tests/test_tool_eligibility.py` (new, 45 tests): every reason code, the ambiguity pre-pass (incl. the R3 laundering attack and precedence over config rules), the profile matrix, disjointness invariants, deep-schema credential regression, health-flag computation from a real metrics store (window aging, category filtering, fail-open), `get_tool_error_stats`, and the ProxyManager wire-in (telemetry + ranking + resurrection invariant + session stability + `start()` health snapshot).
- Extended: `test_selection_log.py`, `test_tool_relevance.py`, `test_proxy_manager_lifecycle.py` (#261 contract moved to exposure time), `test_server_tools.py` (health rendering).
- Full suite: **3288 passed** (`-m "not ollama"`), ruff clean, mypy delta-clean (8 pre-existing errors untouched). CI green on all jobs.

## Follow-ups (not in this PR)

- Surface reject counts / current verdict via `stm_proxy_stats` or `mms status` (recorded #467 follow-up).
- Per-tool profile *evaluation* override (false-positive escape hatch finer than the global profile switch) — wait-for-signal.
- toolgraph `eligible_tools` as an optional feature provider once toolgraph#15/#13 land — plugs into this filter as an additional rule source.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -263,6 +263,9 @@ Full example with all options:
         "search_code": {
           "compression": "selective",
           "max_result_chars": 8000
+        },
+        "delete_repository": {
+          "expose_in_profiles": ["explore"]
         }
       }
     }
@@ -309,6 +312,13 @@ Full example with all options:
   "tool_relevance": {
     "enabled": true,
     "top_n": 20
+  },
+  "exposure": {
+    "profile": "strict",
+    "health_window_hours": 24.0,
+    "health_min_calls": 5,
+    "health_error_rate_threshold": 0.95,
+    "review_risk_penalty": 0.5
   }
 }
 ```
@@ -326,6 +336,24 @@ startup, like `metrics.enabled` — toggling it requires a restart.
 `candidate_features` — telemetry input only, exposure never changes. It is
 inert unless `selection_telemetry` is enabled, hot-reloadable, and bounded
 by `top_n`. Details in [selection-telemetry.md](selection-telemetry.md).
+
+`exposure` configures the tool-exposure hard filter: at advertisement time
+the proxy rejects tools that are config-hidden, scoped out of the active
+profile via `expose_in_profiles` (per-upstream or per-tool — tool wins),
+structurally broken (composed name over the 64-char MCP limit, duplicate
+composed names), or — under the `strict` profile — flagged by a signal:
+credential-looking text in the tool's description/schema, or a
+consistently-failing recent history in `proxy_metrics.db`
+(`health_error_rate_threshold` over `health_min_calls`+ calls within
+`health_window_hours`; only upstream-attributable errors count). Under
+`review`, signal-flagged tools stay advertised but carry
+`review_risk_penalty` in tool-relevance telemetry; under `explore`, signal
+rules are off. Health is evaluated once at startup, so the advertised set
+is stable for the session — a health-hidden tool is re-evaluated at the
+next restart and returns once its failures age out of the window. Reject
+reasons land in each selection event's `reject_reasons` when
+`selection_telemetry` is enabled; see
+[selection-telemetry.md](selection-telemetry.md) for the reason vocabulary.
 
 ### Token-equivalent budgets (CJK / non-Latin workloads)
 
--- a/docs/selection-telemetry.md
+++ b/docs/selection-telemetry.md
@@ -134,16 +134,23 @@ reason code`:
 | `config_hidden` | per-tool `hidden: true` override | all |
 | `profile_excluded` | `expose_in_profiles` does not include the active profile | all |
 | `name_overflow` | composed client-side name exceeds the 64-char MCP limit | all |
-| `duplicate_name` | composed name already claimed by an earlier eligible tool | all |
 | `sensitive_metadata` | credential-pattern match in the tool's description/schema | rejects under `strict`; demotes under `review` |
 | `unhealthy` | upstream-attributable error rate over the recent metrics window crossed the threshold | rejects under `strict`; demotes under `review` |
 
 Reason codes are the only #465 payload in the log — no tool metadata, no
-error text — so the redaction policy below is untouched. A rejected tool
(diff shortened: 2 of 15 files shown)


pandas-studio (approved):
LGTM

Issue #475, opened 2026-06-11
feat(cli): round-trip transition — origin provenance, prune backup log, and `mms eject`


memtomem:
## Background — why now

#226 shipped the forward half of the host↔STM transition: `mms add --import --prune` (plus `mms init --prune-originals` and the post-hoc `mms prune`) moves directly-registered MCP servers behind the proxy and collapses the dual registration by removing the host-side entries. That path is currently **one-way and lossy**:

- Prune deletes the host original with no record anywhere (`_prune_from_source` → `claude mcp remove` shell-out / Desktop JSON rewrite).
- `_normalize_client_entry` is a lossy mapping: HTTP `headers` are not imported at all, dangerous env keys are filtered, and the client `type` hint is folded into the transport enum — the host entry cannot be reconstructed from the STM entry.
- `stm_proxy.json` records no provenance: nothing says where an imported entry came from.
- `mms remove` only edits the STM config. After `--prune`, removing a server leaves it registered **nowhere**; the user must re-create the host entry from memory, secrets included.

Concrete scenario: import a server with `--prune`, later decide to stop proxying it (debugging, latency, churn) → today that is a hand-reconstructed `claude mcp add-json`.

## Why existing machinery doesn't cover this

- `mms remove` is STM-side only by design and should stay that way.
- The `~/.mms/` registry (`mms import` / `mms host sync`) is a separate subsystem: `mms add --import` writes nothing there, and its sidecar stores a human-readable source label without machine-readable scope — it cannot drive a precise restore. This proposal does not touch `~/.mms/`.
- Host configs cannot be re-derived from STM entries because the import normalization is lossy (above).

### 1. Origin provenance (captured at import, per entry in `stm_proxy.json`)

```jsonc
"origin": {
  "schema_version": 1,
  "source": { "kind": "claude-user", "pruned": true, "pruned_at": "..." },
  "duplicates": [ { "kind": "claude-desktop", "pruned": false } ],
  "imported_at": "...",
  "original": { /* verbatim host entry */ }
}
```

- Structured source kinds (`claude-user`, `claude-project`+path, `mcp-json`+path, `claude-desktop`) kept in lockstep with the prune writers' label mapping via one shared table.
- Per-source `pruned` status — prune already permits partial failure across primary + duplicate sources.
[...]

PR #476, opened 2026-06-11
feat(cli): capture origin provenance at import, redact it from --json (#475)

memtomem:
PR1 of 4 for #475 (round-trip transition: origin provenance, prune backup log, `mms eject`). PR2 (backup log + per-source `pruned` + write lock) and PR3 (`mms eject` core) build on the data captured here; PR1/PR2 do not depend on the `claude mcp add-json` expressiveness probe gating PR3.

## What

- **Candidate raw capture** — `_discover_candidates` now keeps the **verbatim** host entry (`raw`) and a structured `{kind, path?}` source ref per candidate, for the primary source and every duplicate. Normalization is lossy (HTTP `headers` not imported, dangerous env keys filtered, client `type` folded into the transport enum), so the host entry cannot be reconstructed from the STM entry — the verbatim copy is the only faithful restore payload.
- **Shared source table** — discovery labels, `_source_removal_hint`, and `_prune_from_source` previously repeated the same label literals in three places. They now read one `_SOURCE_SPECS` table (label ↔ `kind` ↔ `claude mcp remove -s` scope), so the mapping that `origin.source.kind` persists cannot drift from what the prune writers act on.
- **Origin block on both import paths** (`mms init`, `mms add --import`):

  ```jsonc
  "origin": {
    "schema_version": 1,
    "source": { "kind": "claude-user", "pruned": false },
    "duplicates": [ { "kind": "claude-desktop", "pruned": false } ],
    "imported_at": "2026-06-11T05:00:00Z",
    "original": { /* verbatim host entry */ }
  }
  ```

  Stored inline in the entry: the CLI load/save path is raw-dict passthrough, so every existing command (`add`, `remove`, `surfacing`, …) preserves it with no migration. `pruned` stays `false` in this PR — the per-source updates land with the backup log (PR2).
- **`UpstreamServerConfig.origin`** — validated optional field documenting the schema server-side. The proxy runtime never reads it; the CLI constructs the block *through* the model (`UpstreamOrigin` / `OriginSource`), so the written shape cannot drift from the documented one. Older binaries ignore the field via pydantic's default `extra="ignore"`, and since the server never writes the config back there is no loss path.
- **Redacted `--json` serializer** — ships in the same PR as capture (separating them would leave a vulnerable in-between state): `mms status --json` / `mms list --json` previously dumped the raw `upstream_servers` dict, which would leak `origin.original`'s verbatim `env`/`headers`. The serializer keeps the provenance summary and replaces `original` with a `has_original` flag. Redaction is output-only; the file keeps the original (0600, unchanged exposure).

## Behavior change

- Imported entries in `stm_proxy.json` gain an `origin` block, including a **verbatim copy of the host entry's `env`** (the file was already the secret-bearing surface, written `0600`; no new exposure class).
- `mms status --json` / `mms list --json` output for origin-bearing entries changes shape: `origin.original` is never emitted; `origin.has_original` appears instead. Entries without `origin` (manual adds, pre-existing configs) are byte-identical to before.
- Manual entry paths (`mms add NAME`, init's manual prompt flow) intentionally get no origin block.
- Note: these `--json` paths already expose each entry's own `env` today — pre-existing surface, called out in #475 as out of scope for this PR.

## Tests

- Discovery: verbatim `raw` kept where normalization is lossy (headers / unknown fields / dangerous env keys); structured `source_ref` per kind incl. `path` for `claude-project` (resolved cwd) and `mcp-json` (file path); structured per-source `duplicates` records.
- Both import paths persist the origin block; end-to-end round-trip from real sandbox host files pins `origin.original` deep-equals the host entry (incl. an `LD_PRELOAD` key that the active entry filters); the saved block re-validates against `UpstreamOrigin` (CLI↔server lockstep pin).
- Compat: full block validates on `UpstreamServerConfig`; unknown origin keys / future source kinds ignored (`extra="ignore"` pin); entries without origin unaffected.
- Preservation: `mms surfacing` / `mms remove` of a sibling leave the block untouched (raw-dict passthrough pin).
- Redaction: `status --json` + `list --json` emit no `original` and no secret value, keep summary keys + `has_original`, and don't write back; origin-less entries pass through unchanged.
- `uv run pytest -m "not ollama"`: 3302 passed, 3 skipped. `ruff check` / `ruff format --check src` clean; `mypy src` unchanged from main (8 pre-existing errors in untouched files).

### DIFF
--- a/src/memtomem_stm/cli/proxy.py
+++ b/src/memtomem_stm/cli/proxy.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import asyncio
+import copy
 import json
 import logging
 import os
@@ -14,6 +15,7 @@ import sys
 import tomllib
 from collections.abc import Iterator
 from contextlib import AsyncExitStack, contextmanager
+from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, TextIO
 
@@ -30,6 +32,7 @@ from memtomem_stm.mms.import_hosts import (
     _is_self_reference,
     _read_json_safely,
 )
+from memtomem_stm.mms.state import utc_now_iso
 from memtomem_stm.proxy import tool_name_budget
 from memtomem_stm.utils.fileio import atomic_write_text
 from memtomem_stm.utils.redact import redact_exception_text, redact_url_userinfo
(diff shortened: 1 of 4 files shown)


pandas-studio (approved):

memtomem:
Codex review (dev-trio reviewer role) R1: **SHIP** — no blocker/major/minor findings.

Checked: origin capture paths (`_discover_candidates` / `_build_origin` / both import commands), redaction paths (`status --json`, `list --json`, probe/health output, server-side MCP status tools), `_SOURCE_SPECS` behavior vs prior removal-hint/prune dispatch incl. unknown-label fallback, back-compat for candidates without `raw`/`source_ref` and configs without `origin`, plus a direct aliasing/redaction probe of `_build_origin` / `_redacted_servers_json`. Reviewer re-ran the focused test files (291 passed) and ruff on changed files.

PR #479, opened 2026-06-11
feat(cli): remove eject hint + mms list origin summary + docs (#475)

### LINKED ISSUE #475: feat(cli): round-trip transition — origin provenance, prune backup log, and `mms eject`

memtomem:
## Summary

PR4 of 4 for #475 — the last PR of the round-trip chain (#476 origin capture → #477 backup log + locks → #478 `mms eject`). Closes out the issue's PR table: `mms remove` eject hint, `mms list` origin summary, and reference docs for the whole feature.

### `mms remove` — orphaning hint

The scenario #475 opened with: import with `--prune`, later `mms remove` → the server is registered **nowhere**. `remove` now prints a note before the confirm prompt when the recorded provenance says that would happen — the primary source was pruned and no un-pruned duplicate remains:

```
Note: 'github' was imported from Claude Code (user) and the host original was pruned —
  removing it from STM leaves it registered nowhere.
  To restore it to the host instead, run: mms eject github
```

Advisory by design: the removal is never blocked, the hint precedes the prompt so a TTY user can abort, and the pruned flags are trusted as recorded — `mms eject`'s live no-clobber check (#478) stays the authoritative guard. No origin / un-pruned origin / surviving duplicate → no hint (removing those just stops proxying).

### `mms list` — ORIGIN column

```
NAME       PREFIX  TRANSPORT  COMPRESSION  ORIGIN           COMMAND / URL
--------------------------------------------------------------------------
manual     mn      stdio      auto         -                npx -y @demo
imported   im      stdio      auto         claude-user*     npx -y @gh
dual       du      sse        auto         claude-desktop   https://x.example/mcp

3 server(s) configured.
* host original pruned — `mms eject NAME` restores it (see `mms eject -h`).
```

The cell shows `origin.source.kind` (machine identifier, matching what `--json` exposes), `*` marks a pruned host original, and the legend only renders when a starred row exists. `--json` output is unchanged — the redacted origin block shipped in #476.

### Docs

- **docs/cli.md**: new `eject` section (restore semantics, host-write-first invariant, no-clobber, verified release / `--accept-schema-loss`, argv secret gate, per-entry failure contract, backup-log suggestion); `list` ORIGIN column + `--json` redaction; `remove` hint; backup-before-delete log in `prune`; reversibility paragraph in the `add --import` docs; `eject` added to the Commands index and Examples.
- **README**: one sentence noting the import transition is reversible via `mms eject`.

### Structural docs fix (consequent, isolated in its own commit)

The `prune` section's "Dual-registered" / "STM's own config" paragraphs and its examples block were stranded under `### surfacing` (that section was inserted mid-prune by #422). Inserting `### eject` between prune and surfacing required resolving the split, so the second commit moves them back verbatim — no prose changes to the moved paragraphs.

## Why the hint logic doesn't live-probe host configs

A live `_host_existing_entry` check per remove would catch a manually re-added host entry, but the hint is advisory and the recorded flags are correct for the flow the feature targets (our own prune writers set them per source in #477). Over-firing in the manually-re-added case is harmless — `mms eject` then detects the identical host entry, skips the write, and removes the STM entry, which is exactly what the user wants.

## Tests

- `TestRemoveEjectHint` (7 cases): hint fires only when source pruned **and** all duplicates pruned; precedes the confirm prompt (decline keeps the entry); never fires for manual adds, un-pruned origins, or surviving duplicates; removal proceeds under `--yes`.
- `TestListServers`: ORIGIN column content (`kind`, `*`, `-`), legend presence/absence, and the existing header-alignment pin extended to the new column.
- `tests/test_docs_sync.py` (2 new pins): the cli.md Commands index must equal the click group's registered command set (both directions), and the eject section must mention every long flag the command actually exposes plus the backup-log filename pinned against the source literal — the "verbatim from --help" claims in this PR are test-enforced, not aspirational.

`pytest -m "not ollama and not bench_qa_meta and not bench_qa_llm_judge"`: **3359 passed, 3 skipped**. `ruff check` / `ruff format --check src` clean; mypy errors unchanged from main (8 pre-existing, none in `cli/proxy.py`).

## Behavior change

`mms list` human-readable output gains a column and (when applicable) a trailing legend line; anything parsing the human table by column offset will see the COMMAND/URL column shifted right. `--json` is unchanged and remains the supported scripting surface. `mms remove` may print up to three extra stdout lines before the prompt/removal.

Closes #475.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ mms add filesystem \
 
 `--prefix` is required: it's the namespace under which the upstream server's tools will appear (e.g. `fs__read_file`). Repeat for each MCP server you want to proxy.
 
-If you've already configured MCP servers in Claude Desktop, Claude Code, or a project `.mcp.json`, `mms add --import` (alias `--from-clients`) reuses the init wizard to bulk-select them — skipping anything already registered.
+If you've already configured MCP servers in Claude Desktop, Claude Code, or a project `.mcp.json`, `mms add --import` (alias `--from-clients`) reuses the init wizard to bulk-select them — skipping anything already registered. Moving behind the proxy is reversible: imports capture where each entry came from, and `mms eject NAME` restores it to its host client if you later want to stop proxying it (see [docs/cli.md](https://github.com/memtomem/memtomem-stm/blob/main/docs/cli.md#eject)).
 
 ```bash
 mms list      # show what you've added
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -37,6 +37,7 @@ Options:
 Commands:
   add        Add an upstream MCP server to the proxy configuration.
   daemon     Manage the local surfacing daemon (warm LTM connection for...
+  eject      Restore imported upstream(s) to their host MCP client, then...
   health     Check upstream server connectivity.
   hook       Compress and/or surface for a host's built-in tool call...
   host       Host-config inspection and sync (RFC §7.3).
@@ -180,6 +181,8 @@ Use `--from-clients` (alias `--import`) to bulk-pick additional servers from the
 
 To remove the original direct registrations after a successful import, pass `--prune`. On a TTY you get a `(name, source)` confirm prompt that defaults to **No** before any file edits; in non-TTY callers (CI, scripts) you must pass `--prune` explicitly — the flag never auto-fires on inferred consent. A candidate registered in more than one source client is pruned from every source, not just the one it was imported from. Prune failures are non-fatal: the import stays, and each failed entry prints the exact manual `claude mcp remove` or Claude Desktop edit to retry. `--prune` without `--from-clients` is a usage error rather than a silent no-op.
 
+The import→prune transition is reversible: every import records an `origin` provenance block per entry (the structured source plus the verbatim host entry), and every prune backs the deleted host entry up to `~/.memtomem/pruned_upstreams.json` before removing it. [`mms eject`](#eject) walks the whole path back — it restores the captured host entry to where it came from and removes the entry from STM.
+
 > **Note**: The CLI's `--compression` flag exposes 5 of the 10 strategies. The remaining five (`extract_fields`, `schema_pruning`, `skeleton`, `progressive`, `llm_summary`) are configured by editing `stm_proxy.json` directly. See [Compression Strategies](compression.md).
 
 ### `list`
@@ -192,7 +195,9 @@ Options:
   --json         Output as JSON for scripting.
 ```
 
-Prints the configured upstream servers in a table — name, prefix, transport, compression strategy, and the command (stdio) or URL (SSE / HTTP). Reads the config only; does not probe connectivity (use `mms health` for that). With `--json` the output becomes `{"config_path": ..., "servers": {...}}` for scripting; a missing config file returns `{"error": "config_not_found", "path": ...}` instead of a text fallthrough so callers can branch on shape.
+Prints the configured upstream servers in a table — name, prefix, transport, compression strategy, origin, and the command (stdio) or URL (SSE / HTTP). Reads the config only; does not probe connectivity (use `mms health` for that). With `--json` the output becomes `{"config_path": ..., "servers": {...}}` for scripting; a missing config file returns `{"error": "config_not_found", "path": ...}` instead of a text fallthrough so callers can branch on shape.
+
+The ORIGIN column summarizes import provenance: `-` for entries added manually (or imported before provenance capture), otherwise the recorded source kind (`claude-user`, `claude-project`, `mcp-json`, `claude-desktop`). A trailing `*` marks an entry whose host original was pruned — it now exists only behind STM, and [`mms eject`](#eject) can restore it. In `--json` output the `origin` block appears with `origin.original` redacted (`has_original` tells you whether one was captured) because the verbatim host entry may carry secrets.
 
 ### `remove`
 
@@ -205,6 +210,8 @@ Options:
 ```
 
 Removes an upstream MCP server from the proxy configuration by name. Prompts for confirmation on a TTY unless `-y` or `--yes` is passed.
+
+`remove` only edits the STM config — it never touches host configs. If the entry was imported and every host original was pruned, removing it would leave the server registered **nowhere**, so the command prints a note (before the confirmation prompt) pointing at [`mms eject`](#eject), which restores the host entry instead. The hint is advisory; the removal itself is never blocked.
 
 ### Examples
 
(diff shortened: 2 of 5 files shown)


pandas-studio (approved):

memtomem:

- **R1: NEEDS-FIX** — (Major) the `mms list` pruned marker starred any entry whose primary source was pruned, contradicting `mms remove`'s orphaning hint when an un-pruned duplicate still registered the server; (Minor) the two new docs-sync pins were weaker than their stated intent (name-set / mention-only).
- **Fix** (`4742485`): extracted `_origin_fully_pruned()` (primary **and** all duplicates pruned) shared by both surfaces + partial-prune regression test + docs state the every-source semantics. Strengthened the docs-sync pins: Commands index = bidirectional set equality + each description must be a verbatim prefix of the command's actual short help; eject usage block = bidirectional flag-set equality. Byte-equality against rendered `--help` was evaluated and rejected — click's wrap/truncation point shifts with rendering width (piped CLI vs CliRunner differ by one column), so a byte pin would be environment-flaky.
- **R2: SHIP** — R1 resolved; one minor carried forward: the shared predicate trusted truthiness, so hand-edited provenance (`"pruned": "false"`, dict `duplicates`) classified as fully pruned.
- **Hardening** (`be5b9b6`): strict `is True` booleans, `duplicates` must be a list of dicts, malformed shapes answer False (advisory surfaces must under-claim on untrusted data). Unit tests cover each malformed variant.

Full suite after all rounds: **3363 passed, 3 skipped**.

PR #480, opened 2026-06-11
docs: auto-index Stage 4 is library-mode only, by design

### LINKED ISSUE #299: docs/configuration.md auto-index Stage 4 description does not match server runtime behaviour

memtomem:
## Summary

Closes out the STM→LTM write-integration design question (#288 lineage) at its decision point: the bundled `mms` server **intentionally** wires no LTM write adapter, and the docs now say so as a design statement instead of a temporary gap.

## Why now

The question was opened 2026-05-02 with a ~6-week decision window and an explicit default (document the status quo honestly). The window closes 2026-06-15. Signal check across the whole window, all four reopen conditions:

1. **Operator reports of inert `AUTO_INDEX__*` config**: zero. The only issue ever filed (#288) was self-identified during an internal review, and its operator-visible half was already resolved by the startup `inert` warning (#339, shipped May).
2. **A use case wanting STM to warm LTM** (agent pipeline leaving indexable artifacts): none filed — issues or Discussions.
3. **Write-side payload redaction shipping**: not shipped (still deferred).
4. **`mem_add`/`mem_index` schema contract stabilizing in core**: no signal; core upstream remains alpha-pinned.

Zero incidents → per the pre-committed decision rule (and plain YAGNI), adopt the zero-code option rather than build an adapter nobody has asked for.

## What changes

- **docs/configuration.md** Stage 4 NOTE: reframed from "currently inert — tracking issue for the adapter that will unblock this: #288" (forward-looking promise; #288 is closed) to "**library-mode only, by design**" — the `AUTO_INDEX__*`/extraction settings take effect only for library callers constructing `ProxyManager(..., index_engine=...)` directly (the hook the `FileIndexer` Protocol exists for). Inert-in-bundled-server semantics and the #339 startup warning reference are unchanged.
- **tests/test_docs_sync.py**: the two existing paired pins (server `ProxyManager` source pin ↔ configuration.md NOTE pin, both from #339) keep enforcing both directions; their docstrings/messages now state the decision framing, and the NOTE pin additionally requires the "library" keyword. Stale "close #299" instructions in failure messages updated (#299 closed in May).

No code or behavior change. `FileIndexer` and the `index_engine=` kwarg stay as-is (zero churn, cheap to reverse).

**Reversal trigger** (any one): an operator actually reports inert `AUTO_INDEX` config expecting server-side writes; a concrete warm-LTM use case lands; or core's `mem_add` MCP contract stabilizes enough to negotiate (mirroring the existing `search_formats` capability negotiation).

## Tests

`tests/test_docs_sync.py` 11 passed; ruff clean. Doc-only — full suite unaffected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,12 +38,15 @@ flowchart TB
     STM <-.->|surfacing<br/>via MCP| LTM
 ```
 
-> The **INDEX** stage requires a `FileIndexer` engine. The standalone
-> `mms` server does not wire one today, so `auto_index` and `extraction`
-> config is inert in the default deployment — enabling them logs an
-> `inert` warning at startup but does not write back to LTM. See
-> [#288](https://github.com/memtomem/memtomem-stm/issues/288) for the
-> tracking issue on a future MCP-protocol-only adapter.
+> The **INDEX** stage requires a `FileIndexer` engine, and the bundled
+> `mms` server wires none **by design** — `auto_index` and `extraction`
+> config is inert in the default deployment; enabling them logs an
+> `inert` warning at startup but does not write back to LTM. The hooks
+> are library-mode only: pass `index_engine=` when constructing
+> `ProxyManager` yourself (see
+> [docs/configuration.md](https://github.com/memtomem/memtomem-stm/blob/main/docs/configuration.md)).
+> [#288](https://github.com/memtomem/memtomem-stm/issues/288) has the
+> history.
 
 ## Installation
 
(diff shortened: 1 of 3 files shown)


pandas-studio (approved):

memtomem:

- **R1: NEEDS-FIX** — (Blocker) `docs/caching.md:111` references a private docs path; (Major) both README write-path callouts still framed the missing adapter as pending work, contradicting this PR's by-design framing; (Minor) the Stage 4 NOTE keyword pin was too loose.
- **Disposition**: Blocker adjudicated out of scope — it predates this PR (#425) and is one of several pre-existing references across public files; tracked as a separate follow-up so this PR stays scoped to the decision. Major fixed in `9807680` (both callouts now state library-mode-only by design, #288 cited as history). Minor fixed (pin now also requires `fileindexer`, `bundled`, `no runtime effect`).
- **R2: SHIP** — zero findings; verified against actual wiring (`server.py` ProxyManager without `index_engine`; `manager.py` still accepts library callers passing one) and confirmed the diff introduces no new private-path references.

PR #481, opened 2026-06-11
docs: remove private docs-repo paths from user-facing surfaces

memtomem:
## Summary

Follow-up to the #480 review's Blocker finding (adjudicated out of scope there as pre-existing): user-facing files referenced the private docs repo **by path**, pointing readers at files they cannot open. A repo-wide scan found three path-form references; all are replaced with public guidance, and a regression pin prevents reintroduction.

## Changes

- **docs/caching.md** — the auto-index Note told operators that wiring instructions "are kept in the private `memtomem-docs/.../custom-integration.md`". Replaced with self-contained public guidance aligned with the #480 library-mode framing: protocol contract at `src/memtomem_stm/proxy/protocols.py`, wiring via `ProxyManager(..., index_engine=...)`, inert bundled-server behavior cross-linked to configuration.md.
- **src/memtomem_stm/mms/__init__.py** — module docstring cited a private spec path (and a bare section ref meaningless without it). The inline architectural summary stays; the pointers go.
- **CHANGELOG.md** (v0.1.19 entry) — the `stm_surfacing_stats` item ended with a private RFC path; now reads "documented in a private design note". Historical content otherwise untouched.
- **tests/test_docs_sync.py** — new pin: no `memtomem-docs/` (path form, with separator) across README, CHANGELOG, `docs/*.md`, and `src/**/*.py`.

## Scope boundary

Bare repo-name mentions explicitly marked private (CONTRIBUTING.md, notebooks/README.md, CLAUDE.md, a test docstring's `memtomem-docs#31` issue ref) are deliberate contributor-facing context and stay — the pin matches the path form only, on user-facing surfaces only.

## Tests

Full suite: **3364 passed, 3 skipped**; ruff clean. The new pin fails pre-fix (verified by construction — the three offenders were found by the same scan it encodes).


pandas-studio (approved):

memtomem:

- **R1: NEEDS-FIX** — (Major) the new private-path pin scanned `docs/*.md` non-recursively, so nested markdown under future docs subdirectories was invisible despite the claimed scope.
- **Fix** (`9465e89`): the pin now enumerates via `git ls-files` with recursive pathspecs — scope is exactly the git-tracked published surface; untracked local artifacts (`docs/reports/`) are excluded by construction instead of a hardcoded carve-out. Mutation-verified: a staged `docs/<subdir>/leak.md` fails the test.
- **R2: SHIP** — zero findings. Verified pathspec recursion in a throwaway repo; the subprocess `git` dependency judged acceptable for this repo's CI shape (actions/checkout tree on both OSes), with `check=True` correct because a failure means the invariant cannot establish its scanned surface.

PR #487, opened 2026-06-15
test(bench_qa): cross-version drift gate (advisory)

memtomem:
## What

Post-gate roadmap **item 8**: a cross-version drift gate that pins the whole canonical `report.json` to a committed baseline (`tests/bench/fixtures/drift_baseline.json`).

The per-PR `bench_qa` gate enforces *absolute* per-scenario floors (`qa_gate_min`, `ratio_violation`). A metric can sit just above its floor forever while eroding across many PRs — each one passing its own floor, the slow slide invisible. This gate closes that gap: **any** report change — even one still above the floor — must pass through a reviewed baseline diff.

## Why an exact snapshot, not per-metric tolerances

The suite is byte-deterministic — the determinism gate (#486) proves `report.json` is identical run-to-run and across `PYTHONHASHSEED`. With **zero** run-to-run variance, a tolerance can only *mask* a real regression smaller than itself — exactly the slide this gate exists to catch. So **equality is the gate**; a non-gating directional classifier only *frames* a failure as REGRESSION / IMPROVEMENT / NEUTRAL for the human reading it.

(A design panel weighed snapshot vs. directional-tolerance vs. hybrid; the directional-tolerance approach was rejected precisely because the suite has no variance to fit a tolerance to.)

## How

- **`replay.run_full_suite(base_dir)`** replays the whole roster into one report (each scenario in its own subdir so per-run stores never cross-talk), reusing the existing `run_scenario_once` drivers. `SUITE_SCENARIOS` becomes the single-source roster shared with the determinism gate (no two-file drift).
- **`drift.classify_drift` / `format_drift_md`** — a pure directional classifier (`compression_ratio` up = bad, `qa.ratio`/`recall` down = bad, `ratio_violation` up = bad, `verdict` pass→fail = bad, …). `rule_judge.score` is excluded (always-0.0 sentinel under the replay drivers). Never decides pass/fail.
- **`regen_drift_baseline.py`** — `uv run python tests/bench/bench_qa/regen_drift_baseline.py` rebuilds the suite, prints a directional drift summary vs the old baseline, and writes the file byte-stably. The only writer; never hand-edit.
- **`test_bench_qa_drift.py`** — the gate (deep-equal + classifier failure message + schema_version pre-check + `drift.md` artifact) under the advisory `bench_qa_drift` marker; plus required machinery guards (roster lockstep + coverage floor, classifier-coverage tripwire, marker pin, strip sanity, planted-regression + false-positive unit tests).
- **CI** — advisory `bench_qa_drift` job (`continue-on-error: true`, self-contained, distinct `bench-qa-drift-report` artifact). Marker registered in `pyproject.toml` and excluded from the default `test` job filter, so the advisory gate runs **only** in its own job and never blocks a required check.
- **CONTRIBUTING.md** — the regen workflow + a `## Baseline change` PR-callout convention.

## Verification

- 16 drift tests + the required `bench_qa` gate (23) pass; full CI-faithful suite green (`test_docs_sync` pin kept in sync).
- **End-to-end planted drift**: mutating the baseline reds the gate with the correct directional labels (`s03.qa.ratio` REGRESSION, `s03.compression_ratio` IMPROVEMENT); restore → green.
- Marker isolation verified: the required `bench_qa` job and the default `test` job both collect **0** drift gate tests.

## Adversarial review (multi-agent)

An independent review pass returned **SHIP_WITH_NITS** — no blocker/high, no live-regression escape. Its findings (the classifier walked only a hard-coded subset of fields, so a future top-level field — e.g. a `cache_hit` like #485 — could fail the gate while the summary said "No drift") were **fixed in this PR** rather than deferred: `classify_drift` now walks every report key (unknown → NEUTRAL "unclassified"), the coverage tripwire derives its field set from `ScenarioReport.__annotations__`, and a roster coverage floor + two regression tests pin the cases.

## Decisions (recommended defaults; flag if you'd prefer otherwise)

- **Advisory-first.** Promote to required (flip `continue-on-error` + add to branch protection) after an observation window, mirroring #482. The property is deterministic and two-run-proven, so a short window suffices — the only false-positive source is a forgotten re-baseline.
- **Snapshot reds on any move**, including improvements — keeps the enforced baseline honest (no stale-low floor).
- Baseline lives at `tests/bench/fixtures/drift_baseline.json` (next to the source-of-truth fixtures).
- The optional sticky-PR-comment job and pinning the runner Python patch for byte-stability are **deferred** to before promotion.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
         with:
           version: "latest"
       - run: uv sync --frozen
-      - run: uv run pytest -m "not ollama and not bench_qa_meta and not bench_qa_llm_judge and not bench_qa_sweep" --tb=short -q
+      - run: uv run pytest -m "not ollama and not bench_qa_meta and not bench_qa_llm_judge and not bench_qa_sweep and not bench_qa_drift" --tb=short -q
 
   notebooks:
     runs-on: ubuntu-latest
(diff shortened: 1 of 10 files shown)


pandas-studio (commented):
LGTM

PR #501, opened 2026-06-17
test(proxy): characterization net for _call_tool_inner refactor (A1 PR0)

memtomem:
## Background

The 2026-06-17 core-logic review flagged `ProxyManager._call_tool_inner()` (the ~862-line central proxy call pipeline) as the dominant remaining complexity risk (Area 1, High). The plan is a **behavior-preserving** extract-method refactor that splits each pipeline stage into a helper returning an explicit stage-result dataclass, shipped as a sequence of focused, sequentially-merged PRs:

**PR0 (this) → PR1 (`_fetch_upstream`) → PR2 (`_shape_response`) → PR3 (`_compress_and_surface`) → PR4a (`_run_index_stage`/`_run_extract_stage`) → PR4b (`_store_cache`).**

This PR is **PR0: the characterization-test net.** No production code changes.

## What this adds

A new `tests/test_call_tool_inner_characterization.py` that pins the cross-stage invariants the refactor must preserve byte-for-byte. The existing per-stage unit tests are strong, but none exercise the **orchestration seams** the refactor touches end-to-end. The new tests lock:

| Trap | Invariant pinned |
|------|------------------|
| **R1** | `compressed_chars` is `len(compressed)` on the compress branch but `len(cleaned)` on the progressive branch (the two branches disagree by design) |
| **R3** | the cache stores the **pre-surfacing** `compressed`, never `surfaced` |
| **R5** | the cache key uses the **unmutated** `cache_args` (no `_trace_id`) |
| **R6** | progressive paths surface via `_apply_surfacing_on_progressive`, non-progressive via `_apply_surfacing`; `surfacing_on_progressive_ok` / `surface_error` match the engine outcome per injection mode (append/prepend/raise) |
| **R8** | a non-text-only response records a `0/0` metric and returns the list; an empty response records **nothing** and returns the sentinel string |
| **R10** | every downstream consumer receives the **mutated** `upstream_args`; only `cache.set` receives the unmutated `cache_args` |
| — | `scorer_fallback` reflects a relevance-scorer fallback during compression |

This module must stay **green unchanged** through PR1-PR4b — that invariant is the behavior-preservation proof of the refactor.

## Review provenance

The refactor plan was reviewed by two independent reviewers before this PR; their findings (explicit args contract, `cfg_snap` vs live-config preservation, surfacing-helper-selection e2e, timing-field non-determinism, mandatory PR4a/PR4b split) are folded into the plan and this test net.

## Testing

- `uv run pytest tests/test_call_tool_inner_characterization.py` — 16 passed.
- `uv run pytest -m "not ollama"` — 3597 passed, 3 skipped.
- `uv run ruff check src && uv run ruff format --check src` — clean.

### DIFF
--- a/tests/test_call_tool_inner_characterization.py
+++ b/tests/test_call_tool_inner_characterization.py
@@ -1,0 +1,408 @@
+"""Characterization net for the ``ProxyManager._call_tool_inner`` refactor (A1).
+
+These tests pin the cross-stage invariants that the staged extract-method
+refactor (PR1 → PR4b) must preserve byte-for-byte. They intentionally target the
+ORCHESTRATION seams the refactor touches, which the per-stage unit tests do not
+exercise end-to-end:
+
+- **R1**  ``compressed_chars`` is ``len(compressed)`` on the compress branch but
+  ``len(cleaned)`` on the progressive branch (the two branches disagree by design).
+- **R3**  the cache stores the PRE-surfacing ``compressed``, never ``surfaced``.
+- **R5**  the cache key uses the unmutated ``cache_args`` (no ``_trace_id``).
+- **R6**  progressive paths surface via ``_apply_surfacing_on_progressive`` while
+  non-progressive paths use ``_apply_surfacing`` — and the recorded
+  ``surfacing_on_progressive_ok`` / ``surface_error`` match the engine outcome.
+- **R8**  a non-text-only response records a ``0/0`` metric and returns the list;
+  an empty response records NOTHING and returns the sentinel string.
+- **R10** every downstream consumer receives the mutated ``upstream_args`` (with
+  ``_trace_id``); only ``cache.set`` receives the unmutated ``cache_args``.
+- ``scorer_fallback`` reflects a relevance-scorer fallback during compression.
+
+This module must stay GREEN UNCHANGED through PR1-PR4b — that invariant is the
+behavior-preservation proof of the refactor.
+"""
+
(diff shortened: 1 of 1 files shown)


pandas-studio (approved):

memtomem:
## Codex review (diff) — addressed

Ran an independent Codex review of this diff. Verdict: **NEEDS-FIX** (no blockers) — several "characterization" locks were weak enough to pass even if the refactor broke the invariant. All five majors + the nit are now fixed in `226fbc2`:

| Finding | Fix |
|---------|-----|
| **R3** weak (`"hello world"` → cleaned == compressed == surfaced) | three distinct values: cleaned source / `COMPRESSED-SENTINEL` / surfaced marker; cache must hold exactly the sentinel |
| **R1 compress** only asserted `< cleaned_chars` | now `compressed_chars == len(returned compressed)` |
| **R1 progressive** compared two columns to each other | clean disabled; both columns pinned to the known cleaned length |
| **R10** only observed upstream + surfacing | added a sync auto-index/extract scenario asserting both stage-4 consumers receive the mutated `upstream_args` (`_trace_id`) |
| progressive surfacing success didn't assert the read-more footer survived | fake engine now **appends**; test asserts `stm_proxy_read_more` remains present |
| leaked `MetricsStore`/`ProxyCache` handles | `make_mgr` fixture closes them at teardown |

Full suite: **3598 passed, 3 skipped**; `ruff check/format` clean.

PR #504, opened 2026-06-17
refactor(proxy): extract _compress_and_surface from _call_tool_inner (A1 PR3)

memtomem:
## Background

PR3 of the behavior-preserving split of `ProxyManager._call_tool_inner()` — the **largest and highest-risk** stage. Follows **PR0 (#501)**, **PR1 (#502)**, **PR2 (#503)**, all merged.

Series: **PR0 ✅ → PR1 ✅ → PR2 ✅ → PR3 (this) → PR4a index/extract → PR4b `_store_cache`.**

## What changes

Extract the **COMPRESS/PROGRESSIVE branch + dynamic retention floor + AUTO resolution + the 3-tier ratio-guard fallback ladder (progressive → hybrid → truncate) + the inline SURFACE step** into a new `_compress_and_surface(...) -> CompressionResult`.

**Kept atomic on purpose (plan R6):** the PROGRESSIVE branch surfaces *before* the compress branch runs, and `progressive_fallback` (set only in the ratio-guard ladder) selects which surfacing helper runs. Splitting compress from surface would re-introduce the implicit flag threading this refactor removes.

`CompressionResult` (new, in `pipeline_stages.py`) carries:
- **`compressed`** (pre-surfacing — the cache payload) and **`surfaced`** (post-surfacing — return/index input) as **distinct** fields (R3),
- the branch-dependent **`compressed_chars_for_metrics`** (`len(cleaned)` progressive vs `len(compressed)` compress) (R1),
- the fully-mutated **`metrics_strategy`** label (R2),
- `ratio_violation`, `progressive_passthrough_on_error`, `surfacing_on_progressive_ok`, `surface_error`, `compress_ms`, `surface_ms`.

The orchestrator unpacks `comp.*` into the same locals the still-inline INDEX/EXTRACT/metrics/cache stages read (those move in PR4a/PR4b). The **scorer-fallback delta** is captured by the caller (it brackets compress *plus* the later stages), so it is deliberately not a `CompressionResult` field — the measurement window is unchanged.

### Consequent test updates
Two source-inspection pins (`test_observability.py`, `test_retention_ladder_invariants.py`) now inspect `_compress_and_surface` instead of `_call_tool_inner` — the call sites moved, behavior is identical, no assertion loosened.

## Behavior preservation — verified three ways

1. **PR0 characterization net passes unchanged** — exact `metrics_strategy` labels, dual `compressed_chars`, cache-stores-pre-surfacing-compressed, surfacing-helper selection, `surfacing_on_progressive_ok`/`surface_error` per mode.
2. **Full suite: 3598 passed, 3 skipped** (identical baseline).
3. **7-way adversarial behavior-preservation review** — independent checks on line-by-line block equivalence (mechanical diff vs `main`), R1, R2, R3, R6, the scorer-fallback window, and unpack-completeness + per-path field binding. **All preserved, zero regressions.** The equivalence check confirmed the moved block differs from the original only by the three sanctioned changes (header→docstring, `_pre_scorer_fb` hoisted to the caller, trailing `return CompressionResult(...)`).

`uv run ruff check src && uv run ruff format --check src` and `uv run mypy` clean.

### DIFF
--- a/src/memtomem_stm/proxy/manager.py
+++ b/src/memtomem_stm/proxy/manager.py
@@ -68,7 +68,11 @@ from memtomem_stm.proxy.memory_ops import (
     format_fact_md,
     index_content_matches_privacy,
 )
-from memtomem_stm.proxy.pipeline_stages import ShapedResponse, ShapePassthrough
+from memtomem_stm.proxy.pipeline_stages import (
+    CompressionResult,
+    ShapedResponse,
+    ShapePassthrough,
+)
 from memtomem_stm.proxy.privacy import CREDENTIAL_PATTERNS as PRIVACY_CREDENTIAL_PATTERNS
 from memtomem_stm.proxy.token_estimate import tokens_to_chars
 from memtomem_stm.proxy.tool_eligibility import (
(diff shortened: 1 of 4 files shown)


pandas-studio (approved):

memtomem:
## Codex review (diff) — SHIP, zero findings

Independent Codex review of `main..HEAD`. **Verdict: SHIP** — no blockers, majors, or minors. It compared the moved block against `git show main:…/manager.py` and confirmed the extracted `_compress_and_surface` matches the original inline COMPRESS/PROGRESSIVE + ratio-guard + SURFACE behavior. Specifically verified:

- progressive metrics use `len(cleaned)`, compress-path uses `len(compressed)` (R1);
- `metrics_strategy` mutations still flow into `CallMetrics` (R2);
- cache stores pre-surfacing `compressed`, not `surfaced` (R3);
- `_pre_scorer_fb` captured right after CLEAN, before the helper call, with no scorer-touching work inserted (scorer window unchanged);
- the 2 source-inspection test retargets are correct and not loosened.

It also ran the retargeted source-inspection tests, the progressive-passthrough cache-skip test, ruff, and `git diff --check` — all clean.

This corroborates the in-PR 7-way adversarial behavior-preservation review (all preserved, zero regressions) and the unchanged PR0 characterization net.

PR #518, opened 2026-06-23
feat(daemon): surface + stop daemons orphaned under a stale fingerprint (#517)

### LINKED ISSUE #517: feat(daemon): surface/manage orphaned daemons left under a stale fingerprint (incl. PROTOCOL_VERSION bumps)
## Background

The surfacing daemon's handshake + lock files are keyed by `config_fingerprint(config)` (`stm-daemon-<fp>.json`). Daemons under different effective configs publish to distinct files and **coexist** by design; `mms daemon stop`/`status` only ever act on the handshake at the *current* fingerprint (`daemon_cmd.py:209`, `:259`). The design assumes an orphaned daemon left after the fingerprint changes "removes its own on graceful teardown / idle timeout" (`discovery.py:22-24`).

## The gap

`DaemonConfig.idle_timeout_seconds=0` explicitly **pins** a daemon (no idle shutdown — `server.py:415`). When the fingerprint changes while such a daemon is running, the old daemon:
- never idle-times-out (pinned), and
- is at a *different* fingerprint, so it is **invisible to `mms daemon stop`/`status`** (they key to the current fingerprint).

Result: the pinned old daemon (+ its warm, possibly multi-GB LTM child) survives indefinitely and is unmanageable via the CLI; the operator must find and kill the PID manually.

Two triggers reach this:
1. **Config drift** (pre-existing): change any fingerprinted field (`surfacing.*`, `daemon.host`, `MEMTOMEM_STM_HOOK_SURFACE_TOOLS`, `record_feedback_events`) while pinned.
2. **Protocol bump** (new in #516): `PROTOCOL_VERSION` is now folded into the fingerprint, so **every STM upgrade across a protocol bump** changes it automatically — no config change needed.

## Why now / why separate

#516 (B2 step 2) makes trigger #2 automatic on upgrade, which is why this surfaced now. But the underlying gap (a pinned, stale-fingerprint daemon is unmanageable) is **pre-existing** and orthogonal to the wire change, and the fix is a daemon-lifecycle feature with its own semantics decision — so it is tracked separately rather than bolted onto the wire PR. #516 ships a **doc-honesty fix only** (corrects the "always self-cleans" claim and tells pinned-daemon operators to stop manually).

## Proposed approach (to be designed)

Enumerate `stm-daemon-*.json` under `data_dir` and, for any whose `config_fingerprint` differs from the current one but whose recorded `pid` is alive:
- **`mms daemon status`**: report them (e.g. "1 daemon under a different/older config is running: pid=…, fp=…"), so they are visible.
- **`mms daemon stop`**: decide policy — likely report + a `--all`/`--stale` opt-in to terminate them (default stays "current config only", since a *different live config* daemon may be legitimately wanted).

This also closes the pre-existing config-drift orphan, not just the protocol case. (The narrow alternative — recomputing each legacy `PROTOCOL_VERSION`'s fingerprint in `stop` — is special-cased and grows unboundedly with versions; rejected.)

## Acceptance
- `mms daemon status` reports a running same-user daemon at a foreign fingerprint.
- Stopping such a daemon has a defined, opt-in path.
- Tests cover the pinned (`idle_timeout_seconds=0`) + fingerprint-change scenario.

_Found by codex review of #516._

memtomem:
## Summary

Closes #517. Makes daemons **orphaned under a stale config fingerprint** visible and stoppable.

A daemon pinned with `daemon.idle_timeout_seconds=0` never idle-shuts-down. When the config fingerprint changes while it runs — any fingerprinted field, or a `PROTOCOL_VERSION` bump (now automatic on every upgrade since #516 folded it into the fingerprint) — the old daemon lingers under the *old* fingerprint, where `mms daemon stop`/`status` (keyed to the *current* fingerprint) never look. It and its warm LTM child then survive indefinitely, unmanageable via the CLI.

## What changed

- `discovery.iter_foreign_handshakes(data_dir, current_fp)` — enumerates the per-config `stm-daemon-<fp>.json` files whose fingerprint differs from the current one (fp from the filename, the authoritative routing key). Pure enumeration; degrades to `[]` on a missing/unreadable `data_dir`.
- `daemon_cmd._live_foreign_daemons` — filters those to pid-alive ones. Returns only non-sensitive fields (`fingerprint`/`pid`/`host`/`port`) — **never the auth `token`**.
- `mms daemon status` — reports live foreign daemons (a text warning + hint, and a `foreign` array in `--json`), orthogonal to the running/stale/stopped state.
- `mms daemon stop --all` — SIGTERMs them. Off by default.

## Behavior change

- **`mms daemon stop` default scope is unchanged** (current config only); `restart` is unaffected.
- New `--all` flag on `stop`; new `foreign` reporting on `status` (additive `--json` key, new text line). No change to the daemon, the wire, or the fingerprint — this is CLI-only.

## Design notes (preempting review)

- **Why SIGTERM-by-pid, not the graceful socket op?** A foreign daemon may speak an older `PROTOCOL_VERSION`; both `client._request` and the daemon's `_dispatch` reject frames whose `v` mismatches, so a graceful shutdown can't cross a protocol bump. SIGTERM works regardless — the daemon's signal handler sets its shutdown event independent of `idle_timeout_seconds`, so even a pinned daemon exits. This mirrors the **pre-existing** same-config SIGTERM fallback in `stop`.
- **Why pid-liveness, not a ping?** Same reason — we can't ping across a protocol gap, so `is_pid_alive` is the available signal.
- **Why `--all` opt-in, not default?** A live daemon under *another* config may be intentional; default-off avoids reaping it.
- **We don't unlink the foreign handshake ourselves** — the daemon removes its own on teardown; unlinking here could hide a survivor that ignored the signal.

## Known limitation → follow-up

Foreign liveness is the recorded pid only. A daemon that crashed hard (SIGKILL/OOM, not a graceful teardown) leaves its handshake on disk; the OS could recycle that pid to an unrelated process, which `stop --all` would then SIGTERM. This is the **same residual risk the current-config SIGTERM fallback already carries** — `--all` only raises its probability. It is documented honestly here (help text + docstrings + the corrected `discovery.py` "liveness" invariant) and a process-identity cross-check is tracked as a follow-up rather than shipped as a partial fix.

## Review

Authored, then run through a 5-lens adversarial review (correctness/races, security/leak, tests, conventions, docs); the confirmed test/doc gaps are folded into the second commit. The pid-recycling finding is the tracked follow-up above.

## Tests

`tests/daemon/test_internals.py`: `iter_foreign_handshakes` enumeration (current-fp excluded, malformed skipped, sorted, fp→pid map); status reports foreign + omits the token (stopped **and** running paths); bare stop leaves foreign untouched; `stop --all` SIGTERMs foreign; dead-pid orphan dropped (not reported/signalled); `stop --all` continues + exits 0 when a SIGTERM raises `OSError`; the acceptance scenario (a real fingerprint change strands a pinned daemon, then `stop --all` reaches it by pid); empty-foreign messaging. Full `tests/daemon` + `tests/cli` + docs-sync suites green; `ruff` clean.


memtomem:
Follow-up for the documented pid-recycling residual risk (the `Known limitation` section): #519 — a process-identity cross-check before SIGTERM, covering this `--all` path and the pre-existing same-config SIGTERM fallback.

memtomem:
**codex review round 1 → NEEDS-FIX (1 Major, addressed in 0bfda86).**

Codex caught a real Windows correctness bug distinct from #519: `is_pid_alive` returns `True` for any positive pid on Windows, and the foreign path can't ping across a protocol gap, so `_live_foreign_daemons` would report every stale foreign handshake as live there. Fix: foreign detection is now POSIX-only (returns `[]` on Windows; current-config behavior unchanged), with the dead Windows branch in `_stop_foreign_daemons` removed and docs/CHANGELOG/a unit test updated. The proper cross-platform primitive (a TCP connect-probe, which also closes the #519 recycling window) is folded into #519.

memtomem:
**codex review round 2 → SHIP.** No blocker/major/minor. Verified: `_live_foreign_daemons()` returns `[]` on Windows before any pid-liveness use; `_stop_foreign_daemons()` has no dangling Windows branch; default `stop` stays current-config-only and `--all` is opt-in; status JSON/text never print the handshake token. Ran the CI-filtered suite (3664 passed, 3 skipped, 22 deselected), ruff clean, and `git diff --check` clean.

PR #527, opened 2026-06-24
fix(daemon): gate foreign/stale orphan reaping on a connect-probe (#519)

### LINKED ISSUE #519: Harden daemon orphan reaping against pid recycling (process-identity cross-check before SIGTERM)
## Background

#518 (issue #517) added `mms daemon stop --all` + `mms daemon status` reporting for daemons orphaned under a stale config fingerprint. A foreign-fingerprint daemon may speak an older `PROTOCOL_VERSION`, so it can't be reached by the socket `ping`/shutdown path (both ends reject a version-mismatched frame). It is therefore identified **solely by its recorded pid** (`is_pid_alive`), and `stop --all` reaps it with `os.kill(pid, SIGTERM)`.

## The gap

A daemon that exits **non-gracefully** (SIGKILL/OOM/crash) never runs `server._teardown`, so `remove_handshake_if_owner` is never called and its `stm-daemon-<fp>.json` lingers on disk with its now-dead pid. The lifetime lock auto-releases (it's advisory), but the handshake JSON does not. Over successive config/`PROTOCOL_VERSION` changes these stale files accumulate under `data_dir`.

If the OS later **recycles** one of those pids to an unrelated process, that pid passes `is_pid_alive`, so:
- `mms daemon status` reports a "running" foreign daemon that is actually an unrelated process, and
- `mms daemon stop --all` sends it `SIGTERM`.

`is_pid_alive` confirms only that *a* process with that pid exists (it even returns `True` on `EPERM`, i.e. a pid owned by another user) — never that it is still our daemon. This contrasts with `remove_handshake_if_owner`, which gates on **both** pid and token.

## Why now / why separate

- Surfaced by adversarial review of #518 (two independent lenses).
- The risk is **pre-existing**: the same-config SIGTERM fallback in `mms daemon stop` (`_stop_current_config_daemon`) already SIGTERMs a stale-handshake pid after a failed graceful shutdown. `--all` only **amplifies the probability** by acting on every stale file rather than one.
- The conjunction needed (hard crash → that exact pid recycled → operator runs `stop --all`) is low-probability, so #518 ships the visibility/reaping feature with the residual risk **documented honestly** (the `--all` help text, the `_live_foreign_daemons`/`_stop_foreign_daemons` docstrings, and the corrected `discovery.py` "liveness is proven by ping, not pid" invariant) rather than blocking on a hardening design.
- The robust fix is its own design decision (see below) with cross-platform and latency tradeoffs — hence a separate issue.

## Proposed approaches (to be designed)

A process-identity cross-check before acting on a foreign pid. Candidates:

1. **Bare TCP connect probe** to the handshake's recorded `host:port`. A foreign daemon (any protocol) is still *listening*, so a connect succeeds even when the protocol handshake would be rejected; a recycled unrelated pid is almost never bound to that exact ephemeral port, so a connect is refused → skip it. Cheap, no new deps (asyncio already used), no platform code. **Caveat**: the OS could re-assign that ephemeral port to a different process, so connect-success is not full identity proof — but `is_pid_alive AND connect-succeeds` is materially stronger than pid alone. Adds a bounded connect timeout to `status`/`stop --all` per foreign entry.
2. **Process start-time cross-check** — compare the daemon's process create-time against the handshake's `created_at` (which it already records), skipping a pid whose start-time post-dates the handshake. Needs per-platform plumbing (`/proc/<pid>/stat` btime on Linux, `psutil.Process(pid).create_time()`, etc.) or a new `psutil` dependency.
3. **GC the stale files** — prune dead-pid foreign handshakes (and/or those whose `created_at` long predates the named pid's start) so the recycling surface shrinks over time.

Consider applying the chosen check to the **pre-existing same-config SIGTERM fallback** too, not just the `--all` path, since it shares the exposure.

## Acceptance

- A foreign/stale handshake whose recorded pid has been recycled to a non-daemon process is **not** SIGTERMed by `mms daemon stop --all` (and not reported as running by `status`).
- The chosen identity check is applied consistently to the foreign path and, ideally, the same-config SIGTERM fallback.
- Tests cover: a recycled-pid handshake (alive pid, not our daemon) is skipped; a genuine foreign daemon (alive, listening, protocol-mismatched) is still reaped.

_Found by adversarial review of #518._

memtomem:
Closes #519.

## Problem

`mms daemon status` / `mms daemon stop --all` identified a foreign-fingerprint
daemon by its **recorded pid alone** (`is_pid_alive`). A non-graceful exit
(SIGKILL/OOM/crash) never runs the daemon's teardown, so its
`stm-daemon-<fp>.json` lingers on disk naming a now-dead pid. If the OS later
**recycles** that pid to an unrelated process:

- `status` reported a phantom "running" foreign daemon, and
- `stop --all` sent it `SIGTERM`.

`is_pid_alive` only confirms *a* process with that pid exists (it even returns
`True` on `EPERM`) — never that it is still our daemon. The same exposure lived
in the pre-existing same-config SIGTERM fallback (`_stop_current_config_daemon`).

## Fix

Add a bare TCP **connect-probe** (`client.probe_listening`) and gate every
foreign/stale action on **`is_pid_alive` AND the endpoint still accepts a
connect**:

- A foreign daemon (any `PROTOCOL_VERSION`) is still *listening*, so a connect
  reaches it **without sending a frame** — no token, no version negotiation.
- A recycled pid's handshake names an **unbound** port, so the connect is
  refused and the entry is skipped.

The probe is cross-platform, so it also **re-enables foreign detection on
Windows** (where `is_pid_alive` is uninformative — always `True`, no signal-0).
That lets us drop the POSIX-only short-circuit in `_live_foreign_daemons` and
the `os.name != "nt"` guard in the same-config fallback; the connect now carries
the whole signal there.

Applied consistently to all three sites: `_live_foreign_daemons` (status +
`--all` enumeration) and `_stop_current_config_daemon` (same-config fallback).

## Residual / approaches considered

The remaining false-positive needs a **double coincidence** — the pid recycled
*and* its ephemeral port reassigned to another listener — so `--all` stays
opt-in. Full pid identity (approach #2, process start-time cross-check) was
rejected: it needs `psutil` or per-platform `/proc` plumbing for a far narrower
window than the connect-probe already closes.

## Tests

- `test_probe_listening` — unit: real listening socket → `True`; closed port /
  malformed endpoint → `False`.
- `test_recycled_pid_skipped_by_connect_probe` — alive pid, dead endpoint:
  not reported by `status`, not SIGTERMed by `stop --all`.
- `test_stale_handshake_recycled_pid_not_sigtermed` — same gate on the
  same-config fallback (clean handshake, no kill).
- `test_listening_foreign_detected_end_to_end` — genuine listening foreign
  daemon (real socket, no mocks) still reaped.
- `test_foreign_detection_cross_platform` — detection no longer branches on
  `os.name`; the connect decides.
- Foreign-orphan tests lose their POSIX-only `skipif`.

Full suite: 3838 passed, 3 skipped. `ruff check` + `ruff format` clean.

### DIFF
--- a/src/memtomem_stm/cli/daemon_cmd.py
+++ b/src/memtomem_stm/cli/daemon_cmd.py
@@ -329,6 +329,7 @@ def _stop_foreign_daemons(config: STMConfig) -> None:
     coincidence (pid recycled *and* its ephemeral port reassigned to another
     listener); ``--all`` stays opt-in for that narrow window.
     """
+    from memtomem_stm.daemon import client
     from memtomem_stm.daemon.discovery import is_pid_alive
 
     foreign = _live_foreign_daemons(config)
@@ -337,8 +338,13 @@ def _stop_foreign_daemons(config: STMConfig) -> None:
         return
     for d in foreign:
         pid = d["pid"]
-        if not is_pid_alive(pid):
-            continue  # raced to exit between enumeration and now
+        # Re-confirm the *same* two-factor gate at action time, not just
+        # `is_pid_alive`: the daemon may have exited since enumeration and the OS
+        # recycled its pid onto an unrelated process. A pid-only recheck is too
+        # weak for that race (and a no-op on Windows, where is_pid_alive is always
+        # True), so re-run the connect-probe — a recycled pid's endpoint is gone.
+        if not (is_pid_alive(pid) and client.probe_listening(d["host"], d["port"])):
+            continue  # raced to exit / pid recycled between enumeration and now
         try:
             os.kill(pid, signal.SIGTERM)
             click.echo(_ok(f"sent SIGTERM to daemon pid={pid} (fp={d['fingerprint']})"))
--- a/src/memtomem_stm/cli/daemon_cmd.py
+++ b/src/memtomem_stm/cli/daemon_cmd.py
@@ -96,28 +96,26 @@ def _live_foreign_daemons(config: STMConfig) -> list[dict[str, Any]]:
     never look. A daemon with a finite ``idle_timeout_seconds`` clears itself, but
     a pinned ``idle_timeout_seconds=0`` one lingers — this surfaces it.
 
-    Liveness is the recorded pid (:func:`is_pid_alive`): a foreign daemon may
-    speak an older protocol, so a socket ``ping`` would fail the version gate.
-    This is a weaker signal than the current-config path's socket ``ping`` — a
-    daemon that crashed (SIGKILL/OOM, not a graceful teardown) leaves its
-    handshake on disk, and the OS may recycle its pid to an unrelated process,
-    which would then look "alive" here. Closing that window (a process-identity
-    cross-check before acting) is tracked as a follow-up; the residual risk is
-    the same one the current-config SIGTERM fallback already carries. Each entry
-    carries only non-sensitive fields (``fingerprint``/``pid``/``host``/``port``)
-    — never the auth ``token`` — so callers can render them.
+    Liveness is **two-factor**: the recorded pid must exist
+    (:func:`is_pid_alive`) *and* a bare TCP connect to its endpoint must succeed
+    (:func:`~memtomem_stm.daemon.client.probe_listening`). A foreign daemon may
+    speak an older protocol, so a token ``ping`` would fail the version gate —
+    but it is still *accepting*, so a connect-probe reaches it without sending a
+    frame. The pid alone is too weak: a daemon that crashed (SIGKILL/OOM, not a
+    graceful teardown) leaves its handshake on disk, and the OS may recycle its
+    pid to an unrelated process that would then look "alive". Requiring the
+    endpoint to also accept rejects that recycled-pid case (its port is not
+    bound) — a strictly stronger check than pid alone (the residual is a double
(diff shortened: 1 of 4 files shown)


pandas-studio (approved):

memtomem:
## Review trail (codex)

**Round 1 — NEEDS-FIX (1 Major):** `_stop_foreign_daemons` re-checked only `is_pid_alive(pid)` before `os.kill`, leaving an enumeration→kill TOCTOU window: if the daemon exits and the OS recycles its pid before the kill, the pid check passes while the endpoint is gone — and on Windows `is_pid_alive` is always-True, so the pre-kill guard is a no-op there.

**Fix (`757629f`):** re-run the same two-factor gate at action time — `is_pid_alive(pid) and client.probe_listening(d["host"], d["port"])` — immediately before the SIGTERM. The same-config fallback (`_stop_current_config_daemon`) already probes right before its kill, so it needed no change. Added `test_stop_all_reprobes_before_kill` (enumeration probe True → action-time probe False → no kill).

**Round 2 — SHIP (0 findings):** finding resolved; no new `KeyError` (`host`/`port` always emitted by `_live_foreign_daemons`), no import cycle, regression test correct.

Full suite green; `ruff check`/`format` clean.

PR #525, opened 2026-06-24
feat(hook): `mms hook install`/`uninstall` — per-host registration UX (B3)

memtomem:
## What

Adds the per-host hook **registration UX** (Track B / B3). `mms hook --host <name>` (#524) made the runtime bridge host-explicit but left registration manual — you had to hand-edit each host's hook-config file. This wires it up:

`mms hook install --host <claude|codex|cursor|kimi>` read-modify-writes the host's own hook config so it fires `mms hook --host <name>` on PostToolUse:

| host | config | format | block |
|---|---|---|---|
| `claude` | `~/.claude/settings.json` | JSON | `hooks.PostToolUse` matcher-group |
| `cursor` | `~/.cursor/hooks.json` | JSON | `{version:1, hooks.postToolUse}` (no matcher) |
| `kimi` | `~/.kimi/config.toml` | TOML | top-level `[[hooks]]` |
| `codex` | `~/.codex/config.toml` | TOML | `[[hooks.PostToolUse]]` + `.hooks` (regex matcher) |

`mms hook uninstall --host <name>` is symmetric.

## Why this shape

- **Codex schema verified first.** B0 only pinned Codex's *payload*/*output*, not its *registration* TOML. Researched + pinned to the official `developers.openai.com/codex` sample (nested array-of-tables, `command` = single string, `matcher` = tool-name regex). Leaves `[features]` alone (hooks default-on); the install note tells the user to approve via Codex's `/hooks`.
- **Idempotent + symmetric.** STM's block is recognized by command shape (`_is_stm_hook_command`: a `hook` token + an `mms`/`memtomem-stm` executable), so a global or `uv run …` registration with any `--host` is updated in place, never duplicated; uninstall removes exactly that, prunes emptied containers, and leaves hand-written hooks.
- **Safe writes.** Dry-run is the **default**; `--apply` writes after backing up any prior file to `<path>.bak`. A config that exists but doesn't parse is **refused, never clobbered**. TOML hosts are re-serialized (`tomli-w`), so `config.toml` comments aren't preserved — the preview states this and the `.bak` is the safety net (JSON has no comments).
- **Matcher single-sourced.** Derived from each adapter's `native_tool_map` ∩ the read-like surface allowlist (now `READLIKE_SURFACE_TOOLS` in `hook_adapter`, which the live hook's allowlist also reads), so a registered matcher can never advertise a tool the hook ignores.
[...]

PR #530, opened 2026-06-24
fix(hook): hook_hosts.py install/uninstall robustness — sibling clobber, command false-positive, single-slot .bak (#529)

### LINKED ISSUE #529: fix(hook): hook_hosts.py install/uninstall robustness — sibling clobber, command false-positive, single-slot .bak (B3 #525 follow-up)
## Background

`mms hook install` / `uninstall` (#525, B3) read-modify-write a host's hook config. While reviewing #528 (the runtime `--host` fail-open follow-up), a `codex` pass surfaced three robustness/safety gaps in `src/memtomem_stm/cli/hook_hosts.py` that are **pre-existing** in the merged #525 code (none are introduced or touched by #528). I verified all three against `main` (commit `5617e6f`); each is reproducible. Filing as one issue since all three live in `hook_hosts.py` and would be fixed in one focused PR.

Severities below are my own assessment after checking reachability (codex rated #1 Blocker, #2/#3 Major).

---

### 1. `_nested_install()` clobbers sibling handlers co-located in STM's matcher group (Medium)

`hook_hosts.py:174` — on re-install, when a matcher-group entry already contains an STM handler, the **whole entry is replaced** with a block holding only STM's handler:

```python
for i, entry in enumerate(entries):
    if _entry_has_stm_handler(entry):
        entries[i] = block   # block.hooks == [STM handler only]
        return data
```

Verified: a user entry `{matcher, hooks:[STM handler, my-guard.sh]}` becomes `{matcher, hooks:[STM handler]}` after re-install — `my-guard.sh` is lost.

**Reachability is narrow**: a fresh install *appends* its own separate entry (`entries.append(block)`), so STM's entry normally holds only STM's handler. The loss requires a user to manually co-locate their handler inside STM's matcher-group entry. **But** it's a genuine **asymmetry**: `_nested_uninstall()` already carefully preserves siblings (`inner = [h for h in entry["hooks"] if not _is_stm_command_handler(h)]`); install should be symmetric.

→ Fix: preserve non-STM inner handlers on install too (filter + re-add STM), instead of replacing the entry wholesale. Add a mixed-handler regression test.

---

### 2. `_is_stm_hook_command()` false-positives on compound commands (Low)

`hook_hosts.py:97` — the check is "a bare `hook` token anywhere **and** an STM executable basename anywhere", not "an STM executable whose subcommand is `hook`":

```python
if "hook" not in tokens: return False
return any(Path(tok).name in _STM_EXECUTABLES for tok in tokens)
```

Verified: `echo hook && memtomem-stm status` → `True` (a user command that is *not* an STM hook), so install/uninstall could overwrite/remove it.

**Reachability is vanishingly low** (the existing docstring even says so): it needs a hook handler whose command is a compound shell string containing the word `hook` *and* invoking an STM executable for a non-hook subcommand. Still, tightening is cheap.

→ Fix: require an STM executable token immediately followed by a `hook` token (allowing known launchers like `uv run --directory … memtomem-stm hook`). Optionally also require `type == "command"` on the handler.

---

### 3. Single-slot `<path>.bak` overwrites the comment-preserving backup (Medium)

`hook_hosts.py:535,556` — `_backup_path` is a fixed `<name>.bak`, and `apply_change` rewrites it on every apply. For **TOML hosts** (`~/.kimi/config.toml`, `~/.codex/config.toml`) the first `install --apply` re-serializes and drops comments; the `.bak` from that first apply preserves the original. A **second** apply (uninstall, or a re-install) overwrites `.bak` with the already-comment-stripped current file — so the original comment-preserving copy is destroyed. This defeats exactly the safety net #525's docs point at ("a `.bak` backup is kept").

→ Fix: non-clobbering backup naming (`.bak`, `.bak.1`, …, or timestamped) or refuse to overwrite an existing `.bak` unless explicitly forced. JSON hosts have no comments so the impact is lower there, but the prior-state backup is still lost.

---

## Scope / why a follow-up

All three are in already-merged #525 code, unrelated to #528's runtime `--host` change, so they don't block #528. Severities are low/medium with narrow reachability — flagging for triage rather than asserting urgency. Discovered via `codex` review on #528.

memtomem:
Closes #529.

Fixes the three pre-existing robustness/safety gaps a `codex` pass surfaced on #528, all in the #525 (B3) per-host hook-registration code (`src/memtomem_stm/cli/hook_hosts.py`). None are touched by #528, so they are fixed together here.

## 1. Sibling clobber on re-install (Medium)

`_nested_install()` replaced a matcher-group entry **wholesale** when it already held an STM handler, dropping any non-STM handler a user had co-located in STM's group — an asymmetry with `_nested_uninstall()`, which already filters (`[h for h in entry["hooks"] if not _is_stm_command_handler(h)]`) rather than replaces.

Now:
- An **STM-only** entry is still replaced wholesale, so a stale matcher is refreshed (preserves the `test_plan_install_update_replaces_in_place` contract: `matcher: "Read"` → `"Read|Grep|Glob|Bash"`).
- A **shared** group keeps the user's matcher + siblings and swaps **only STM's handler in place** (the first one, dropping any duplicates), preserving handler order.

## 2. `_is_stm_hook_command()` false-positive on compound commands (Low)

The check was "a bare `hook` token anywhere **and** an STM executable basename anywhere", so `echo hook && memtomem-stm status` matched (a user command that is *not* an STM hook) and install/uninstall could overwrite/remove it.

Now requires an STM executable token **immediately followed by** a `hook` token (adjacency). Still matches both registered shapes — global `memtomem-stm hook …` and source `uv run --directory … memtomem-stm hook …` (verified: `_hook_install_command_argv` always emits `<exe> hook --host <host>`).

## 3. Single-slot `<path>.bak` overwrites the comment-preserving backup (Medium)

`_backup_path()` was a fixed `<name>.bak`, rewritten on every `apply_change`. For a **TOML host** the first `--apply` re-serializes and drops comments; the `.bak` from that first apply is the only comment-preserving copy of the original. A **second** apply (uninstall / re-install) overwrote it with the already-stripped current file — destroying exactly the safety net #525's docs point at.

Now non-clobbering: `<name>.bak`, else `<name>.bak.1`, `.bak.2`, … (first free slot). The original always stays at `.bak`.

## Tests / docs

- New `test_plan_install_preserves_cohandler_in_stm_matcher_group` (mixed-handler survives re-install, shared matcher kept, exactly one STM handler).
- New `test_apply_change_second_apply_does_not_clobber_first_backup` (TOML two-apply: original at `.bak`, second at `.bak.1`, comment preserved).
- Added compound-command + bare-exe negative cases to `test_is_stm_hook_command_ignores_foreign_commands`.
- Updated the `--apply` help strings, the TOML rewrite note, the module docstring, and `docs/cli.md` to the non-clobbering backup naming.

All 716 `tests/cli/` tests pass; `ruff check`/`format` clean; `mypy` clean.


pandas-studio (approved):

memtomem:
## Review round (codex)

A `codex` pass on the initial commit raised two **Major** findings, both fixed in `061c08b`:

1. **Multi-group duplicate.** `_nested_install` returned after the *first* STM matcher-group, so a config holding two STM groups (a hand-edit, or an old install) kept firing STM twice after re-install — and a later re-install reported `already` while the duplicate survived, violating "never duplicated". It now scans *all* `PostToolUse` entries like `_nested_uninstall`: install/refresh one STM handler at the first STM group, strip STM from later ones (keeping non-STM siblings, dropping STM-only duplicates).
2. **Backup TOCTOU.** `_backup_path` chose a slot with `exists()` and `apply_change` wrote it later, so two concurrent `--apply` runs could pick the same slot and clobber each other's backup. Replaced with `_write_backup`, which claims each slot with `O_CREAT | O_EXCL` (selection + write are one atomic step) and advances on collision.

Added a two-STM-group dedup regression test and a direct `_write_backup` slot-advance test. A second codex review on the fix commit returned **SHIP** (0 findings).

PR #535, opened 2026-06-29
feat(cache): gate response cache on tool annotations to stop replaying mutating tools

memtomem:
## Background

Part of a small caching-audit series (sibling: #534 toolgraph sqlite guard). This is the headline finding, corroborated by an independent Codex review (Verdict: NEEDS-FIX / Blocker) and a multi-agent audit.

The proxy sits transparently in front of **every** upstream tool and the response cache is **on by default** (`enabled=True`, `default_ttl_seconds=3600`). The store and lookup never consulted the tool's MCP annotations — `readOnlyHint`/`destructiveHint` were preserved on `ProxyToolInfo` and used **only for advertising** (`_fastmcp_compat.py`).

So a mutating tool (`create_*` / `send_*` / `write_*` / `delete_*`) called twice with identical args within the TTL was served the first call's cached success on the fast-path **without re-entering the upstream** — the side effect never re-executed, yet the agent received a success as if it had. Silent data-integrity exposure for any github / filesystem / messaging upstream proxied through STM.

## Change

Add `_tool_cache_eligible(server, tool)` and gate **both** the lookup (`_call_tool_guarded`) and the store (`_store_cache`). Resolution order:

1. per-tool `cache` override (`ToolOverrideConfig.cache`),
2. per-server `cache` override (`UpstreamServerConfig.cache`),
3. global `CacheConfig.tool_annotation_policy` on `readOnlyHint` / `destructiveHint`.

**`tool_annotation_policy`** (new, default `conservative`):

| policy | cached |
|---|---|
| `conservative` (default) | everything **except** self-declared writers (`readOnlyHint is False` or `destructiveHint is True`) |
| `strict` | only `readOnlyHint is True` (un-annotated → may-mutate per MCP spec) |
| `ignore` | pre-gate behavior — cache regardless of annotations |

An ineligible tool is neither served from nor stored in the cache; the lookup goes straight through with **no stampede lock** so every call re-runs and its side effect re-executes. Gating the lookup (not just the store) also refuses to serve any row cached before this gate existed.

The **privacy** and **transient-key** store guards are orthogonal and still apply on top. An unknown server (direct dispatch / tests) stays eligible, preserving prior behavior.

Also adds per-tool / per-server **`cache: bool | None`** overrides so an operator can force a volatile read tool out (`cache: false`) or re-enable a mis-annotated tool (`cache: true`) without disabling the whole cache.

## Behavior change

Default flips from "cache every tool" to "cache every tool **except self-declared writers**". The un-annotated majority and declared read-only tools are unaffected (still cached). Operators who want the old behavior can set `tool_annotation_policy: ignore`.

> Note: miss accounting is unchanged — an ineligible tool still records a cache *miss* via the existing `_call_tool_inner` path (cache configured ⇒ miss recorded). Refining per-tool cache stats is deferred to the observability follow-up.

## Tests

`tests/test_cache_eligibility.py` covers the policy matrix, override precedence, and the key regression via the `call_tool` seam:
- `readOnlyHint=False` tool called twice with identical args → upstream hit **both** times (`await_count == 2`), never cached;
- `readOnlyHint=True` tool → served from cache on the repeat (`await_count == 1`);
- un-annotated tool → still cached under the default.

Full suite: **3880 passed, 3 skipped**. `ruff check`/`format` clean. (Pre-existing unrelated `mypy` advisory in `config.py:135/144` only.)

`docs/caching.md` documents the policy, the override, and cross-links `SECURITY.md` for the privacy exclusion.

### DIFF
--- a/docs/caching.md
+++ b/docs/caching.md
@@ -38,7 +38,8 @@ The key insight: **the cache stores pre-surfacing content**. Surfacing runs on e
     "enabled": true,
     "db_path": "~/.memtomem/proxy_cache.db",
     "default_ttl_seconds": 3600,
-    "max_entries": 10000
+    "max_entries": 10000,
+    "tool_annotation_policy": "conservative"
   }
 }
 ```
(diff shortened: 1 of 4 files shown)


pandas-studio (approved):

memtomem:
### Codex review — adjudication

Ran an independent Codex review. Verdict: NEEDS-FIX with one **Major** (no blockers):

> `_tool_cache_eligible()` resolves the per-tool/per-server `cache` override from `conn.config` (the connection's startup config) rather than the hot-reloaded `cfg_snap`. The docs say `tool_overrides.*` take effect on the next tool call, so a hot-added `cache: false` might be ignored until restart.

**Consciously not actioned in this PR**, after verifying:

1. **It's a pre-existing, repo-wide property, not a regression.** The existing `_resolve_tool_config` already resolves *every* per-server/tool field — `compression`, `cleaning`, `tool_overrides.*` — from `conn.config`, never from `config.upstream_servers[server]`. `conn.config` is the connect-time snapshot and is not refreshed on mtime hot-reload (no refresh code exists; `_reconnect_server` reuses it; adding/removing upstreams already requires a restart). My code mirrors that sibling exactly.
2. **The prescribed fix would introduce inconsistency.** Resolving only the cache override from `cfg_snap.upstream_servers` while compression/cleaning/other overrides keep resolving from `conn.config` would make `cache: false` hot-reload but `compression: …` not — a surprising one-field exception to the established pattern.
3. **The safety-critical path is unaffected.** The writer gate is driven by the tool's MCP annotations plus `cache.tool_annotation_policy`, and the **policy is read from the hot-reloaded `cfg_snap`** — so it already tracks config edits. The `cache: true|false` override is only a manual escape-hatch and follows the same per-server reload semantics as every other override.

Added a code comment in `_tool_cache_eligible` documenting this so it isn't re-flagged. The broader "per-server overrides aren't mtime-hot-reloaded (docs overstate it)" point is a real but pre-existing limitation best addressed repo-wide in `_resolve_tool_config`, out of scope here.

memtomem:
### Codex re-review (rebased onto current main incl. #534/#537) — SHIP

> **Verdict: SHIP** — no path found where a policy-ineligible writer is served from cache; only minor cache-miss accounting skew.

Codex confirmed: conservative logic correct (readOnlyHint=True eligible; readOnlyHint=False or destructiveHint=True ineligible; unannotated eligible; contradictory ineligible); both lookup and store call `_tool_cache_eligible` (legacy rows not served once ineligible); eligible tools keep trace-free keys, stampede lock/double-check, cache-hit surfacing, transient-key exclusion, and privacy guards; override precedence tool>server>policy; tests pin the `await_count==2` regression. No blockers, no majors.

One **Minor** (diagnostics only): ineligible forced-forward calls still record a `record_cache_miss()` in `_call_tool_inner` though no lookup was attempted, skewing the hit-rate metric. This is the deferral already noted in the PR body. It will be fixed in #536 (the PR that surfaces hit-rate %), where the metric and the fix belong together.

PR #536, opened 2026-06-29
feat(cache): surface response-cache hit-rate/occupancy + stop wasting I/O on ttl<=0

memtomem:
## Background

Two low-severity response-cache polish items from the caching audit (series with #534, #535).

## 1. Observability

`stm_proxy_stats` printed only raw cache hit/miss counts, while every sibling figure (`Savings`, `error_rate`) is a derived %. `ProxyCache.stats()` (total/expired entries) was never surfaced in `src/`, and `max_entries` eviction was completely silent — an operator could not tell the cache was thrashing.

- Add a derived **`Cache hit rate: NN.N%`** line next to the raw counts.
- When the response cache is wired, append **`Cache entries: N (expired E, evicted V)`** from `ProxyCache.stats()`. Read inside the operator-invoked tool, so the synchronous sqlite `COUNT` stays off the hot path (consistent with the accepted sync-sqlite-on-loop posture).
- Track lifetime evictions in `ProxyCache._trim` (`cur.rowcount`) and expose via `stats()` so overflow churn is visible instead of silent.

## 2. `ttl<=0` footgun

`CacheConfig.default_ttl_seconds` accepts `0` (`Field(ge=0.0)`), but a `0` TTL makes every row **born-expired** (`is_expired` is `now >= created_at + 0`), so `get()` always missed while `set()` still paid the write+trim I/O — a "cache enabled, 0% hits" state.

`ProxyCache.set()` now short-circuits on `ttl <= 0` (do not store): use `cache.enabled=false` to disable, or a positive TTL to cache. `None` stays the distinct "never expires" sentinel and still stores.

> Chosen over a `gt=0` schema bound, which would reject any pre-existing config that persisted `0` (backward-incompatible).

## Tests

`ttl=0`/negative store-skip, eviction counter tracks trim, `stats()` includes `evictions`, and the `stm_proxy_stats` hit-rate + occupancy lines (present when the cache is wired, hit-rate still shown when it is not).

Full suite: **3867 passed, 3 skipped**. `ruff` clean.

> Docs: the `docs/caching.md` cache section is being edited in #535 (mutating-tool gate); a short note on `ttl=0` semantics + the new stats lines can fold in after both merge to avoid a cross-PR hunk conflict.

### DIFF
--- a/src/memtomem_stm/proxy/cache.py
+++ b/src/memtomem_stm/proxy/cache.py
@@ -103,6 +103,10 @@ class ProxyCache:
         self._max_entries = max_entries
         self._db: sqlite3.Connection | None = None
         self._lock = threading.Lock()
+        # Process-lifetime count of rows dropped by ``_trim`` (max_entries
+        # overflow). Surfaced via ``stats()`` so an operator can see the cache
+        # thrashing instead of it evicting silently. Resets on restart.
+        self._evictions = 0
 
     def initialize(self) -> None:
         self._db_path.parent.mkdir(parents=True, exist_ok=True, mode=0o700)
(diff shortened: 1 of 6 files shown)


pandas-studio (approved):

memtomem:
### Codex review — applied

Independent Codex review found one **Major** (no blockers): the `ttl_seconds <= 0` short-circuit returned before the `INSERT ... ON CONFLICT DO UPDATE`, so a previously-cached **live** row for the same key kept serving stale content (the pre-change `set(ttl=0)` overwrote it with a born-expired row, effectively invalidating it).

**Fixed**: the `ttl <= 0` branch now deletes any existing row for the computed key under the lock before returning, restoring the old invalidation semantics. Added `test_zero_ttl_invalidates_existing_live_row` (cache `"old"` at 60s → set `"new"` at 0 → `get()` is `None`).

Other points Codex checked and cleared: `cur.rowcount` reliability in `_trim`, the `evictions` stats key not breaking existing key-based readers, divide-by-zero-safe hit-rate, null-safe `_cache` lookup, and the operator-invoked (off-hot-path) `stats()` read respecting the sync-sqlite posture.

memtomem:
### Codex re-review (round 2) — SHIP

> **Verdict: SHIP** — prior Major is resolved; `ttl<=0` now bypasses lookup and forces upstream / store-side invalidation without counting a cache miss.

Round 1 found a Major: with `default_ttl_seconds` lowered to `0`, the lookup path still served a row cached earlier under a positive TTL before the `set()`-side invalidation could run. **Fixed**: `_call_tool_guarded()` now bypasses the cache entirely (no lookup, no store, no stampede lock, no miss recorded) when the configured TTL is non-positive — so `ttl<=0` behaves like `cache.enabled=false`; the `set()` short-circuit still deletes the stale row on the forced re-store. New `test_ttl_lowered_to_zero_stops_serving_and_invalidates` pins it via the public `call_tool` path (`await_count==2` + row removed).

Codex round 2 confirmed: `None` TTL not bypassed while numeric `0` is; miss accounting correct (eligible miss once; ineligible/no-cache/ttl-disabled record none); no regression to the merged #534/#535/#537. No blockers/majors/minors.

PR #537, opened 2026-06-29
fix(surfacing): make SurfacingCache hash FIPS-safe (usedforsecurity=False)

memtomem:
## Background

Final item from the caching audit (series: #534, #535, #536).

`SurfacingCache._hash` called `hashlib.md5(query.encode()).hexdigest()` without `usedforsecurity=False`. On a **FIPS-enforced OpenSSL build**, bare `md5()` raises `ValueError` at construction.

## Why it matters

Both `get()` and `set()` compute this hash on **every** `surface()` call, so the raise propagates into the surfacing engine's broad `except Exception`, which records outcome `error_other` and returns the response unchanged. The proxy does not crash, but **proactive memory surfacing is silently disabled for 100% of calls on FIPS systems**, signalled only by an `error_other` counter + a per-call warning log.

This was also the **lone `md5` site** in `src/` (every other digest is `sha256`), so it was inconsistent as well as fragile.

## Change

Pass `usedforsecurity=False` (py312). The digest value is unchanged on non-FIPS builds, so this is **behavior-preserving** there — the existing determinism tests (same-input==same / different!=different) still hold.

Add a **source-inspection** regression test pinning the flag (`inspect.getsource` → assert `usedforsecurity=False` present), since CI cannot run under a FIPS build.

`uv run pytest tests/test_surfacing_cache.py tests/test_surfacing_cache_hit_dedup.py tests/test_surfacing_engine.py` → 128 passed. `ruff` clean.

### DIFF
--- a/src/memtomem_stm/surfacing/cache.py
+++ b/src/memtomem_stm/surfacing/cache.py
@@ -49,4 +49,11 @@ class SurfacingCache:
 
     @staticmethod
     def _hash(query: str) -> str:
-        return hashlib.md5(query.encode()).hexdigest()
+        # ``usedforsecurity=False`` is required: this is a non-cryptographic
+        # cache key, and on a FIPS-enforced OpenSSL build bare ``md5()`` raises
+        # ValueError at construction. Without the flag, every ``get``/``set``
+        # would raise, the surfacing engine's broad ``except`` would swallow it
+        # as ``error_other``, and proactive surfacing would be silently disabled
+        # for 100% of calls on FIPS systems. The digest value is unchanged on
+        # non-FIPS builds, so this is behavior-preserving there.
+        return hashlib.md5(query.encode(), usedforsecurity=False).hexdigest()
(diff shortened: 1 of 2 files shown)


memtomem:
### Codex review — applied

Independent Codex review found one **Major** (no blockers): the source-inspection test was a weak pin — `_hash` now contains `usedforsecurity=False` in a comment *as well as* the call, so a future edit could drop the actual kwarg while leaving the comment and the test would still pass.

**Fixed**: replaced it with two stronger, behavior-oriented tests:
- `test_hash_passes_usedforsecurity_false` — spies on `hashlib.md5` (`wraps`) and asserts it is **called** with `usedforsecurity=False` (robust to comment changes, directly pins the FIPS-safe call shape).
- `test_hash_matches_canonical_md5_vector` — pins `md5("hello") == 5d41402abc4b2a76b9719d911017c592` to prove the digest value is unchanged on non-FIPS builds (behavior preservation / cache-key stability).

Codex also confirmed: no other bare `md5()` in `src` (all else sha256), and `hashlib.md5(..., usedforsecurity=False)` is valid on the 3.12 target.

PR #548, opened 2026-06-30
fix(cache): invalidate a stale row on a non-text response under a disabled (ttl<=0) cache

### LINKED ISSUE #541: Cache: ttl<=0 does not invalidate a previously-cached non-text response (stale row can resurface if TTL is raised)
## Background

`ttl <= 0` (a global `cache.default_ttl_seconds <= 0` from #536, or a per-tool/server `cache_ttl_seconds: 0` from #540) means "caching disabled". The **serve** side is correct for every response shape: `_call_tool_guarded` bypasses the lookup while the resolved TTL is `<= 0`, so a stale row is never served.

The **on-disk invalidation** of a previously-cached row, however, rides the store-side short-circuit `ProxyCache.set(ttl_seconds <= 0)` (delete + no-store) inside `_store_cache`. That delete is skipped for responses carrying non-text content:

- `_store_cache` is gated on `if self._cache is not None and not non_text_content:` (`proxy/manager.py`), so a mixed text+non-text response never reaches `set(...)`.
- A non-text-only response returns earlier in `_call_tool_inner` (`return non_text_content`), before the store path.

## Concrete gap

1. Tool `T` returns a non-text response; row cached under a positive TTL (`created_at`, frozen `ttl`).
2. Operator sets the resolved TTL for `T` to `0`. Lookup is now bypassed — the row is correctly **not served**.
3. A subsequent identical call to `T` (still non-text) does **not** invalidate the row (store path skipped).
4. Operator raises the TTL back to a positive value **before** the row's original frozen TTL elapses → the now-eligible lookup serves the **stale** row.

Text responses self-heal (step 3 deletes via `set(ttl<=0)`), so this is specific to non-text / mixed responses. It is most likely to bite the headline use case for `cache_ttl_seconds: 0` — disabling caching for a binary/volatile tool.

## Why not just delete on the lookup bypass

#536 deliberately made the `ttl <= 0` bypass do **zero I/O** (no lookup, no store, no stampede lock — see `554d36d`). An unconditional delete-by-key on every bypassed call would reintroduce per-call write I/O for a disabled-cache tool, regressing that. A correct fix needs to invalidate without that regression — e.g. invalidate only on the transition to `ttl <= 0`, a tombstone, or moving the `ttl<=0` delete out of the non-text store guard *and* covering the non-text-only early-return path (bounded to calls that actually occur, not every bypass).

## Scope

Affects both the global #536 path and the per-tool #540 path (shared lookup bypass + shared store-side invalidation). Surfaced by the codex review of #540 and adjudicated there as pre-existing rather than bundled into the per-tool feature.

## Acceptance

- A row cached under a positive TTL for a **non-text** response is invalidated (or provably never served) once the resolved TTL is `<= 0`, including after the TTL is later raised within the original window.
- No new per-call I/O on the steady-state `ttl <= 0` bypass path (preserve #536's posture).
- Tests covering: non-text response + global ttl→0; non-text response + per-tool `cache_ttl_seconds: 0`; lower-then-raise within the original TTL window.

memtomem:
## Summary

Closes #541. Fixes a correctness gap in the `ttl<=0` (caching-disabled) path introduced alongside #536 / #540: a row left behind by an **earlier text** response is never invalidated when a later response for the same key **bypasses the text store path**, so the stale text can resurface if the operator raises the TTL back within the row's frozen window.

## Root cause

The store-side `ttl<=0` self-heal (`ProxyCache.set` → delete-by-key) is only reached on the **text** store path. Four return paths in `_call_tool_inner` exit before it:

- a **non-text-only** response early-returns (`return non_text_content`);
- a **mixed** text+non-text response is skipped by the `not non_text_content` gate in `_store_cache`;
- a **text-bearing error** (text-only or mixed) raises at the `if result.isError` block;
- a truly **empty** response returns `"[empty response]"`.

The serve side is already correct for every shape — `_call_tool_guarded` bypasses the lookup while the resolved TTL is `<=0`, so the stale row is never *served* during the disabled window. The gap is purely **on-disk invalidation**: once the TTL is raised back (per-row TTL is frozen at write time), the now-eligible lookup serves the stale text. The on-disk resurfaceable row is always a **text** row (a non-text response is never stored), so the headline trigger is a tool whose response shape flips `text → non-text` on stable `(server, tool, args)` across a TTL down-then-up window — exactly the `cache_ttl_seconds: 0` "disable caching for a volatile/binary tool" case.

## Fix (narrowed approach (3) from the issue)

- Extract the delete-by-key out of `ProxyCache.set(ttl<=0)` into a reusable `ProxyCache.invalidate(server, tool, args)` (same `_make_key`; `set` now calls it).
- Add `ProxyManager._invalidate_disabled_cache(...)`: resolve the TTL (a config read, no I/O) and issue **one** best-effort `invalidate` only when `ttl <= 0`.
- Call it from each of the four non-self-healing return paths.

This preserves #536's posture: the steady-state `ttl<=0` text / no-row path stays zero-I/O; the only added I/O is one `DELETE` per such call under a disabled cache — bounded to calls that actually occur. The helper no-ops under a positive TTL, so no still-valid row is ever spuriously deleted.

## Codex review

A focused codex pass confirmed key-match, lock re-entrancy, no-spurious-delete, and the text-path restructure, and flagged the error-path gap (then text-only-error + empty for full closure) — fixed in 658d607. See the [review-summary comment](https://github.com/memtomem/memtomem-stm/pull/548) for details.

## Tests

`TestTtlZeroNonTextInvalidation` + `TestTtlZeroErrorAndEmptyInvalidation` in `tests/test_cache_eligibility.py`: helper unit (no-op positive TTL / deletes on 0), non-text-only (Site A), mixed (Site B), per-tool `cache_ttl_seconds: 0` (#540), full lower-to-0-then-raise sequence, mixed-error, text-only-error, and empty-response invalidation.

`uv run pytest -m "not ollama"` → 3952 passed, 3 skipped. `ruff` / `mypy` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ Format: [Keep a Changelog](https://keepachangelog.com/en/1.1.0/)
 
 ## [Unreleased]
 
+### Fixed
+
+- **`ttl<=0` now invalidates a stale cached row when a non-text / mixed response is served** (#548, issue #541) — the store-side `ttl<=0` self-heal in `ProxyCache.set` only ran on the TEXT store path, so a row left behind by an earlier text response for a `(server, tool, args)` key was never invalidated once caching was disabled and the same key returned a non-text response (a non-text-only response early-returns before the Stage-5 store; a mixed text+non-text response is skipped by `_store_cache`'s text-only gate). While `ttl<=0` the lookup is bypassed so the stale row is never *served* during the disabled window, but raising the TTL back within the row's frozen window (per-row TTL is frozen at write time) made the now-eligible lookup serve the stale text — most likely on the `cache_ttl_seconds: 0` headline case (disabling caching for a volatile/binary tool whose response shape flips text→non-text). The delete-by-key is now factored into `ProxyCache.invalidate` and issued from both non-text sites (the non-text-only early return and the mixed branch of `_store_cache`) only when the resolved TTL is `<=0`. This preserves #536's zero-I/O posture for the steady-state text / no-row disabled-cache path; the only added I/O is one `DELETE` per non-text call under a disabled cache, bounded to calls that actually occur.
+
 ## [0.1.30] — 2026-06-30
 
 A security-hardening release. Three coordinated fixes close an ungated
--- a/src/memtomem_stm/proxy/cache.py
+++ b/src/memtomem_stm/proxy/cache.py
@@ -215,6 +215,21 @@ class ProxyCache:
             return None
         return entry.result
 
+    def invalidate(self, server: str, tool: str, args: dict[str, Any]) -> None:
+        """Delete any cached row for ``(server, tool, args)`` — a single
+        best-effort DELETE, keyed exactly as ``set``/``get`` (``_make_key``).
+
+        Two callers: the ``set(ttl<=0)`` do-not-store short-circuit below, and
+        the manager's non-text / mixed disabled-cache path, which never reaches
+        ``set`` and so otherwise leaves a stale row from an earlier text response
+        for the same key (#541)."""
+        if self._db is None:
+            return
+        key = _make_key(server, tool, args)
+        with self._lock:
+            self._db.execute("DELETE FROM proxy_cache WHERE cache_key = ?", (key,))
+            self._db.commit()
+
     def set(
         self,
         server: str,
@@ -238,10 +253,7 @@ class ProxyCache:
             # to 0 via hot-reload), and leaving that live row would keep serving
             # stale content. This mirrors the pre-short-circuit behavior, which
             # overwrote the key with a born-expired row.
-            key = _make_key(server, tool, args)
-            with self._lock:
-                self._db.execute("DELETE FROM proxy_cache WHERE cache_key = ?", (key,))
-                self._db.commit()
+            self.invalidate(server, tool, args)
             return
         if contains_sensitive_content(result):
             # SECURITY.md: responses that look like secrets are never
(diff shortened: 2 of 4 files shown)


pandas-studio (approved):

memtomem:
## Codex review follow-up (resolved)

Ran a focused codex pass on the fix. Verdict: **NEEDS-FIX** with one Major finding; the other four correctness checks passed:

- **KEY MATCH — OK**: the lookup strips `_context_query` into `upstream_args` and reads with that key; `_call_tool_inner` snapshots the same shape into `cache_args` before `_trace_id` injection. The `invalidate` key matches the stored row.
- **LOCK — OK**: `ProxyCache.set(ttl<=0)` calls `invalidate()` without already holding `_lock`; no re-entrant deadlock.
- **SPURIOUS DELETE — OK**: `_invalidate_disabled_cache` only deletes when `ttl is not None and ttl <= 0`; a positive TTL never deletes a valid row.
- **TEXT-PATH RESTRUCTURE — OK**: the eligibility → progressive → transient-key → `set()` skip chain is preserved in order.

**Major (fixed in 658d607):** two other return paths in `_call_tool_inner` also exit *before* the Stage-5 store, so they shared the same gap:

1. a **text-bearing error** (text-only or mixed) raises at the `if result.isError` block;
2. a truly **empty** upstream response returns `"[empty response]"`.

Neither invalidated a prior cached text row under a disabled cache. A non-text-*only* error has no text and is already handled by the passthrough branch.

Both now call `_invalidate_disabled_cache` before they exit, unconditionally on the path (the helper no-ops under a positive TTL). This closes the full surface of non-self-healing return paths under `ttl<=0`: **non-text-only, mixed, text-bearing error, empty** — plus the text-success self-heal that already existed via `ProxyCache.set`. Three tests added (`TestTtlZeroErrorAndEmptyInvalidation`).

Full suite: 3952 passed, 3 skipped. ruff + mypy clean.

PR #550, opened 2026-06-30
fix(cache): complete ttl<=0 invalidation for the error and empty-response paths (#548 follow-up)

memtomem:
## Summary

Completes the #541 fix. PR #548 shipped `ttl<=0` disabled-cache invalidation for the **non-text-only** and **mixed** response paths, but it merged before this completion landed. This PR closes the remaining non-self-healing paths so a stale prior **text** row is dropped under a disabled cache on **every** path, not just the ones that happen to reach `ProxyCache.set`'s `ttl<=0` self-heal.

## Paths closed

In `_call_tool_inner`, two return paths exit *before* the Stage-5 store:

1. a **text-bearing error** (text-only or mixed) raises at the `if result.isError` block;
2. a truly **empty** response returns `"[empty response]"`.

And in `_store_cache`, three **text store-skip** branches bypass `set()` entirely (so they never reach its `ttl<=0` self-heal):

3. **cache-ineligible** tool, 4. **progressive-passthrough** degradation, 5. **transient retrieval key**.

All five left a prior positive-TTL text row live under `ttl<=0`, to resurface once the operator raises the TTL back within the row's frozen window — the same defect class as #541/#548.

## Fix

- The two early-return paths (1,2) call `_invalidate_disabled_cache(...)` (added in #548) before they exit.
- `_store_cache` resolves the TTL **once** and, when it is `<=0`, invalidates any prior row and returns **above** the whole store-skip chain (3,4,5) — under a disabled cache a text response never stores anyway. The resolved TTL is reused in the `set()` call, de-duplicating the per-store resolve; the positive-TTL path is unchanged.

The helper no-ops under a positive TTL (no spurious delete), and the steady-state `ttl<=0` text / no-row path stays zero-I/O (#536). A non-text-*only* error has no text and is already handled by #548's passthrough branch.

## Provenance

Surfaced by two codex reviews — one of #548 (after it had merged → paths 1,2) and one of this PR (→ paths 3,4,5). Both NEEDS-FIX findings verified against code and addressed here. See the PR comments for the review summaries.

## Tests

`TestTtlZeroErrorAndEmptyInvalidation` (mixed-error, text-only-error, empty) + `TestTtlZeroTextStoreSkipInvalidation` (cache-ineligible text under `ttl<=0`). `uv run pytest -m "not ollama"` → 3953 passed, 3 skipped; `ruff` / `mypy` clean.


pandas-studio (approved):

memtomem:
## Codex review (resolved)

Ran a focused codex pass on this PR's diff. Verdict: **NEEDS-FIX**, one Major; the two new sites here (error-raise, empty-response) and the four other checks all passed:

- the `isError` invalidation is placed correctly relative to `record_error` / `_mark_recorded` / `raise` and doesn't change the raised error or metrics;
- the empty-response invalidation is correct and `'[empty response]'` is reachable with a prior cached row;
- `cache_args` matches the lookup key at both sites;
- non-text-*only* errors are handled by the passthrough branch (no double-invalidation, no missed path);
- positive TTL never deletes via the helper.

**Major (fixed in 2ee9a65):** the same `ttl<=0` invariant was still incomplete one level deeper — the **text store-skip branches** in `_store_cache` (cache-ineligible, `progressive_passthrough_on_error`, transient-key) bypass `ProxyCache.set` and so never reach its `ttl<=0` self-heal, leaving a stale prior row live.

Fix (codex's first suggested option): hoist the `ttl<=0` check **above the whole skip chain** — resolve the TTL once and, when `<=0`, invalidate and return before any skip reason is evaluated (under a disabled cache a text response never stores anyway). The resolved TTL is reused in `set()`, so the positive path is unchanged and the per-store resolve is de-duplicated. Added `TestTtlZeroTextStoreSkipInvalidation` (cache-ineligible text under `ttl<=0` invalidates a prior row).

The `ttl<=0` invalidation now covers **every** non-self-healing path: non-text-only, mixed, text-bearing error, empty, and text-store-skip (ineligible / progressive / transient). Full suite: 3953 passed, 3 skipped. ruff + mypy clean.

PR #553, opened 2026-07-02
security(privacy): catch AWS secret material by label (SECRET_ACCESS_KEY / SESSION_TOKEN)

memtomem:
## Summary

The credential scan catches the AWS key **IDs** (`AKIA`/`ASIA` prefixes) but not the secret **material** those IDs unlock. Verified empirically on main:

```
False <- AWS_SECRET_ACCESS_KEY=wJalrXUtnFEMI/K7MDENG/b...
False <- AWS_SESSION_TOKEN=FwoGZXIvYXdzEBEa...
False <- aws_secret_access_key = wJalrXUtnFEMI/K7MDENG...
True  <- AKIAIOSFODNN7EXAMPLE
```

The generic label rule misses both spellings the AWS toolchain emits: `secret[_-]?key` needs its two words adjacent (`secret_access_key` splits them), and `access[_-]?token` needs the literal `access` (`session_token` has neither). A tool response echoing an `env | grep AWS` block or an STS AssumeRole JSON therefore scanned clean and could be routed to an external-LLM compressor or stored in the response cache.

## Change

One rule with two alternatives:

- **quoted form** — `"SessionToken": "…"` / `"SecretAccessKey": "…"` (STS JSON, dict repr). The quote must sit **directly on both sides of the label** and the value must **open as a string**. This is load-bearing: scanning consumers walk JSON-serialized structures whose keys are ordinary identifiers — a tool schema declaring a `session_token` property (`tool_eligibility` would hide the tool from `tools/list`) or a telemetry dict keyed by a tool name like `aws.get_session_token` (`selection_log` would drop the whole record). The pre-existing label rules never fire on bare JSON keys (the closing quote blocks their `[:=]`); both non-regressions are pinned as negative samples. (An earlier draft with a loose optional-quote branch had exactly this false-positive; caught in the pre-PR adversarial review pass and redesigned.)
- **unquoted form** — `AWS_SECRET_ACCESS_KEY=`, `aws_session_token = ` (env output, `~/.aws/credentials` INI): same shape and FP profile as the existing generic label rule.

Positive pins cover all four spellings (env-var, INI, both STS JSON keys); negative pins cover label-runs-into-identifier (`SessionTokenError:`, `SecretAccessKeyRotation:`), prose, the JSON-Schema property, and the prefixed telemetry key.

## LTM sync note

This rule is **STM-origin**: `CREDENTIAL_PATTERNS` moves to 17 while LTM's mirrored set stays at 16 until the forward-sync lands (the inverse of the #1488 → #1491 reverse-sync that produced #549). The count-pin comment in `test_privacy.py` documents the divergence window.

## Follow-ups (post-merge)

- **Changelog sync PR** (per the #551 pattern): add a Security entry for this rule, and qualify the existing Unreleased #549 entry — its "byte-identical, 16 patterns on both sides" claim becomes stale once this merges.
- **LTM forward-sync**: mirror the rule into `memtomem/privacy.py` (+ JS-translation SHA regeneration), then move both count pins to 17 together.

## Behavior change

Responses carrying these labels now route away from external-LLM compression strategies and are excluded from the response cache — the same (fail-safe) treatment every other credential label already gets.

### DIFF
--- a/src/memtomem_stm/proxy/privacy.py
+++ b/src/memtomem_stm/proxy/privacy.py
@@ -34,6 +34,32 @@ logger = logging.getLogger(__name__)
 CREDENTIAL_PATTERNS = [
     r"(?i)(api[_-]?key|secret[_-]?key|access[_-]?token)\s*[:=]",
     r"(?i)(password|passwd|pwd)\s*[:=]",
+    # AWS secret material by label. The generic label rule above misses both
+    # spellings the AWS toolchain actually emits: ``secret[_-]?key`` needs its
+    # two words adjacent (``secret_access_key`` splits them) and
+    # ``access[_-]?token`` needs the literal ``access`` (``session_token`` has
+    # neither). Two alternatives in one rule:
+    #
+    # - quoted form — ``"SessionToken": "…"`` / ``"SecretAccessKey": "…"``
+    #   (STS JSON, python-dict repr). The quote must sit DIRECTLY on both
+    #   sides of the label and the value must open as a string: a JSON-Schema
+    #   property (``"session_token": {"type"…`` — object value) and a
+    #   prefixed key (``"aws.get_session_token": "unhealthy"`` — telemetry
+    #   dicts keyed by tool name) must NOT fire, or the eligibility scanner
+    #   hides benign tools and the selection log drops whole records. The
+    #   pre-existing label rules never fire on bare JSON keys (the closing
+    #   quote blocks their ``[:=]``); this form must not regress that.
+    # - unquoted form — ``AWS_SECRET_ACCESS_KEY=`` / ``aws_session_token =``
+    #   (env output, ~/.aws/credentials INI): same shape and FP profile as
+    #   the generic label rule above.
+    #
+    # The AKIA/ASIA rule below catches the key *IDs*; this one catches the
+    # *material* those IDs unlock. STM-origin: this rule is ahead of LTM's
+    # mirrored set until the forward-sync lands (the inverse of the
+    # #1488→#1491 reverse-sync direction; see the count-pin note in
+    # tests/test_privacy.py).
+    r"(?i)(?:[\"'](?:secret_?access_?key|session_?token)[\"']\s*:\s*[\"']"
+    r"|(?:secret[_-]?access[_-]?key|session[_-]?token)\s*[:=])",
     # Provider-prefixed token formats. Anchored by prefix so false positives
     # on arbitrary high-entropy strings are rare.
     r"(?i)(sk-[a-zA-Z0-9]{20,}|ghp_[a-zA-Z0-9]{36}|xox[bps]-[0-9A-Za-z-]+)",
--- a/src/memtomem_stm/proxy/privacy.py
+++ b/src/memtomem_stm/proxy/privacy.py
@@ -53,8 +53,17 @@ CREDENTIAL_PATTERNS = [
     #   pre-existing label rules never fire on bare JSON keys (the closing
     #   quote blocks their ``[:=]``); this form must not regress that.
     # - unquoted form — ``AWS_SECRET_ACCESS_KEY=`` / ``aws_session_token =``
-    #   (env output, ~/.aws/credentials INI): same shape and FP profile as
-    #   the generic label rule above.
+    #   (env output, ~/.aws/credentials INI). Unlike the generic label rule
+    #   above, this branch carries a LEFT boundary: the label must start a
+    #   token (``SESSION_TOKEN=``, namespaced ``TF_VAR_aws_secret_access_key=``)
+    #   or directly follow an ``aws`` separator (``aws.secret_access_key =``,
+    #   TOML dotted key). Without it, identifiers that merely EMBED the label
+    #   fire the rule — ``get_session_token: unhealthy`` (method name),
+    #   ``supports_session_token: true`` (capability flag),
+    #   ``rotateSecretAccessKey: done`` — and the scanning consumers mask
+    #   benign tools/records. A plain optional ``(?:aws[_-])?`` prefix
+    #   instead of the lookbehind pair would drop the namespaced env-var
+    #   forms above, so don't "simplify" it that way.
     #
     # The AKIA/ASIA rule below catches the key *IDs*; this one catches the
     # *material* those IDs unlock. STM-origin: this rule is ahead of LTM's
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):

memtomem (commented):
## Review — multi-angle pass with adversarial verification

Ran 8 finder angles (correctness ×3, cleanup ×3, altitude, conventions) over the diff → 22 candidates → dedup to 8 → one adversarial verifier each, with empirical `re` checks against every pinned sample. **4 findings survived** (3 CONFIRMED, 1 PLAUSIBLE); 4 were refuted. Overall this is a well-built change — the FP-guard design holds exactly as scoped, and most refuted candidates were things the PR body had already anticipated.

### Surviving findings (most severe first)

**1. `src/memtomem_stm/proxy/privacy.py:61` — the quoted-JSON label mechanism is AWS-only; the identical gap remains for the rest of the label vocabulary** (CONFIRMED)
`"password": "hunter2"`, `"api_key": "sk-abc"`, `"access_token": "ya29.x"`, and dict-repr `{'password': 'hunter2'}` match **zero** base patterns and zero branches of the new rule — JSON-serialized generic credentials (docker inspect, `kubectl get secret -o json`, DB connection configs) still scan clean. Notably, widening the quoted branch's vocabulary **inside this same STM-origin rule** leaves the mirrored 16 untouched (no LTM-mirror mutation), and doing it before the LTM forward-sync avoids a double sync. Verified: appending a generalized quoted-label rule passes the full test file (66 passed; only the count pin moves). Fine as an immediate follow-up if out of scope here — but it should be filed before the forward-sync lands.

**2. `src/memtomem_stm/proxy/privacy.py:61` — separator asymmetry between branches: quoted is underscore-only (`_?`), unquoted allows `[_-]?`** (CONFIRMED)
`"session-token": "AQoDYXdz..."` (serialized header dicts) matches neither branch nor any base pattern. Unifying both branches to `[_-]?` was verified to keep **every** pinned outcome identical (all 4 new positives, all 5 new negatives, all 14 pre-existing negatives) while strictly widening coverage — a free fix worth taking in this PR. Related follow-up: `x-amz-security-token: ...` — the actual AWS wire header carrying session-token material, emitted verbatim in botocore DEBUG logs — is covered by no pattern on either side.

**3. `tests/test_privacy.py:158` — two comment-claimed behaviors are unpinned** (CONFIRMED)
Both unquoted positives use `=` and both quoted positives use double quotes, so the unquoted `:` separator (`aws_session_token: Fwo...`, YAML) and the single-quote dict-repr form (`{'SessionToken': 'Fwo...'}` — which the source comment explicitly claims via "python-dict repr") are exercised by nothing. Verified: tightening `[:=]`→`=` or `["']`→`"` keeps the whole suite green while silently breaking exactly those claimed behaviors. Two added positive samples close it. Optional hardening: the quoted branch accepts mismatched quote pairs (`"SessionToken': "...`) despite the comment's "quote DIRECTLY on both sides" contract — a backreference (`(["'])label\1`) enforces it with zero pinned-outcome changes.

**4. `src/memtomem_stm/proxy/cache.py:258` — label-only AWS API-reference boilerplate becomes permanently uncacheable (+ startup purge)** (PLAUSIBLE)
`"SessionToken": "string"`-style API-doc blocks and lone `aws_secret_access_key = ...` lines are the genuinely new uncacheable class (the canonical `~/.aws/credentials` docs block already tripped the adjacent-`AKIA` rule pre-PR, so the marginal breadth is smaller than it first looks). The fail-safe treatment itself is announced in the PR body; suggest one sentence on this breadth in the already-planned changelog follow-up.

### Refuted during verification (for the record)

- **Eligibility tool-hiding FP on `session_token:` docstrings** — real mechanism, but the identical class already fires today on `api_key:`/`password:` param docs, and the inline comment's must-not-fire promise is scoped to exactly the two pinned shapes (both hold). Deliberate, in-family. One nit: the **Behavior change** section could mention the (policy-consistent) tools/list consequence alongside routing + cache.
- **Escaped-JSON (double-encoded) miss** — realistic STS payloads still flag via the co-occurring `ASIA` key-ID rule, and the scan input is decoded `content.text`, so the canonical MCP shape is single-encoded (which the new rule catches).
- **Sync-window is prose-only** — refuted by #559 (content-hash pin), which is the strictly stronger tripwire.
- **Quoted-key + unquoted-value miss** — shared structural trait of every label rule in the file; the canonical carriers (INI, STS JSON) are covered.

### Clean checks

Linear-time confirmed empirically (1 MB adversarial inputs, ≤1.4× the existing label rule; no superlinear unit) · ruff check/format pass on the head branch · all five FP guards hold · cache staleness is self-healing (startup purge + read-rescan — no pre-upgrade poisoned rows survive) · no duplication of existing coverage.

**Recommendation**: take #2 (free) and #3 (two samples) in this PR; #1 as an in-PR widening or a pre-forward-sync follow-up issue; #4 folds into the planned changelog PR.

memtomem:
Applied review findings #2 and #3 in e28e11c: unified the `[_-]?` separator charset across both branches (kebab quoted keys like `"session-token": "…"` now covered; every previously pinned outcome verified unchanged) and pinned the unquoted `:` separator (YAML), the single-quote dict-repr form, and the new kebab coverage as positive samples. Full suite: 3990 passed, 3 skipped; ruff clean. Finding #1 (generalizing the quoted-label mechanism beyond AWS) is left as the pre-forward-sync follow-up per the review.

memtomem:
## Codex review pass — verdict NEEDS-FIX → addressed in ca8a763

Codex surfaced one Major: **the unquoted branch had no left boundary**, so identifiers that merely embed the label fired the rule. All five counterexamples reproduced empirically: `aws.get_session_token: unhealthy`, `get_session_token: unhealthy`, `supports_session_token: true`, `rotateSecretAccessKey: done`, `boto3.session_token: expired` → tool hidden / routing degraded / telemetry record dropped on benign method names and capability flags. (For the record: the generic base rules share this trait today — `get_access_token: unhealthy` and `rotate_api_key: done` fire on main — but a cheap targeted fix existed here, so it was taken rather than extending the class.)

**The prescribed fix was rejected after simulation.** Codex suggested `(?<![A-Za-z0-9_.-])(?:aws[_-]?)?<label>`. Verified: that form drops real namespaced material — `TF_VAR_aws_secret_access_key=…` (Terraform), `PROD_AWS_SESSION_TOKEN=…`, and TOML `aws.secret_access_key = "…"` all scan clean under it. A privacy rule regressing on the most common CI/IaC env-var namespacing is worse than the FP it fixes.

**Applied instead: a lookbehind pair** — token start `(?<![A-Za-z0-9_.-])` OR a directly preceding aws separator `(?<=aws[._-])`. Matrix over every shape: 7/7 pinned positives kept, 4/4 namespaced forms kept (incl. bare `SESSION_TOKEN=`), 5/5 codex counterexamples blocked, 5/5 pre-existing negatives unchanged. All five counterexamples pinned as negatives; the namespaced forms pinned as positives. Full suite: 3998 passed, 3 skipped; ruff clean.

Residual trade-off introduced by the fix (for transparency): a non-aws compound prefix like `customer_secret_access_key=…` no longer fires — inherent to any left boundary; the aws-separator carve-out covers the dominant real-world namespacing (`*_aws_…`/`*_AWS_…`), and semantically a differently-prefixed key is a different system's credential. Note for the LTM forward-sync: the mirrored pattern now contains fixed-width lookbehinds — fine for Python and for the JS translator (`to_js_pattern` passes lookbehinds through; ES2018+).

Issue #562, opened 2026-07-02
privacy: generalize the quoted-JSON label rule beyond AWS labels


memtomem:
## Background

#553 added quoted-JSON-key detection for AWS secret-material labels (`"SessionToken": "…"` / `"SecretAccessKey": "…"`), including the FP-guard design that makes the shape safe for the scanning consumers (quote directly on both sides of the label; value must open as a string; JSON-Schema object values and prefixed telemetry keys stay negative).

The adversarial review pass on #553 verified that the identical gap remains for the **rest of the label vocabulary**: `"password": "hunter2"`, `"api_key": "sk-abc"`, `"access_token": "ya29.x"`, and dict-repr `{'password': 'hunter2'}` match **zero** of the 17 patterns. JSON-serialized credentials are high-frequency tool output (`docker inspect`, `kubectl get secret -o json`, DB connection configs).

## Why the existing rules don't handle this

The pre-existing label rules end in `\s*[:=]`, and a quoted JSON key's closing quote sits between the label and the colon — verified they never fire on any quoted-key form. This is exactly the gap #553 closed for the two AWS labels only.

## Proposed change

One general quoted-label rule reusing #553's FP-guard shape, e.g.:

```
(?i)["'](?:api[_-]?key|secret[_-]?(?:access[_-]?)?key|(?:access|session)[_-]?token|password|passwd)["']\s*:\s*["']
```

Verified during the #553 review: a rule of this shape passes the full existing test file (only the count pin moves) including all five #553 FP guards.

**FP note to resolve in the PR**: `pwd` should likely be **excluded** from the quoted vocabulary — shell/file tools legitimately return `"pwd": "/home/user"` (working-directory fields), which would otherwise classify as credential-bearing. `"password": ""` placeholders will fire; that matches the accepted fail-safe profile of the AWS branch (`"SessionToken": ""` fires too).

## Why a separate PR

#553 is scoped to AWS secret material and is the subject of an LTM forward-sync; widening the vocabulary is STM-origin work with its own FP pin set, and should coordinate with the #559 content-hash pin so the LTM mirror moves once, not twice.

Issue #559, opened 2026-07-02
privacy: pin LTM<->STM secret-class pattern sync by content, not by count


memtomem:
## Background

`proxy/privacy.py` documents that its secret-class `CREDENTIAL_PATTERNS` are kept coherent with memtomem LTM's `DEFAULT_PATTERNS` (mirrored both ways; #1488 -> #1491 / #549). Today the sets are byte-identical.

## Gap

The only guard on each side is a **count** pin: STM's `test_credential_set_count_pinned` asserts `len(CREDENTIAL_PATTERNS) == N`, and LTM pins its own count. Nothing compares the actual regex strings. Someone tightening one side's Anthropic body (e.g. a length change after a key-format revision) leaves both suites green while the routing gate (STM) and the write-rejection gate (LTM) silently diverge — the exact coherence the module docstring promises.

STM deliberately has no Python dependency on `memtomem` core (CLAUDE.md invariant), so a cross-import equality test is off the table.

## Proposed direction

A committed golden SHA-256 over the newline-joined secret-class pattern strings, asserted on both sides (LTM already computes `JS_PATTERNS_SHA` for its JS translation — same trick, sync-scoped). Updating a pattern then forces a deliberate two-sided hash bump, and a one-sided edit fails that side's suite. The STM-side test documents the current expected divergence window (STM 17 vs LTM 16 until the AWS-label forward-sync lands) — the hash pin should land together with or after that sync so the sets are byte-identical again.

## Why separate

Coordination with the LTM repo (the hash constant and the sync note must land on both sides) — not something to ride along a pattern-addition PR.

Found during the 2026-07-02 post-release review sweep of #549.

Cross-refs: the STM-side AWS-label addition (the current 17-vs-16 divergence) is #553.

memtomem:
Status update: the 17-vs-16 divergence window is closed — memtomem-stm#553 merged (STM `5ab5467`) and the forward-sync landed on the LTM side as memtomem/memtomem#1533 (`5a6da273`), with the rule inserted at index 2 matching STM's order. Verified byte-identical at 17 patterns via an elementwise import-level comparison of both modules. The content-hash pin proposed here can now land with the sets equal on both sides.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #564, opened 2026-07-02
security(privacy): catch x-amz-security-token (header + presigned-URL query param)

### LINKED ISSUE #561: privacy: catch x-amz-security-token (header + presigned-URL query param)
## Background

#553 catches AWS secret material by config/JSON **label** (`SECRET_ACCESS_KEY` / `SESSION_TOKEN` spellings). The review pass verified that the label AWS actually puts on the wire for the same material is covered by **no** pattern on either side:

- `x-amz-security-token: FwoGZXIvYXdz…` — the HTTP header carrying the session token, emitted **verbatim in botocore DEBUG logs** and in any HTTP-header dump a proxying/debugging tool returns;
- `X-Amz-Security-Token=FwoGZXIvYXdz…` — the presigned-URL query-parameter form, present in every presigned S3 URL generated with temporary credentials.

Catching these payloads today relies entirely on an `ASIA…` key ID happening to co-occur in the same text.

## Why the existing rules don't handle this

`session[_-]?token` cannot cross the `security-token` spelling, and the kebab/header shape has no `aws` separator immediately before the label, so #553's left-boundary form doesn't reach it either (verified empirically during review).

## Proposed change

A dedicated header/query rule, e.g.:

```
(?i)x-amz-security-token["']?\s*[:=]
```

## Why a separate PR

The header shape has a different FP surface than config labels (header NAME mentions in API docs — e.g. "set the `x-amz-security-token` header" prose without a value — must be weighed; requiring the separator like the label rules keeps parity but the docs-prose colon case, `header: x-amz-security-token`, needs a negative pin). It deserves its own positive/negative pin set, and should coordinate with the #559 content-hash pin for the LTM mirror.

memtomem:
## Background

Closes #561.

#553 catches AWS secret material by config/JSON **label** (`SECRET_ACCESS_KEY` / `SESSION_TOKEN` spellings). The label AWS actually puts on the **wire** for the same material was covered by no pattern on either side:

- `x-amz-security-token: FwoGZXIvYXdz…` — the HTTP header carrying the session token, emitted **verbatim in botocore DEBUG logs** and in any HTTP-header dump a proxying/debugging tool returns;
- `X-Amz-Security-Token=FwoGZXIvYXdz…` — the presigned-URL query-parameter form, present in every presigned S3 URL generated with temporary credentials.

Catching these payloads previously relied entirely on an `ASIA…` key ID happening to co-occur in the same text.

## Why the existing rules don't reach it

`session[_-]?token` cannot cross the `security-token` spelling, and the kebab/header shape has no `aws` separator immediately before the label, so #553's left-boundary form doesn't reach it either (all four new positive samples verified to scan clean against the 17 patterns on main).

## The rule

One new `CREDENTIAL_PATTERNS` rule with two alternatives, mirroring #553's shape:

- **quoted form** — `"x-amz-security-token": "…"` (serialized header dicts, JSON or python repr). Quote directly on both sides of the label + a string-opening value, so an OpenAPI/JSON-Schema header *definition* (`"X-Amz-Security-Token": {"type"…` — object value, ubiquitous in AWS API specs) never fires and the scanning consumers (tool_eligibility, selection log) don't mask benign tools.
- **unquoted form** — the raw header line and the presigned-URL query param. The left boundary rejects only a directly preceding **separator** (`(?<![_.\-])`) — narrower than #553's boundary, on purpose (adjudicated with the codex cross-review; see the review comment below). Kebab/dotted compounds that merely name the header stay negative (`forward-x-amz-security-token: true` config knobs, `proxy.headers.x-amz-security-token` dotted telemetry heads — consistent with #553's dotted-heads-negative design), while a directly preceding *alphanumeric* stays a match: bytes-repr wire dumps render the newline before the header line as a literal `\r\n` (`send: b'…\r\nx-amz-security-token: FwoG…'` — `http.client` debuglevel / botocore DEBUG), putting `n` right before the `x`. Prose that merely NAMES the header stays negative — the separator must directly follow the label, which pins the issue's docs-prose case (`header: x-amz-security-token`) as a negative test, alongside a prose sentence and the OpenAPI object-value guard. A SigV4 canonical-request signed-headers *list* (`host;x-amz-security-token` with no value) also stays negative (verified; the header lines above it in a real canonical request carry the value and fire correctly).

Accepted fail-safe (same profile as every label rule): `x-amz-security-token=<placeholder>` in docs fires even with a placeholder value, matching how `password: <value>` behaves today.

## Sync bookkeeping

- `CREDENTIAL_PATTERNS` moves 17 → 18 (count pin bumped; the rule sits at index 3, directly after the #553 rule, so the LTM forward-sync should insert at the same index to keep the sets same-order).
- The #553 rule's "awaiting forward-sync" comments (privacy.py + the count-pin note) are updated in place to cite the landed memtomem#1533 — required anyway to rewrite the pin comment truthfully, and without it the file would claim two rules are awaiting sync when only this one is.
- The #559 content-hash pin should land together with (or after) this rule's forward-sync so the LTM mirror moves once — coordinated with #562, which widens the quoted-label vocabulary next.

## Behavior change

Responses carrying the `x-amz-security-token` label (botocore DEBUG output, presigned URLs, header dumps) now route away from external-LLM compression strategies and are excluded from the response cache — the same fail-safe treatment every other credential label already gets — and tool metadata carrying it is withheld from `tools/list` under the strict exposure profile.

## Testing

- 5 new positive pins (header line, presigned URL, JSON header dict, python-repr header dict, literal-`\r\n` bytes-repr wire dump) — each verified to scan clean without the new rule, so they pin it specifically.
- 5 new negative pins (docs-prose `header: x-amz-security-token`, prose sentence, OpenAPI header definition, kebab config knob, dotted telemetry head).
- 1 new credential-set-alone classification pin.
- `uv run pytest -m "not ollama"` and `uv run ruff check src` / `ruff format --check src` pass.

### DIFF
--- a/src/memtomem_stm/proxy/privacy.py
+++ b/src/memtomem_stm/proxy/privacy.py
@@ -66,13 +66,39 @@ CREDENTIAL_PATTERNS = [
     #   forms above, so don't "simplify" it that way.
     #
     # The AKIA/ASIA rule below catches the key *IDs*; this one catches the
-    # *material* those IDs unlock. STM-origin: this rule is ahead of LTM's
-    # mirrored set until the forward-sync lands (the inverse of the
-    # #1488→#1491 reverse-sync direction; see the count-pin note in
+    # *material* those IDs unlock. STM-origin; forward-synced into LTM's
+    # mirrored set as memtomem#1533 (the inverse of the #1488→#1491
+    # reverse-sync direction; see the count-pin note in
     # tests/test_privacy.py).
     r"(?i)(?:[\"'](?:secret[_-]?access[_-]?key|session[_-]?token)[\"']\s*:\s*[\"']"
     r"|(?:(?<![A-Za-z0-9_.-])|(?<=aws[._-]))"
     r"(?:secret[_-]?access[_-]?key|session[_-]?token)\s*[:=])",
+    # The label AWS actually puts on the WIRE for that same session-token
+    # material: botocore DEBUG logs emit the ``x-amz-security-token`` request
+    # header verbatim, and every presigned URL generated with temporary
+    # credentials carries the ``X-Amz-Security-Token=…`` query parameter.
+    # No rule above reaches either: ``session[_-]?token`` cannot cross the
+    # ``security-token`` spelling, and the kebab header shape has no ``aws``
+    # separator directly before the label, so the left-boundary form above
+    # misses it too — catching these payloads otherwise relies on an
+    # ``ASIA…`` key ID co-occurring in the same text. Two alternatives,
+    # mirroring the rule above:
+    #
+    # - quoted form — ``"x-amz-security-token": "…"`` (serialized header
+    #   dicts, JSON or python repr). Quote directly on both sides of the
+    #   label and a string-opening value, so an OpenAPI/JSON-Schema header
+    #   *definition* (``"X-Amz-Security-Token": {"type"…`` — object value,
+    #   ubiquitous in AWS API specs) never fires.
+    # - unquoted form — the raw header line (``x-amz-security-token: FwoG…``)
+    #   and the presigned-URL query param (``…&X-Amz-Security-Token=FwoG…``).
+    #   No left boundary, unlike the rule above: the label is kebab-case, so
+    #   snake_case/camelCase identifiers cannot embed it, and its real left
+    #   neighbors (``?``, ``&``, quotes, line starts) are all non-word.
+    #   Prose that merely NAMES the header (``set the x-amz-security-token
+    #   header``, ``header: x-amz-security-token``) stays negative — the
+    #   separator must directly follow the label.
+    r"(?i)(?:[\"']x-amz-security-token[\"']\s*:\s*[\"']"
+    r"|x-amz-security-token\s*[:=])",
     # Provider-prefixed token formats. Anchored by prefix so false positives
     # on arbitrary high-entropy strings are rare.
     r"(?i)(sk-[a-zA-Z0-9]{20,}|ghp_[a-zA-Z0-9]{36}|xox[bps]-[0-9A-Za-z-]+)",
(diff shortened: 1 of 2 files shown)


pandas-studio (approved):

memtomem:
## Codex cross-review + adjudication

A codex review pass on this diff returned one Major finding (no Blockers):

> The unquoted branch has no left boundary, so it matches embedded kebab strings such as `not-x-amz-security-token: disabled` or `docs-x-amz-security-token=placeholder`. Add a left guard, e.g. `(?<![A-Za-z0-9_.-])x-amz-security-token\s*[:=]`.

**Finding: confirmed.** The original comment's claim ("snake_case/camelCase identifiers cannot embed it") was true but incomplete — *kebab/dotted* compounds can embed the label, and the realistic instance is a config knob **about** the header (`forward-x-amz-security-token: true`), which would mask a benign tool's metadata under the strict exposure profile.

**Prescribed fix: rejected as written; a narrower boundary applied instead.** The full class `(?<![A-Za-z0-9_.-])` regresses a real positive named by the issue itself: a bytes-repr wire dump renders the newline before the header line as a **literal `\r\n`**, so an alphanumeric (`n`) directly precedes the `x` —

```
send: b'GET /k HTTP/1.1\r\nx-amz-security-token: FwoG…\r\n'
```

— which is exactly what `http.client` debuglevel / botocore DEBUG emit. Verified empirically: the prescribed variant misses that sample; the applied variant catches it.

**Applied:** `(?<![_.\-])` — reject only a directly preceding *separator*, keep alphanumerics matching. This kills the kebab-compound FP class (`forward-x-…`, `not-x-…`) and the dotted-head class (`proxy.headers.x-amz-security-token` — consistent with the #553 rule's dotted-heads-are-negative design, cf. its `boto3.session_token` pin) while keeping the escaped-`\r\n` dump positive. Both variants were run against the full pos/neg matrix; three new test pins land with the fix (bytes-repr dump positive, kebab-knob negative, dotted-head negative).

Issue #576, opened 2026-07-03
cache: ProxyCache.get() sqlite fault on the hot path fails every cache-eligible call — degrade to MISS


memtomem:
## Background

Operational-stability audit (2026-07-03). The response cache is best-effort **on the write side**: `_store_cache` wraps `set()` in `try/except` (`manager.py:3365-3379`, "response unaffected"), and degraded responses skip the cache (#496 lesson). The read side is not:

- `cache.py:177-181` — `ProxyCache.get()` runs the SELECT with no `except sqlite3.Error` (only the privacy-eviction DELETE inside it is guarded, `cache.py:202-212`).
- `manager.py:2307` / `:2317` — both call sites are bare.

## Failure scenario

A runtime `sqlite3.OperationalError` / `DatabaseError` on the lookup (disk I/O error, page corruption appearing mid-session, an external process holding the file past the busy timeout) propagates out of the call pipeline → the proxied tool call fails with INTERNAL_ERROR even though the upstream is healthy and the cache is optional. A persistently broken cache DB takes down every cache-eligible call until restart.

`GraphConsultCache.get()` already guards exactly this class ("must degrade to a plain MISS, never raise", `toolgraph_cache.py:139-159`) — the response cache lacks the same guard.

## Suggested fix

Catch `sqlite3.Error` in `ProxyCache.get()` (or at the two call sites), log at WARNING once/rate-limited, and return `None` (MISS) so the call proceeds to the upstream. Mirror the `GraphConsultCache` pattern.

Severity: medium-high — trigger is rare, but the failure inverts the cache's contract (an optimization becoming a single point of failure on the hot path).

Related: unwrapped store `initialize()` at startup is tracked separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #575, opened 2026-07-03
server: unwrapped metrics/cache store initialize() crashes startup on corrupt DB; check-then-ALTER migration races across processes


memtomem:
## Background

Operational-stability audit (2026-07-03). The lifespan wraps most store inits in a degrade-to-disabled `try/except` (compression feedback `server.py:150-160`, progressive reads `:167-176`, selection log `:183-197`, feedback tracker, toolgraph cache in `manager.py:616-629`). Exactly two are unwrapped:

- `server.py:143` — `metrics_store.initialize()`
- `server.py:267` — `proxy_cache.initialize()`

## Gap

A failure in either init raises out of `app_lifespan`, so the MCP server never starts — **every proxied tool goes down because a telemetry/cache DB is unhealthy**. Reachable failure modes:

1. **Corrupt DB file** (non-SQLite bytes): first PRAGMA in `tune_connection` raises `sqlite3.DatabaseError: file is not a database`. Zero-byte files are fine (SQLite reinitializes); corrupt bytes are not. There is no quarantine/recreate fallback.
2. **Concurrent-migration race**: `metrics_store.py:243-285` does check-then-ALTER (`PRAGMA table_info` → `ALTER TABLE ADD COLUMN`) with no cross-process guard. Two `mms` processes starting right after an upgrade (the normal multi-session Claude Code shape) can both read `table_info` before either ALTERs; the loser gets `sqlite3.OperationalError: duplicate column name` → startup crash. The `mms hook` path runs the same `_migrate` but is fully wrapped (`hook_cmd.py:650-651`) and degrades.

## Suggested fix

Wrap both inits in the same degrade-to-disabled pattern as their siblings (metrics → `metrics_store = None`, cache → `proxy_cache = None`, WARNING log). This also neuters the migration race's blast radius from "server down" to "metrics disabled for one session". Optionally treat `duplicate column name` as benign inside `_migrate` (another process won the race).

## Why separate

Fix is confined to `server.py` lifespan (+ a small `_migrate` tolerance); the request-path guards for `ProxyCache.get()` are a different fix site and tracked separately.

Severity: high — corrupt-file trigger is rare but the blast radius is total; the migration race trigger is the deployment's normal shape after every upgrade that adds a column.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #589, opened 2026-07-03
fix(compression): degrade SELECTIVE/HYBRID store-write failure to truncate, don't discard the response (#577)

### LINKED ISSUE #577: compression: SELECTIVE/HYBRID pending-store failure discards a successful upstream response (sqlite backend)
## Background

Operational-stability audit (2026-07-03). The PROGRESSIVE branch of `_compress_and_surface` has an explicit passthrough guard whose comment states the design intent (`manager.py:2631-2634`): *"an escape here is recorded as INTERNAL_ERROR and DISCARDS an otherwise-successful upstream response"*. The ratio-guard fallback tiers got the same guard (`:2764+`).

## Gap

The SELECTIVE/HYBRID path did not:

- `manager.py:2714` — `await self._apply_compression(...)` is bare.
- Inside it, SELECTIVE (`:1493`) and HYBRID (`:1477-1481`) reach `SelectiveCompressor._store_and_build_toc` → `self._store.put(...)` (`compression.py:892`).
- `SQLitePendingStore` has zero `sqlite3.Error` handling in any method (`pending_store.py:130-203`; only corrupt `chunks_json` is treated as a miss).
- The lazily-built store's `initialize()` is also unguarded on the request path (`manager.py:1331` `_create_selective`, `:1804` `_get_progressive_store`).

## Failure scenario

With the opt-in `pending_store: "sqlite"` backend (its stated purpose is multi-instance sharing), a writer holding the lock past the 3s busy timeout — or disk-full, or a corrupt pending DB — makes `put()` raise; the exception escapes `_call_tool_inner`, is recorded as INTERNAL_ERROR, and the already-fetched successful upstream response is thrown away.

Related (lower): `read_more` (`manager.py:1880-1881`) and `select_chunks` (`compression.py:974-980`) and `CompressionFeedbackStore.record` (`compression_feedback_store.py:135-142`) are also unguarded — a locked DB fails those tool calls, though without discarding an upstream response.

## Suggested fix

Give the SELECTIVE/HYBRID store writes the same degrade contract as PROGRESSIVE: on store failure, fall back to `truncate` (mirroring the LLM `no_config` fallback at `manager.py:1530-1540`) and make sure the degraded response skips the cache via the existing `passthrough_on_error`-style flag.

Severity: medium (conditional on the opt-in sqlite pending backend; the default is `memory`).

memtomem:
## Summary

Closes #577 (the response-discard gap). The PROGRESSIVE branch of `_compress_and_surface` wraps its store-touching call and degrades to a zero-loss passthrough on failure — its comment states the intent: *"an escape here is recorded as INTERNAL_ERROR and DISCARDS an otherwise-successful upstream response"*. The ratio-guard fallback tiers got the same guard. **The SELECTIVE/HYBRID primary path did not.**

The `_apply_compression` call on the compress branch (`manager.py:~2714`) was bare, and `SQLitePendingStore` has no `sqlite3.Error` handling in any method (`put`/`get`/`touch`/`evict_*` all call `_get_db().execute()` raw). So under the opt-in `pending_store: "sqlite"` backend (its stated purpose is multi-instance sharing), a pending-store write fault — a writer holding the lock past the 3s busy timeout, disk-full, or a corrupt DB — raised straight out of `_call_tool_inner`, was recorded as INTERNAL_ERROR, and **threw away the already-fetched successful upstream response.**

## Change

- **`manager.py`** — wrap the primary `_apply_compression` call in `sqlite3.Error` (the only DB in that path, so the catch is scoped to the store fault and can't mask a compressor logic error). On failure, degrade to a boundary-aware `TruncateCompressor` at the retention budget — the same terminal tier the ratio-guard ladder already falls back to — skip the now-redundant ratio ladder, and label metrics `selective→truncate_on_store_error`. Mirrors the PROGRESSIVE passthrough guard.
- **`pipeline_stages.py`** — new `CompressionResult.selective_store_error` flag (defaults `False`).
- **`manager.py` cache gate** — the degraded truncation skips the cache the same way `progressive_passthrough_on_error` does, so a transient store failure can't pin the lossy truncation for the cache TTL and suppress the chunk-TOC protocol on identical calls after the store recovers (the #496 lesson). Records `cache_unstorable`.

## Behavior change

On a SELECTIVE/HYBRID pending-store write fault, the call now returns a boundary-aware truncation (one WARNING, metrics `…→truncate_on_store_error`, not cached) instead of failing the tool call and discarding the upstream response. No change on a healthy store, or for the default `pending_store: "memory"` backend.

## Scope

Closes the response-discard gap only. The lower-severity `stm_proxy_read_more` / `stm_proxy_select_chunks` / `CompressionFeedbackStore.record` unguarded store reads noted in #577 — which fail their *own* tool call without discarding an upstream response — are left as follow-ups.

## Tests

`test_compression_ratio_guard.py` — two new tests mirroring the PROGRESSIVE ones: a store `sqlite3.OperationalError` degrades to truncation (no `"selection_key"` TOC, metrics label asserted, no raise); and the degraded truncation is not cached (next identical call is a MISS that re-runs upstream and re-attempts the real SELECTIVE TOC). Full suite: 4080 passed, 3 skipped.

## Series

Third of the ops-stability audit fixes (2026-07-03). #575 merged (#587); #576 open (#588).

### DIFF
--- a/src/memtomem_stm/proxy/manager.py
+++ b/src/memtomem_stm/proxy/manager.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 import functools
 import logging
+import sqlite3
 import time as _time
 import uuid
 from collections import Counter
@@ -2614,6 +2615,11 @@ class ProxyManager:
         # passthrough after a store failure (below). Read at the cache-store
         # gate to keep that transient-failure-degraded response out of the cache.
         progressive_passthrough_on_error = False
+        # Set True when the SELECTIVE/HYBRID path degrades to a boundary-aware
+        # truncation after a pending-store write failure (below). Defined here
+        # (not only in the else-branch) because the shared CompressionResult
+        # construction at the end reads it on the PROGRESSIVE path too.
+        selective_store_error = False
         if tc.compression == CompressionStrategy.PROGRESSIVE and tc.progressive:
             pcfg = tc.progressive
             if len(cleaned) <= pcfg.chunk_size:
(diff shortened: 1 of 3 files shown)


pandas-studio (approved):

memtomem:
Addressed the codex review in 4eb3794:
- **Major (over-broad catch)**: the `except sqlite3.Error` now re-raises unless `effective_compression` is SELECTIVE or HYBRID — sqlite is only reachable there via the pending-store write, so any other strategy's sqlite error keeps its prior INTERNAL_ERROR behavior instead of being relabeled as a store degradation.
- **Major (tests didn't exercise the real path)**: added an end-to-end test using the real sqlite pending backend with `SQLitePendingStore.put` forced to raise (drives `_create_selective` → `compress` → `put`), plus a negative test that a `sqlite3.Error` from a non-SELECTIVE/HYBRID strategy is not converted.
- **Nit**: removed the `selective_store_error` default to honor the dataclass's "no defaults for branch-set fields" contract.

Full suite: 4082 passed, 3 skipped.

Issue #600, opened 2026-07-03
circuit breaker: is_open/state are state-mutating property reads — stm_proxy_health flips the breaker and misreports it


memtomem:
## Background

Operational-stability audit (2026-07-03, low-severity backlog re-verified). `CircuitBreaker` (`src/memtomem_stm/utils/circuit_breaker.py`) exposes `is_open` and `state` as `@property` reads, but the **read itself performs the open→half-open transition as a side effect**:

```python
@property
def is_open(self) -> bool:
    if self._state == "closed":
        return False
    if self._state == "open" and time.monotonic() - self._opened_at >= self._reset_timeout:
        self._state = "half-open"   # <-- state mutated inside a property READ
        return False
    return self._state == "open"
```

`state` (`:41-47`) mutates the same way.

## Gap

`stm_proxy_health` renders the surfacing breaker by **reading** `is_open` (`server.py:784-786`):

```python
cb = surfacing._circuit_breaker
cb_state = "open (failing)" if cb.is_open else "closed (healthy)"
```

So once the breaker is open and `reset_timeout` has elapsed, an operator running `stm_proxy_health` **flips the breaker to half-open as a side effect of the health check** — and then the same read returns `False`, so the line prints `closed (healthy)` even though the dependency is still down. A read that both mutates and misreports the state it claims to observe.

Separately, half-open admits **every** concurrent caller: there is no in-flight probe counter, so after the transition every `is_open` read hits `return self._state == "open"` → `False`. The class docstring (`:16`) claims "one probe call allowed", but the code doesn't enforce it. (This half is largely theoretical for the single-MCP-client proxy, where calls to a given path are effectively serialized — the mutating-read defect is the real one.)

## Failure scenario

Surfacing breaker opens after `circuit_max_failures` (default 3) LTM failures. `circuit_reset_seconds` (default 60s) elapses while the LTM is still down. An operator runs `stm_proxy_health` to check: the render transitions the breaker to half-open and prints `closed (healthy)`. The operator concludes the LTM recovered; meanwhile the very next surfacing call is the probe that would otherwise have been the gated half-open probe.

## Suggested fix

- Make `is_open` / `state` **pure** — compute the effective state (accounting for elapsed `reset_timeout`) without assigning `self._state`. Perform the actual open→half-open transition only in the admission path a real call takes (e.g. an `allow_request()` / the surfacing gate), not in a property read.
[...]

Issue #578, opened 2026-07-03
proxy: timeout-retry replays non-idempotent upstream tools — retry loop never consults tool annotations


memtomem:
## Background

Operational-stability audit (2026-07-03). The upstream retry loop (`manager.py:2365-2496`) retries on `(OSError, ConnectionError, asyncio.TimeoutError, EOFError)` plus MCP errors outside `_NO_RETRY_CODES`, reconnecting between attempts. Per-call it is textbook: exponential backoff, overall deadline, failure taxonomy.

## Gap

On `asyncio.TimeoutError` the upstream call **may already have executed its side effect** — the proxy cancels its wait, reconnects, and re-invokes the same tool up to `max_retries` times. The manager already knows which tools self-declare as writers (`_tool_cache_eligible` reads `readOnlyHint`/`destructiveHint`, `manager.py:3204-3222`) but the retry loop never consults annotations.

## Failure scenario

A slow upstream write tool (e.g. an LTM `mem_add`-style call) commits its mutation, then the per-attempt timeout fires → the retry replays the mutation. With `max_retries=3`, one slow write can be applied up to 4 times. A transparent proxy retrying blind is a semantic change relative to the client calling the upstream directly (the client would see one timeout, not N replayed writes).

Note this is distinct from *client-level* retries, which are documented as inherent ("every call re-runs and its side effect re-executes", `manager.py:2296-2304`) — this is the proxy itself manufacturing duplicate writes.

## Suggested fix (design discussion welcome)

Gate **timeout**-retry (not connection-refused/EOF-before-send retry, where the request provably never completed) on the same annotation policy the cache uses: tools with `readOnlyHint=false`/unknown or `destructiveHint=true` get at most the reconnect-for-next-call treatment, then propagate the timeout. Needs a `## Behavior change` callout when implemented.

Severity: medium-high in write-heavy setups; the fix touches retry semantics so it deserves its own focused PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #558, opened 2026-07-02
metrics: cache hits are invisible in total_calls and savings; hit-rate uses a different denominator


memtomem:
## Background

#536 added cache hit-rate/occupancy to `stm_proxy_stats` / `mms stats`. The counters live in `TokenTracker` (`metrics.py`): `record()` increments `_total_calls` / `_total_original` / `_total_compressed`, and `record_cache_hit()` / `record_cache_miss()` increment the lookup counters.

## Gap (three related accounting mismatches)

1. **`total_calls` undercounts by exactly the number of cache hits.** `_on_cache_hit` (`manager.py`) calls only `record_cache_hit()` then `_apply_surfacing` — it never calls `tracker.record(...)`, which is reached only on the live-miss/passthrough paths. An operator reading "Total calls: 100 / Cache hits: 40" cannot reconcile the numbers with actual tool invocations (140).
2. **Savings % excludes cache hits**, which are the *maximal* saving (zero upstream I/O, zero re-compression). The cache's headline benefit is structurally invisible in the compression-savings metric it sits next to.
3. **Mixed-content read-only tools can never register a hit but always record a miss.** `_store_cache` correctly refuses to store a response with `non_text_content`, but the miss was already recorded on the eligible-miss path — so a tool that *always* returns text+image re-misses forever: `cache_misses` grows unbounded, its hit-rate contribution is a permanent 0%, and the diagnostics point at a tool the cache can never help.

## Why not a quick fix in the metrics renderer

The three exclusions live at different layers (hit path, savings aggregation, miss accounting vs. store rejection) — a display-side patch would paper over denominators that still disagree. Needs a small design pass: either count hits into `total_calls`+savings with an explicit `served_from_cache` component, or render an explicit `total_invocations = total_calls + cache_hits` line and an `unstorable` counter so every number is reconcilable.

## Why separate

Observability-only (no wrong caching behavior), and #536's counters are test-pinned — changing semantics deserves its own PR with the counter-snapshot assertions updated deliberately, not as a rider.

Found during the 2026-07-02 post-release review sweep, alongside real-data checks (`proxy_metrics.db` / `proxy_cache.db`) of the #534–#550 cache work.

PR #620, opened 2026-07-04
feat(proxy): per-upstream circuit breaker on the call_tool path

### LINKED ISSUE #608: proxy: upstream call_tool path has no circuit breaker — SECURITY.md claims per-upstream breaker isolation
## Why now

A full-service review (2026-07-04, follow-up to the ops-stability audit that shipped in v0.1.31) cross-checked SECURITY.md's resilience claims against the code. One claim does not hold:

> **Circuit breaker**: Per-upstream circuit breaker isolates failures; a misbehaving upstream cannot cascade into other proxied tools. (SECURITY.md:32)

`src/memtomem_stm/proxy/manager.py` contains **zero** `CircuitBreaker` usage (`grep -c "CircuitBreaker\|circuit_breaker"` → 0). The breaker in `utils/circuit_breaker.py` protects LLM compression (`compression.py`), LLM extraction (`extraction.py`), and LTM surfacing (`surfacing/engine.py`) — but not the upstream MCP `call_tool` path.

## Current behavior

Upstream failures are bounded only by per-attempt `call_timeout_seconds` (default 90s), `connect_timeout_seconds` (30s), and bounded retries under `overall_deadline_seconds` (default 180s), with the idempotency-gated retry guard from #578. That time-bounds a *hung* upstream, but a *persistently failing* upstream is retried on every call — each caller pays the full timeout/retry cost indefinitely, and there is no fast-fail or per-upstream isolation.

## Doesn't the reconnect path already handle this?

No — reconnect (#586) restores a broken session but does not stop routing calls to an upstream that keeps failing. The breaker semantics (closed → open → half-open, pure reads per #600) exist in `utils/circuit_breaker.py` and are already battle-tested on three other dependencies; the upstream path is the one dependency without it.

## Proposal

Add a per-upstream `CircuitBreaker` (one instance per upstream server, reusing `utils/circuit_breaker.py`) around `call_tool` dispatch in `ProxyManager`:

- open after N consecutive failures → fast-fail with a clear MCP error naming the upstream and time-until-retry
- half-open probe on the next call after the reset timeout
- expose per-upstream breaker state in `stm_proxy_health` (alongside the existing surfacing breaker line) and `mms health`

Alternative (if we decide the retry/deadline bounds are sufficient): amend SECURITY.md to describe the actual mechanism instead. Either way the doc and the code should agree; implementing the breaker is the recommended direction since the primitive already exists.

## Why separate

This is a behavior change on the hot path (new fast-fail failure mode) and deserves its own PR with breaker-state tests, not a ride-along on unrelated hardening.

memtomem:
Closes #608.

## Why

SECURITY.md §Resilience claims "Per-upstream circuit breaker isolates failures", but the upstream `call_tool` path had no breaker — a persistently-failing upstream was retried on every call, each caller paying the full retry/deadline budget (default up to 180s) indefinitely. This PR wires the existing 3-state pure-read `CircuitBreaker` (#600) per upstream connection, making the doc claim true.

## Design decisions

- **Breaker lives on `UpstreamConnection`** — `_reconnect_server` mutates the connection in place (pinned invariant), so breaker state survives reconnects. Built once in `_connect_server`; `circuit_max_failures: 0` skips construction (disabled).
- **One failure count per call, not per attempt.** Only the three terminal transport/timeout exits count: retries exhausted / replay-unsafe (#578), overall deadline exceeded, and a mid-loop reconnect failure (without the last one, a dead stdio upstream whose respawns fail would escape breaker counting entirely). The retry-continue path records nothing.
- **`isError` results record breaker *success*.** A completed round-trip proves the upstream is alive; tool-level errors are a tool-health signal (already `UPSTREAM_ERROR`), not a dependency-health signal. Protocol (-326xx) and programming errors never count either way.
- **Fast-fail sits after the cache fast-path**, so cached responses keep serving while the breaker is open, and other upstreams are unaffected (isolation, pinned by test).
- **New `ErrorCategory.CIRCUIT_OPEN`** keeps the #558 counter reconciliation (`total_invocations = calls + cache_hits + errors`, exactly one `record_error` per fast-fail, `_mark_recorded` prevents the outer INTERNAL_ERROR double-count). It is added to `UPSTREAM_ERROR_CATEGORIES`: fast-fail rows are upstream-attributable by construction — excluding them would inflate the call-count denominator in `get_tool_error_stats` and *dilute* per-tool error rates during exactly the window the upstream is down.
- **Half-open follows the surfacing breaker's posture**: `is_open` is a pure read that turns false once `circuit_reset_seconds` elapses, so the next call proceeds as the probe (no single-probe serialization, same as #600 documented).

## Behavior change

New fast-fail failure mode, **on by default** (`circuit_max_failures: 3`, `circuit_reset_seconds: 60.0`): after 3 consecutive failed calls, an upstream's tools fail fast with a `circuit_open` `ToolError` naming the upstream and time-to-probe, for up to 60s per window, instead of each call burning the full retry/deadline budget. Opt out per upstream with `circuit_max_failures: 0`. Config is a connect-time snapshot like `max_retries` (restart to change; documented in the hot-reload table). `stm_proxy_health` gains a per-upstream `circuit breaker:` line (absent when disabled or startup-failed, so absence never reads as healthy).

## Tests

- `tests/test_proxy_error_paths.py` — new `TestUpstreamCircuitBreaker` (12 cases): open-after-N + fast-fail without touching the session; counter reconciliation (`circuit_open`/`transport`/`internal_error` + `total_invocations`); one-count-per-call under retries; success reset; isError-as-success; protocol/programming excluded; half-open probe success/failure (time-patched); overall-deadline terminal counting; disabled breaker; per-upstream isolation; cache-hit-while-open. `_make_manager` defaults to `circuit_max_failures=0` so all pre-existing error-path tests keep exercising raw retry mechanics.
- `tests/test_server_tools.py` — per-upstream health line in all three states (+ `retry in ~Ns` only while open), and absence for disabled/startup-failed servers.
- `tests/test_config_constraints.py` — field bounds and defaults.

## Verification

- Full suite: 4169 passed, 3 skipped (`-m "not ollama"`); ruff clean; mypy unchanged (same 9 pre-existing errors).
- E2E over a real stdio transport (FastMCP upstream whose tool hangs, `call_timeout=0.5s`): calls 1-2 time out terminally (~3s each incl. respawn), breaker opens at 2, call 3 fast-fails in **0.00s** with `Upstream 'slow' temporarily unavailable: circuit breaker open after 2 consecutive failures; retry in ~57s`, and `get_upstream_health()` reports `circuit_state: open, circuit_reset_in: 57.4`.


pandas-studio (approved):

pandas-studio:
Addressed a codex review finding in df84e42.

**Blocking (high) — fixed**: the JSON-RPC protocol-error branch (`err_code in _NO_RETRY_CODES`) recorded neither success nor failure on the breaker, contradicting the completed-round-trip design used for `isError` results. A protocol error is a reply *from* the upstream — the transport is alive — so transport/timeout failures separated by a protocol response would still accumulate as "consecutive" and could spuriously open the breaker. The branch now calls `record_success()`, resetting the streak, exactly like an `isError` result. Programming/internal exceptions (no round-trip) still record neither.

**Test findings — fixed**:
- Added `test_protocol_round_trip_resets_failure_streak`: two transport failures → protocol reply resets to 0 → next transport failure starts a fresh streak (the scenario the bug broke).
- Added `test_mid_loop_reconnect_failure_records_breaker_failure`: pins the third terminal exit (retryable transport error whose follow-up reconnect itself fails) records exactly one breaker failure — previously uncovered.

Docs/CHANGELOG wording updated to state that any completed round-trip (tool `isError` **or** protocol error) resets the streak. Full suite green, E2E re-verified (fast-fail still 0.00s).

Issue #611, opened 2026-07-04
config: unknown keys are silently ignored and parse failures silently fall back — add `mms config validate` and louder failure


memtomem:
## Why now

Highest-impact usability gap from the 2026-07-04 service review. Two compounding dark-failure modes:

1. **Typo'd keys vanish.** No config model sets `extra="forbid"` (Pydantic default `extra="ignore"` is explicitly relied on — proxy/config.py:584 comment). `"max_result_char"` (missing `s`) or a misspelled nested block is dropped with no error and no warning, across a surface of 100+ tunable fields.
2. **Broken config degrades to "proxy off".** `ProxyConfig.load_from_file` returns `None` on parse/validation failure with only a `logger.warning` to stderr (proxy/config.py:1131-1138); the server then runs with defaults, where the proxy is `enabled=False`. For an MCP-launched server, stderr is buried in the client's logs — the user just sees no proxied tools with no visible reason.

These compound each other: a typo can't be caught by validation because unknown keys are legal, and even a hard parse error doesn't stop the server.

## Doesn't hot reload already handle this?

`ProxyConfigLoader` keeps last-known-good config on a failed *reload* — good, but that protects the running process, not first startup, and does nothing for silently-ignored keys.

## Proposal

1. **`mms config validate [--config PATH]`** — parse + validate, report unknown keys at every nesting level (compare against model fields recursively), non-zero exit on error. Also surfaces the existing permissive-file-mode warning.
2. **Warn on unknown keys at load time** (log each dropped key path). Full `extra="forbid"` is likely too breaking for 0.1.x (forward-compat: older binaries reading newer configs), so warn-and-continue at load + strict mode in the validate command is the pragmatic split.
3. **Make startup failure visible**: when file parse fails, surface it in `stm_proxy_health` / `mms health` and `mms status` ("config file present but failed to parse — running defaults"), not just a stderr line.

## Why separate

Config-layer only; independently testable; no pipeline behavior change.

Issue #582, opened 2026-07-03
cli: mms project mutators bypass the write lock — lost updates on shared projects.toml


memtomem:
## Background

Operational-stability audit (2026-07-03). Every registry-domain CLI mutator runs under the cross-process write lock — `mms import` (`mms_import.py:178`), `mms host sync` (`mms_host.py:909`), and all five `stm_proxy.json` mutators via `with_config_write_lock`. The `mms project` subcommands are the exception: `mms_project.py` does not import `_write_lock` at all.

## Gap

`mms project init` (`mms_project.py:145-163` via `_refresh_index` at `:123-127`), `enable` (`:344`), `disable` (`:380`), and `list --prune` (`:263`) each do an unlocked read-modify-write of the **shared** `~/.mms/projects.toml` (plus per-project `project.toml`):

```python
idx = state.load_projects_index()
idx = state.upsert_project_in_index(idx, name=name, path=str(root))
state.save_projects_index(idx)
```

## Failure scenario

Two concurrent `mms project init` runs in different directories (plausible with parallel agent sessions) interleave: both load the same base index, the loser's `save_projects_index` silently drops the winner's row. No file corruption (writes are atomic via `atomic_write_text`), only lost updates — but a dropped row means that project stops resolving by name until re-registered.

## Suggested fix

Decorate the four mutating callbacks with the existing `@with_write_lock` (same lock domain as import/host-sync, which mutate the same state files). Read-only invocations (`list` without `--prune`, `status`) stay unlocked, mirroring the `--plan` skip convention.

Severity: medium — real lost-update race on a shared file; every sibling mutator in the codebase is locked, so this is an inconsistency rather than a design choice.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #642, opened 2026-07-04
feat(cli): mms tune --apply — apply tuning recommendations to stm_proxy.json

### LINKED ISSUE #615: cli: mms tune --apply — apply stm_tuning_recommendations to stm_proxy.json instead of hand-editing
## Why now

From the 2026-07-04 service review: `stm_tuning_recommendations` (server.py:1802) produces per-tool `max_result_chars` / strategy / retention-floor suggestions but ends with "apply manually to stm_proxy.json". The loop is: agent surfaces recommendation → user opens JSON → finds the right nested per-tool override block → edits by hand. That is exactly the error-prone step the recommendations were meant to remove.

## Proposal

`mms tune [--apply] [--config PATH]`:

- default (no flag): print current recommendations (same data as the MCP tool) as a diff-style preview against the current config;
- `--apply`: write the accepted overrides into `stm_proxy.json` via the existing raw-dict load/save path (which already preserves unknown keys — proxy/config.py:584), honoring the CLI write lock (#582) and creating a timestamped backup like the prune-backup pattern;
- `--yes` for non-interactive; per-tool selection prompts under TTY, all-or-nothing otherwise (respecting the MMS_NO_TUI convention).

## Why separate

New command with write-path implications (locking, backup, validation) — needs its own tests and review, not a ride-along.

memtomem:
Closes #615.

## Why

`stm_tuning_recommendations` produces per-tool `max_result_chars` / `compression` suggestions but ends with "apply manually to `stm_proxy.json`". The loop — agent surfaces a recommendation, user opens the JSON, finds the right nested `tool_overrides` block, edits by hand — is exactly the error-prone step the recommendations were meant to remove.

## What

New command: `mms tune [--apply] [--yes] [--config PATH] [--since-hours F] [--tool NAME] [--json]`

- **`mms tune`** — preview. Runs the same `CompressionTuner` analysis as the MCP tool, offline against the on-disk stores (no running server needed), and renders exactly what `--apply` would write as `field: current -> recommended` lines plus reason and confidence (preview == apply-set, the prune invariant). `--json` emits the same data for scripting (preview-only; combining it with `--apply` is a UsageError).
- **`mms tune --apply`** — per-tool selection (enter-to-toggle picker with all rows pre-selected; `MMS_NO_TUI` degrades to sequential confirms; no TTY without `--yes` errors loudly like `prune`), then writes the accepted overrides into `upstream_servers.<server>.tool_overrides.<tool>` via the raw-dict `_load`/`_save` path — unknown keys survive by construction — under the config write lock (#582).
- **Backup**: before saving, the config's original bytes are snapshotted to a timestamped `stm_proxy.json.bak-<UTC>` slot (O_CREAT|O_EXCL non-clobber loop, mode 0600 since the config can carry env secrets). A running proxy hot-reloads the file, so restore is one `cp`, echoed in the output.

Correctness details:

- **Action merge policy** — a single recommendation can carry colliding actions (strategy-pin + feedback both emit `compression`; a budget heuristic + feedback both emit `max_result_chars`). `compression` is last-wins (an agent-reported deficiency beats a latency pin); `max_result_chars` takes the numeric max ("budget too small" signals combine; feedback beats the savings heuristic). Merged changes are annotated in the rendered reason.
- **Never write a config the server can't parse** — the typed load must succeed before tuning (a file failing validation is one a running server silently ignores, #611), and the mutated dict is schema-validated again before the backup/save; failures exit 1 with nothing written.
- **Preview stays read-only** — `MetricsStore`/`CompressionFeedbackStore.initialize()` run their idempotent DDL/migrations (same step as server startup), so they only run when the DB files already exist; a preview never creates anything. Recommendations for servers present in metrics but absent from the config file (env-defined / renamed upstreams) are reported as skipped instead of written.
- Typed values: `max_result_chars` is written as `int`, `retention_floor` (not emitted by today's heuristics, handled defensively) as `float`; recommended `compression` strings are `CompressionStrategy` members, so no mapping.

## Tests

15 new tests: `TestTune` (CLI integration over a real seeded metrics DB — preview no-op guard, diff render, `--json` shape + UsageError, `--apply --yes` typed write + unknown-key preservation, timestamped backup content/mode + same-second `.1` slot, non-TTY error, no-recommendation no-op, invalid-config refusal) and `TestMergeTuneChanges` (merge policy unit tests). `tune --apply` joins the `TestConfigWriteLock` mutator parametrize; plain `tune` joins the skip-path assertion. Full suite 4337 green; ruff + mypy clean.

Docs: `docs/cli.md` gains a `### tune` section + Commands-index row (pinned by `test_cli_md_commands_index_matches_click_group`), and the `stm_tuning_recommendations` docstrings/table row now point at `mms tune --apply`.

CHANGELOG entry follows in this PR citing its number.


pandas-studio (approved):

memtomem:
**codex review round 1: NEEDS-FIX (2 Major, 0 Blocker)** — dispositions:

1. **Major — tuner joins feedback by tool name only** (`tuner.py:103` / `compression_feedback_store.py`): CONFIRMED, but pre-existing on `main` (it mis-attributes feedback in today's `stm_tuning_recommendations` display too, independent of this PR). Per repo convention (drive-by fixes ship separately), fixed in **#643** with store-level + tuner-level regression tests. **Sequencing: merge #643 first**, so the apply path never ships against the tool-name join.

2. **Major — `_apply_tune_changes` crashes on a malformed raw server entry masked by an env override** (typed validation sees the env-overlaid composite; the raw file entry can still be a non-dict, or carry a non-dict `tool_overrides`): CONFIRMED and fixed here in cfe9667 — the writability guard in `_merge_tune_changes` now requires a dict entry with dict-or-absent `tool_overrides` and skips anything else through the same human-readable warning path as servers missing from the file, with parametrized regression tests.

Round 2 re-review to follow.

memtomem:
**codex review round 2: NEEDS-FIX (1 Major)** — the round-1 server-entry guard missed the same env-masked crash one level deeper: a non-dict *per-tool* entry (`"tool_overrides": {"t": "bad"}`) would make `_apply_tune_changes`' `setdefault(tool, {})[field]` raise `TypeError` before the clean skip/validation path.

**Disposition: FIXED in 6bb736e** — the writability guard in `_merge_tune_changes` now also requires the recommendation's target per-tool entry to be absent or a dict, skipping malformed ones with a warning; `per-tool-override-not-dict` added to the parametrized regression test. Round 3 re-review running.

memtomem:
**codex review round 3: NEEDS-FIX (1 Major)** — the per-tool guard skips a malformed recommendation *target*, but a malformed **sibling** `tool_overrides` entry (untouched by the tuner) still failed the post-mutation whole-dict validation, stranding healthy changes after the analysis/selection work.

**Disposition: FIXED in 8d3155f, via the root cause rather than deeper guards** — guards can never pre-empt every masked invalidity (any part of the raw file can be invalid-as-written while an env override masks it from the typed load). Once the typed load succeeds, `mms tune` now validates the file **as written** up front: `--apply` refuses immediately, pointing at `mms config validate` (whose as-written semantics this matches); preview warns but still renders the analysis. Merge-level guards remain as defense-in-depth for the preview path; post-mutation validation remains the backstop for our own writes. End-to-end regression added (`test_apply_refuses_env_masked_invalid_raw_file`: masked malformed sibling + healthy recommendation → apply refuses with file untouched and no backup; preview warns and renders). Round 4 re-review running.

memtomem:
**codex review round 4: SHIP (0 findings)** — verified the as-written validation runs after the typed load and before any analysis/backup/write work; preview still renders with a warning; merge guards and post-mutation validation confirmed as backstops. codex also ran the `TestTune` + `TestMergeTuneChanges` suites (20 passed) and a whitespace check.

Review summary: R1 NEEDS-FIX (2 Major: feedback join → adjudicated to #643, merge first; raw-entry crash → fixed) → R2 NEEDS-FIX (per-tool entry crash → fixed) → R3 NEEDS-FIX (malformed sibling strands healthy changes → root-cause fix: up-front as-written validation) → **R4 SHIP**.

